PR #23, opened 2017-02-04
drm: Fix Mesa driver names, add mapping of amdgpu kernel driver to radeonsi Mesa driver

fhvwy:
As ML.

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -43,6 +43,7 @@ static const struct driver_name_map g_driver_name_map[] = {
     { "hybrid",     6, "hybrid" }, // Intel OTC Hybrid driver
     { "nouveau",    7, "nouveau"  }, // Mesa Gallium driver
     { "radeon",     6, "r600"     }, // Mesa Gallium driver
+    { "amdgpu",     6, "radeonsi" }, // Mesa Gallium driver
     { NULL, }
 };
 
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -41,8 +41,8 @@ static const struct driver_name_map g_driver_name_map[] = {
     { "pvrsrvkm",   8, "pvr"    }, // Intel UMG PVR driver
     { "emgd",       4, "emgd"   }, // Intel ECG PVR driver
     { "hybrid",     6, "hybrid" }, // Intel OTC Hybrid driver
-    { "nouveau",    7, "gallium" }, // Mesa Gallium driver
-    { "radeon",     6, "gallium" }, // Mesa Gallium driver
+    { "nouveau",    7, "nouveau"  }, // Mesa Gallium driver
+    { "radeon",     6, "r600"     }, // Mesa Gallium driver
     { NULL, }
 };
 


fhvwy:
Verification beyond looking at the files in a distribution package as above is probably going to require suitable hardware and building Mesa:

* Find a machine with hardware using the amdgpu kernel driver (recent AMD).
* Get current Mesa (13 or git master); build with the correct driver enabled and install it.
* Run vainfo (or anything else), make sure it loads the driver you think it should (radeonsi).

Repeat for radeon/r600 (older AMD) and nouveau (most Nvidia).

I have done this for amdgpu and it works as expected.  The others aren't tested, but I observed them in the source at the same time and it seemed sensible to correct the names.  (I've been mildly annoyed by this issue for a while, but I generally worked around it with LIBVA_DRIVER_NAME.  That doesn't work once you load i965 in the same process, though, which is what prompted actually fixing it properly like this.)

seanvk:
@fhvwy Will removal of gallium as a backend choice break anything existing?  We originally merged this support for gallium backend identification.  How has it changed so that now the actual driver is referenced instead of gallium?  Can you point to a Mesa commit that made this patch set necessary?

seanvk:
@fhvwy Also, please create an Issue in this project for these patches and include a fixes # in the commit log as explained in CONTRIBUTING.md.  Thanks, Sean

fhvwy:
https://cgit.freedesktop.org/mesa/mesa/commit/?id=0c0f841e5de27d01312f8857641668ca439b1ab1
https://cgit.freedesktop.org/mesa/mesa/commit/?h=12.0&id=c01ebdc83e03307fa1373f3364131ed0c2d380ee

va/drm/va_drm_utils.c
```diff
@@ -41,8 +41,8 @@ static const struct driver_name_map g_driver_name_map[] = {
     { "pvrsrvkm",   8, "pvr"    }, // Intel UMG PVR driver
     { "emgd",       4, "emgd"   }, // Intel ECG PVR driver
     { "hybrid",     6, "hybrid" }, // Intel OTC Hybrid driver
-    { "nouveau",    7, "gallium" }, // Mesa Gallium driver
-    { "radeon",     6, "gallium" }, // Mesa Gallium driver
+    { "nouveau",    7, "nouveau"  }, // Mesa Gallium driver
+    { "radeon",     6, "r600"     }, // Mesa Gallium driver
```
xhaihao: How about if user still uses an old version of mesa?

seanvk: Yes that was my earlier concern.  I'm leery of chasing a Mesa change when we break existing users.  I'd also like Julien to weigh in on this since he originally requested this change. @julien.isorce@gmail.com

seanvk: @CapCom

fhvwy: Looking a bit more closely, there doesn't appear to be any released version of Mesa which used the gallium_drv_video.so name.  The change above happened after the 12 branchpoint, but before the actual 12 release.  Versions before 12 did not include the gallium VAAPI driver at all.

seanvk: Julien's support for nouveau st/va landed in 11.x.  When 12 release rolled around, gallium_drv_video.so was replaced with hard links to the respective backends.  These patches from you keep VA consistent with the current Mesa release.  Prior to this point it was an inconsistent level of support.  So I'm okay with this change based on the Mesa 12 commits which imply a switch to the platform names.  lgtm.

CapOM:
lgtm, sorry for the delay.

PR #27, opened 2017-02-15
Wayland-drm: Fix not finding wl_drm_interface on systems with libglvnd

jwrdegoede:
We do not want just any libEGL.so.1 we want mesa's libEGL.so.1 as that
is the only way which defines the wl_drm_interface symbol we need,
one systems with libglvnd libEGL.so.1 is a dispatcher library provided
by libglvnd and the actual mesa libEGL we want is named libEGL_mesa.so.0
so try that first.

Signed-off-by: Hans de Goede <hdegoede@redhat.com>

### DIFF
--- a/va/wayland/va_wayland_drm.c
+++ b/va/wayland/va_wayland_drm.c
@@ -38,7 +38,10 @@
 #include "wayland-drm-client-protocol.h"
 
 /* XXX: Wayland/DRM support currently lives in Mesa libEGL.so.* library */
-#define LIBWAYLAND_DRM_NAME "libEGL.so.1"
+/* First try the soname of a glvnd enabled mesa build */
+#define LIBWAYLAND_DRM_NAME "libEGL_mesa.so.0"
+/* Then fallback to plain libEGL.so.1 (which might not be mesa) */
+#define LIBWAYLAND_DRM_NAME_FALLBACK "libEGL.so.1"
 
 typedef struct va_wayland_drm_context {
     struct va_wayland_context   base;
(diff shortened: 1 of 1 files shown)


va/wayland/va_wayland_drm.c
```diff
@@ -38,7 +38,10 @@
 #include "wayland-drm-client-protocol.h"
 
 /* XXX: Wayland/DRM support currently lives in Mesa libEGL.so.* library */
-#define LIBWAYLAND_DRM_NAME "libEGL.so.1"
```
seanvk: Good point, I agree.

seanvk: We need to make similar changes to 01org/libva-utils

jwrdegoede:
Hi,

On 15-02-17 18:43, Sean V. Kelley wrote:
> *@seanvk* commented on this pull request.
>
> ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>
> In va/wayland/va_wayland_drm.c <https://github.com/01org/libva/pull/27>:
>
>> @@ -38,7 +38,10 @@
>  #include "wayland-drm-client-protocol.h"
>
>  /* XXX: Wayland/DRM support currently lives in Mesa libEGL.so.* library */
> -#define LIBWAYLAND_DRM_NAME "libEGL.so.1"
>
> We need to make similar changes to 01org/libva-utils

[hans@shalem projects]$ grep -r libEGL libva-utils
[hans@shalem projects]$

So that does not seem to be necessary.

Regards,

Hans

seanvk:
It's a potential dep in the future because if we write specific EGL based tests.  This is a result of splitting libva-utils from libva.

seanvk:
But you are right, we don't need it for libva-utils at this time. @jwrdegoede

seanvk:
I actually removed the EGL deps from libva-utils configure.ac recently

PR #33, opened 2017-02-23
wayland: Don't commit and ship generated files

jku:
I believe shipping wayland-drm-client-protocol.h is wrong: The header
should always be generated by the wayland-scanner that matches the
runtime wayland version. Currently when someone clones the repo and
builds, the shipped version is used.

Signed-off-by: Jussi Kukkonen <jussi.kukkonen@intel.com>


va/wayland/Makefile.am
```diff
@@ -53,7 +53,7 @@ protocol_source_h = \
 noinst_LTLIBRARIES		= libva_wayland.la
 libva_waylandincludedir		= ${includedir}/va
 libva_waylandinclude_HEADERS	= $(source_h)
-libva_wayland_la_SOURCES	= $(source_c) $(protocol_source_h)
```
seanvk: Please file an issue on GitHub and add Fixes #n to the commit header otherwise, I'm fine with this change.

va/wayland/Makefile.am
```diff
@@ -65,5 +65,8 @@ EXTRA_DIST = \
 	wayland-drm.xml         \
 	$(NULL)
 
+BUILT_SOURCES = $(protocol_source_h)
+CLEANFILES = $(BUILT_SOURCES)
+
```
xhaihao: Please check wayland-scanner in configure.ac for wayland platform. It is possible user doesn't install wayland-scanner.

jku: I didn't want to include that here because the configure checks seemed a little non-standard  (like you can say "--enable-wayland" without having wayland-client installed and configure succeeds but wayland is disabled). But I can make configure fail if scanner is not available when $USE_WAYLAND.

jku:
V2 Changelog:
* Added a paragraph and "Fixes #N" to commit message
* configure now fails if $USE_WAYLAND and wayland-scanner cannot be found

PR #37, opened 2017-04-07
A series patches for Android

cwhuang:
These patches fix building issues of Android on the master branch.

Tested OK on Android-x86 7.1 cm branch.


xuguangxin:
@LindaYu17 please help review this

cwhuang:
Personally I don't think libva*.so should go to /vendor/lib folder. That folder is usually for vendor's proprietary libraries. But libva and vaapi are open source libs. I understand they are mainly developed by Intel. But I believe you have developers or contributors from others. Right? In particular, libva is not limited to be used by one vendor. So we should not considered them as vendor provided libraries. We should just treat them as system libs, just like the normal linux desktop.

But if you insist, I can add additional patches to do that.

xhaihao:
Agree with @cwhuang,  besides intel-vaapi-driver, libva supports other vaapi backend drivers, such as r600, nouveau and radeonsi.

kalyankondapally:
HALS/Drivers going into vendor/ can be updated independent of system changes i.e. driver bug fixng etc. We already have patches floating here: https://github.com/android-ia/libva/commits/master.

LindaYu17:
relative path should be set to /vendor/lib but not /vendor/lib/hw to avoid linking error.

+LOCAL_PROPRIETARY_MODULE := true
+LOCAL_MODULE_RELATIVE_PATH := .

va/android/va_android.cpp
```diff
@@ -194,7 +194,7 @@ extern "C"  {
 VAStatus vaPutSurface (
     VADisplay dpy,
     VASurfaceID surface,
-    sp<ISurface> draw, /* Android Surface/Window */
+    sp<ANativeWindow> draw, /* Android Native Window */
     short srcx,
     short srcy,
     unsigned short srcw,
```
chivakker: are you using vaPutSurface on Android?

va/Android.mk
```diff
@@ -44,57 +39,38 @@ LOCAL_SRC_FILES := \
...
 
-# Android Jelly Bean defined ALOGx, older versions use LOGx
-ifeq ($(ALOG_VERSION), $(ALOG_VERSION_REQ))
-LOCAL_CFLAGS += -DANDROID_ALOG
-else
-LOCAL_CFLAGS += -DANDROID_LOG
-endif
+LOCAL_CFLAGS := \
+	$(if $(filter user,$(TARGET_BUILD_VARIANT)),,-DENABLE_VA_MESSAGING) \
+	-DLOG_TAG=\"libva\"
```
chivakker: is the build variant actually needing a full trace from libva?

xuguangxin:
@kalyankondapally seems it's hard to align on directory thing. Do you think it's ok we maintain a local patch on Android IA  to change directory to vendor?

va/android/va_android.cpp
LindaYu17: no, we don't use it in Android.

va/Android.mk
xuguangxin: Hi @LindaYu17 , any comments for this?

xuguangxin:
Hi @cwhuang 
@kalyankondapally convinced @xhaihao and me, Could you help to change directory as we did in https://github.com/android-ia/libva/commits/master?

Also please address Daniel's second comments
>is the build variant actually needing a full trace from libva?

thanks

va/Android.mk
cwhuang: Why not? It's just for debug build.

Besides, what trace should I see? Except these I didn't see other log from libva:

04-23 22:04:40.498  1453  3028 I libva   : VA-API version 0.40.0
04-23 22:04:40.498  1453  3028 I libva   : Open DRM devices /dev/dri/card0 for (null)
04-23 22:04:40.498  1453  3028 I libva   : va_getDriverName() returns 0
04-23 22:04:40.498  1453  3028 I libva   : Trying to open /system/lib/dri/i965_drv_video.so
04-23 22:04:40.498  1453  3028 I libva   : Found init function __vaDriverInit_0_33
04-23 22:04:40.499  1453  3028 I libva   : va_openDriver() returns 0

How could I enable more verbose log?

cwhuang:
Sorry for reply late. I just rebase my patches and change libs path to /vendor as request.
(though I still didn't see any real benefit from it)

xuguangxin:
@cwhuang, thanks, According to @kalyankondapally , /system will control by google, vendor/ will control by the vendor. If will in a different partition. It's easy to update when google have new android version

xuguangxin:
@LindaYu17 , please help to review.thanks

LindaYu17:
+1

PR #46, opened 2017-05-02
Move all the libraries to /vendor folder

LindaYu17:
Signed-off-by: Yu, Linda <linda.yu@intel.com>

PR #47, opened 2017-05-04
android: move libva-android to /vendor

cwhuang:
This was missed in the previous commit bed97ef.

### DIFF
--- a/va/Android.mk
+++ b/va/Android.mk
@@ -89,6 +89,7 @@ LOCAL_C_INCLUDES += \
 
 LOCAL_MODULE_TAGS := optional
 LOCAL_MODULE := libva-android
+LOCAL_PROPRIETARY_MODULE := true
 
 LOCAL_SHARED_LIBRARIES := libva libdrm
 
--- a/va/Android.mk
+++ b/va/Android.mk
@@ -93,39 +93,3 @@ LOCAL_MODULE := libva-android
 LOCAL_SHARED_LIBRARIES := libva libdrm
 
 include $(BUILD_SHARED_LIBRARY)
-
-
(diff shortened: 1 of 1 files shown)


xhaihao:
Could you squash the commits in https://github.com/01org/libva/pull/46 and https://github.com/01org/libva/pull/47 into one commit ?

cwhuang:
Oh! I didn't notice Linda posted a similar patch in #46. But it's still incomplete.
Actually #47 has contained the change of #46 .

xuguangxin:
Hi @cwhuang and @LindaYu17 
I only push libva and libva-android to board . The sample_decod is working.
Do we really need libva-tpi and libva-egl?

cwhuang:
Hi @xuguangxin,
It just make the change (move libs to /vendor) complete.
I'm not sure when or how they are used.

Or do you suggest to remove them at all if they are unnecessary?

xuguangxin:
Hi cwhuang:
@LindaYu17  correct me if I am wrong.
This is unnecessary, please help remove it. But the tpi headers used by our internal object. I do not know the header install details, please do not remove it from headers.
thanks

xhaihao:
Both libva-tpi and libva-egl should not be used any more.  libva provides a unified mechanism of sharing buffers with the third party library.  We can merge this change because some applications on android still includes the libva-tpi and libva-egl headers. However it would be better that android team can remove the dependency on libva-tpi and libva-egl, we will deprecate the use of libva-tpi and libva-egl.

cwhuang:
OK. I've updated the patches to remove rules of libva-tpi and libva-egl.

PR #48, opened 2017-05-08
android: open renderD128 by default

xuguangxin:
if we open card0, we need be a root to call I915_PARAM_CHIPSET_ID,
but mediaservice's owner is not root, open renderD128 will fix this.


xuguangxin:
Hi @xhaihao , could you merge this?

PR #57, opened 2017-06-02
Remove the dummy driver

xhaihao:
No one is interested in this dummy driver and it is not maintained.

This fixes https://github.com/01org/libva/issues/28


yakuizhao:
It will be better to keep them.
The example code in dummy driver can be referred to. For example: Heap management, the basic flowchart of driver initialization and so on.

xhaihao:
The heap management is not the requirement of vaapi, each backend driver may implement its object management. As for the basic flowchart, user can refer to any open source driver.

PR #59, opened 2017-06-06
Refine the data type of slice_data_flag in VASliceParameterBufferHEVC…

yakuizhao:
… to be consistent with other codecs

Currently the data type of slice_data_flag in VASliceParameterBufferHEVC is
uint16_t while it is uint32_t for other codecs. It is inconsistent.
As the memory is allocated as aligned, it still allocates four bytes for it.
So it will be better to use the consistent data type.

This fixes https://github.com/01org/libva/issues/58

Signed-off-by: Zhao Yakui <yakui.zhao@intel.com>

### DIFF
--- a/va/va_dec_hevc.h
+++ b/va/va_dec_hevc.h
@@ -254,7 +254,7 @@ typedef struct  _VASliceParameterBufferHEVC
     /** \brief The offset to the NAL unit header for this slice */
     uint32_t                slice_data_offset;
     /** \brief Slice data buffer flags. See \c VA_SLICE_DATA_FLAG_XXX. */
-    uint16_t                slice_data_flag;
+    uint32_t                slice_data_flag;
     /**
      * \brief Byte offset from NAL unit header to the begining of slice_data().
      *


pengche1:
Agree with Yakui. And I suggest to merge this change when bump the API version

fhvwy:
No - this would be an ABI break.  It changes the location of the actually-used two bytes of slice_data_flag on big-endian platforms (e.g. Mesa on POWER).

yakuizhao:
@fhvwy , Thanks for your advice. This patch is part of updating libva version.  If both driver and library are compiled, it won't break anything.

fhvwy:
@yakuizhao  Right, if the SONAME changes at the same time then there is no ABI concern.

To make it clearer what's going to happen, perhaps this could be added immediately to a new "version bump" branch of libva rather than keeping it here as a diff against master?  (That branch can then be merged atomically onto master at some future point, including the SONAME change.)

xhaihao:
@fhvwy thanks for your suggestion, I will create a new branch for the new version of libva.

xhaihao:
@yakuizhao Your patch has been merged into v2.0-branch,  so I closed this PR.

PR #61, opened 2017-06-09
Fix small typos.

giuscri:
/etc/libav.conf was spelled incorrectly; one line
of code was indented one level more than needed.

Signed-off-by: Giuseppe <giuscri@gmail.com>

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -60,7 +60,7 @@
 
 /*
  * read a config "env" for libva.conf or from environment setting
- * liva.conf has higher priority
+ * libva.conf has higher priority
  * return 0: the "env" is set, and the value is copied into env_value
  *        1: the env is not set
  */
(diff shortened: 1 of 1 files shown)


va/va.c
```diff
@@ -77,7 +77,7 @@ int va_parseConfig(char *env, char *env_value)
     while (fp && (fgets(oneline, 1024, fp) != NULL)) {
 	if (strlen(oneline) == 1)
 	    continue;
-        token = strtok_r(oneline, "=\n", &saveptr);
+    token = strtok_r(oneline, "=\n", &saveptr);
```
xhaihao: I still see the wrong indentation when viewing the whole file in github.  the reason is that line 80 was indented by whitespace however line 78 was indented by a tab.

giuscri:
@xhaihao I've retabbed the whole function and rebased into one commit.

PR #65, opened 2017-06-13
Wayland thread-safety

### LINKED ISSUE #11: wayland: thread safety and leaked proxies
migrated from Bugzilla [#90200](https://bugs.freedesktop.org/show_bug.cgi?id=90200)
status NEEDINFO severity *normal* in component *core* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: haihao

Original attachment names and IDs:
- _[0001-wayland-Use-a-private-queue-and-free-registry-proxy.patch](https://bugs.freedesktop.org//attachment.cgi?id=115381)_ (ID 115381)
- _[0001-wayland-Use-private-queue-and-free-proxy.patch](https://bugs.freedesktop.org//attachment.cgi?id=115382)_ (ID 115382)

On 2015-04-27 18:21:27 +0000, Olivier CrÃªte wrote:

> Created attachment 115381
> libva: wayland: Use a private queue and free registry proxy
> 
> both libva and the intel driver's wayland integration use the global queue. This is not thread safe and may be problematic for some applications. Also, it doesn't clear the proxies and queue correctly before unloading the libEGL.so, which leads to segfaults as libwayland-client tries to access the static wl_interface in the bit that has been unloaded!

On 2015-04-27 18:21:56 +0000, Olivier CrÃªte wrote:

> Created attachment 115382
> intel-driver: Basically the same patch for the driver

On 2015-04-30 01:03:44 +0000, haihao wrote:

> The old version of wayland-client.h doesn't have wl_display_roundtrip_queue. Could you update the dependency on wayland in configure.ac too ?

On 2015-11-24 04:43:38 +0000, haihao wrote:

> Could you update your patch to check wayland version? otherwise your patch might break driver building.

On 2016-08-31 04:03:06 +0000, Gautham Kantharaju wrote:

> Using vaapisink for video playback, trying to stop/start the video playback or switch video files.
> 
> SIGSEGV is caught in libwayland-client. Pulled in the patches w.r.t libva and libva-intel-driver. Still the issue exists.
> 
> Can you let me know, if anything needs to be changed the vaapi gstreamer plugin also in order to resolve the issue?

On 2016-09-05 01:50:46 +0000, haihao wrote:

> Someone also reported crash issue when using wayland, the issue disappears after upgrading wanyland-drm in the driver. Could you have a try with the following patches:
> 
> libva:
> https://lists.freedesktop.org/archives/libva/2016-September/004356.html
> https://lists.freedesktop.org/archives/libva/2016-September/004357.html
> 
> libva-intel-driver:
> https://lists.freedesktop.org/archives/libva/2016-September/004361.html
> https://lists.freedesktop.org/archives/libva/2016-September/004362.html

pkerling:
Fixes #11 

I guess that `va_wayland_emgd.c` would need the same fix, but I can't figure out how it is supposed to work. It only binds to the global `wl_emgd` and then does... nothing? Does it even work?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,7 +78,8 @@ m4_define([libva_lt_age],
 m4_define([libdrm_version], [2.4])
 
 # Wayland minimum version number
-m4_define([wayland_api_version], [1.0.0])
+# 1.11.0 for wl_proxy_create_wrapper
+m4_define([wayland_api_version], [1.11.0])
 
 AC_PREREQ(2.57)
 AC_INIT([libva],
--- a/va/wayland/va_wayland_drm.c
+++ b/va/wayland/va_wayland_drm.c
@@ -46,6 +46,7 @@
 typedef struct va_wayland_drm_context {
     struct va_wayland_context   base;
     void                       *handle;
+    struct wl_event_queue      *queue;
     struct wl_drm              *drm;
     struct wl_registry         *registry;
     void                       *drm_interface;
@@ -150,6 +151,11 @@ va_wayland_drm_destroy(VADisplayContextP pDisplayContext)
         wl_drm_ctx->registry = NULL;
     }
 
+    if (wl_drm_ctx->queue) {
+        wl_event_queue_destroy(wl_drm_ctx->queue);
+        wl_drm_ctx->queue = NULL;
+    }
+
     if (wl_drm_ctx->handle) {
         dlclose(wl_drm_ctx->handle);
         wl_drm_ctx->handle = NULL;
@@ -187,6 +193,19 @@ static const struct wl_registry_listener registry_listener = {
     NULL,
 };
 
+static bool
+wayland_roundtrip_queue(struct wl_display *display,
+                         struct wl_event_queue *queue)
+{
+    if (wl_display_roundtrip_queue(display, queue) < 0) {
+        int err = wl_display_get_error(display);
+        va_wayland_error("Wayland roundtrip error: %s (errno %d)", strerror(err), err);
+        return false;
+    } else {
+        return true;
+    }
+}
+
 bool
 va_wayland_drm_create(VADisplayContextP pDisplayContext)
 {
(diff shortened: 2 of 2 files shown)


xhaihao:
va_wayland_emgd doesn't work with intel-vaapi-driver. it works with another vaapi backend on GMA500

xhaihao:
Could you add 'signed-off-by your name <your email>' at the end of each commit? BTW some commits are based on https://bugs.freedesktop.org//attachment.cgi?id=115381, could you also add 'signed-off-by Olivier Crete <olivier.crete@collabora.com>' as well?

xhaihao:
could you please remove all unnecessary whitespace? I got lots of warning messages when applying your patches to master.

> .git/rebase-apply/patch:21:  trailing whitespace.
> 
> .git/rebase-apply/patch:102: trailing whitespace.
>     
> .git/rebase-apply/patch:108: trailing whitespace.
>     
> .git/rebase-apply/patch:141: trailing whitespace.
>     
> .git/rebase-apply/patch:146: trailing whitespace.
> 
>

pkerling:
> va_wayland_emgd doesn't work with intel-vaapi-driver. it works with another vaapi backend on GMA500

Does that mean it can be ignored for now?

Other comments are fixed, please recheck.

xhaihao:
yes, we can ignore va_wayland_emgd.

configure.ac
```diff
@@ -78,7 +78,8 @@ m4_define([libva_lt_age],
 m4_define([libdrm_version], [2.4])
 
 # Wayland minimum version number
-m4_define([wayland_api_version], [1.0.0])
+# 1.11.0 for wl_proxy_create_wrapper
+m4_define([wayland_api_version], [1.11.0])
```
uartie: ubuntu xenial, ubuntu trusty, and fedora 24 don't have packages for wayland 1.11.  Thus, it will break compilation in libva-utils when attempting to use --enable-wayland all around and relying on distro wayland packages.  This will force everyone to compile wayland locally... I don't think that's a good idea!

pkerling: Hm I can see that this is unfortunate, but what are you suggesting then? Keep it broken? There is just no way to realize completely thread-safe behavior prior to 1.11.
Guess the distros will have to update wayland to 1.11 if they want to update libva. If you're compiling libva by yourself already and not using the distro libva packages, I don't really see the problem. It's to be expected that you will also have to build some dependencies if you need the latest version.

uartie: It's not just me compiling libva... it's our CI and entire automated infrastructure that will need amended.  It's not impossible to do, but needs to be considered when making these kinds of changes.  Anyhow, I'll work with maintainers to come to a decision about how we want to deal with these kinds of breakages/requirements.

pkerling: I was not talking about you specifically - should have made that clear, sorry. I'm only here fixing things that need fixing, how maintainers want to handle deps issues is not my decision :)
Speaking of CI, why was this PR marked OK by CI anyway? Does it not build the wayland stuff?

uartie: The CI *does* use --enable-wayland in PR builds.  Unfortunately, libva's configure silently disables wayland if requirements aren't met :(.  That is a bug I guess.

For PR builds, the CI does not build downstream dependents (e.g. libva-utils) on top of the PR.  That only happens for post-merge CI builds right now (where we noticed the breakage, i.e. libva-utils expected libva-wayland et. al. to exist).

Nonetheless, I agree we need to keep moving forward... we'll figure out the best path to deal with this.  Thankfully, we build inside pre-configured docker containers so adding a pre-compiled wayland 1.11 to them might be the best solution for now.  But that's not your problem to worry about :)

PR #67, opened 2017-06-13
wayland: Do not rely on external library for wl_drm_interface

pkerling:
Doing the runtime lookup is error-prone (as indicated in the code
already by the two library names) and not necessary at all.
Just compile wl_drm_interface as part of libva which is the usual
way to use extra Wayland protocols. Care must be taken as to not
accidentally export the symbol though.


xhaihao:
LGTM, @pkerling Could you help to resolve the conflicts? BTW please remove the unnecessary whitespaces in the patches.

pkerling:
@xhaihao rebased

PR #69, opened 2017-06-16
new features serials ---RIR ROI ect.

xfengcarl:
this PR include 12 patches , include processing rate , range of interest , rolling intra refresh ...

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -919,6 +919,21 @@ VAStatus vaQueryConfigAttributes (
   return ctx->vtable->vaQueryConfigAttributes( ctx, config_id, profile, entrypoint, attrib_list, num_attribs);
 }
 
+VAStatus vaQueryProcessingRate (
+    VADisplay dpy,
+    VAConfigID config_id,
+    VAProcessingRateParameter *proc_buf,
+    unsigned int *processing_rate	/* out */
+)
+{
+  VADriverContextP ctx;
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+  if(!ctx->vtable->vaQueryProcessingRate)
+      return VA_STATUS_ERROR_UNIMPLEMENTED;
+  return ctx->vtable->vaQueryProcessingRate( ctx, config_id, proc_buf, processing_rate);
+}
+
 /* XXX: this is a slow implementation that will be removed */
 static VAStatus
 va_impl_query_surface_attributes(
--- a/va/va.h
+++ b/va/va.h
@@ -1551,6 +1551,9 @@ typedef enum
      *surface, it will override the "skipCheckDisable" setting in VAEncMiscParameterEncQuality.
      */
     VAEncMacroblockDisableSkipMapBufferType = 53,
+    /** decode stream out buffer, intermedia data of decode, it may include MV, MB mode etc.
+      * it can be used to detect motion and analyze the frame contain  */
+    VADecodeStreamoutBufferType             = 56,
     VABufferTypeMax
 } VABufferType;
 
--- a/va/va.h
+++ b/va/va.h
@@ -577,6 +577,17 @@ typedef enum
      * VAConfigAttribValEncRateControlExt union.
      */
     VAConfigAttribEncRateControlExt   = 26,
+    /**
+     * \brief Processing rate reporting attribute. Read-only.
+     *
+     * This attribute conveys whether the driver supports reporting of
+     * encode/decode processing rate based on certain set of parameters
+     * (i.e. levels, I frame internvals) for a given configuration.
+     * If this is supported, vaQueryProcessingRate() can be used to get
+     * encode or decode processing rate.
+     * See \c VA_PROCESSING_RATE_xxx for encode/decode processing rate
+     */
+    VAConfigAttribProcessingRate    = 27,
 
     /**
      * \brief Encode function type for FEI.
@@ -855,6 +866,15 @@ typedef union _VAConfigAttribValEncRateControlExt {
     uint32_t value;
 } VAConfigAttribValEncRateControlExt;
 
+/** @name Attribute values for VAConfigAttribProcessingRate. */
+/**@{*/
+/** \brief Driver does not support processing rate report */
+#define VA_PROCESSING_RATE_NONE                       0x00000000
+/** \brief Driver supports encode processing rate report  */
+#define VA_PROCESSING_RATE_ENCODE                     0x00000001
+/** \brief Driver supports decode processing rate report  */
+#define VA_PROCESSING_RATE_DECODE                     0x00000002
+/**@}*/
 /**
  * if an attribute is not applicable for a given
  * profile/entrypoint pair, then set the value to the following 
@@ -1412,6 +1432,68 @@ typedef enum
     VAEncFEIMVPredictorBufferType       = 47,
     VABufferTypeMax
 } VABufferType;
+
+/**
+ * Processing rate parameter for encode.
+ */
+typedef struct _VAProcessingRateParameterEnc {
+    /** \brief Profile level */
+    uint8_t         level_idc;
+    uint8_t         reserved[3];
+    /** \brief quality level. When set to 0, default quality
+     * level is used.
+     */
+    uint32_t        quality_level;
+    /** \brief Period between I frames. */
+    uint32_t        intra_period;
+    /** \brief Period between I/P frames. */
+    uint32_t        ip_period;
+} VAProcessingRateParameterEnc;
+
+/**
+ * Processing rate parameter for decode.
+ */
+typedef struct _VAProcessingRateParameterDec {
+    /** \brief Profile level */
+    uint8_t         level_idc;
+    uint8_t         reserved0[3];
+    uint32_t        reserved;
+} VAProcessingRateParameterDec;
+
+typedef struct _VAProcessingRateParameter {
+    union {
+        VAProcessingRateParameterEnc proc_buf_enc;
+        VAProcessingRateParameterDec proc_buf_dec;
+    };
+} VAProcessingRateParameter;
+
+/**
+ * \brief Queries processing rate for the supplied config.
+ *
+ * This function queries the processing rate based on parameters in
+ * \c proc_buf for the given \c config. Upon successful return, the processing
+ * rate value will be stored in \c processing_rate. Processing rate is
+ * specified as the number of macroblocks per second.
+ *
+ * If NULL is passed to the \c proc_buf, the default processing rate for the
+ * given configuration will be returned.
+ *
+ * @param[in] dpy               the VA display
+ * @param[in] config            the config identifying a codec or a video
+ *     processing pipeline
+ * @param[in] proc_buf       the buffer that contains the parameters for
+        either the encode or decode processing rate
+ * @param[out] processing_rate  processing rate in number of macroblocks per
+        second constrained by parameters specified in proc_buf
+ *
+ */
+VAStatus
+vaQueryProcessingRate(
+    VADisplay           dpy,
+    VAConfigID          config,
+    VAProcessingRateParameter *proc_buf,
+    unsigned int       *processing_rate
+);
 
 typedef enum
 {
--- a/va/va.h
+++ b/va/va.h
@@ -692,6 +692,12 @@ typedef enum
      * application can change the rounding setting by VAEncMiscParameterTypeCustomRoundingControl
      */
     VAConfigAttribCustomRoundingControl = 36,
+    /**
+     * \brief Encoding QP info block size attribute. Read-only.
+     * This attribute conveys the block sizes that underlying driver
+     * support for QP info for buffer #VAEncQpBuffer.
+     */
+    VAConfigAttribQPBlockSize            = 37,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
--- a/va/va.h
+++ b/va/va.h
@@ -686,6 +686,12 @@ typedef enum
      * tiling structure, should be ignored. 0 - unsupported, 1 - supported.
      */
      VAConfigAttribEncTileSupport        = 35,
+    /**
+     * \brief whether accept rouding setting from application. Read-only.
+     * This attribute is for encode quality, if it is report,
+     * application can change the rounding setting by VAEncMiscParameterTypeCustomRoundingControl
+     */
+    VAConfigAttribCustomRoundingControl = 36,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
@@ -1581,7 +1587,7 @@ typedef struct _VAProcessingRateParameter {
  * This function queries the processing rate based on parameters in
  * \c proc_buf for the given \c config. Upon successful return, the processing
  * rate value will be stored in \c processing_rate. Processing rate is
- * specified as the number of macroblocks per second.
+ * specified as the number of macroblocks/CTU per second.
  *
  * If NULL is passed to the \c proc_buf, the default processing rate for the
  * given configuration will be returned.
@@ -1629,6 +1635,12 @@ typedef enum
     VAEncMiscParameterTypeDirtyRect      = 13,
     /** \brief Buffer type used for parallel BRC parameters. */
     VAEncMiscParameterTypeParallelBRC   = 14,
+    /** \brief Set MB partion mode mask and Half-pel/Quant-pel motion search */
+    VAEncMiscParameterTypeSubMbPartPel = 15,
+    /** \brief set encode quality tuning */
+    VAEncMiscParameterTypeEncQuality = 16,
+    /** \brief Buffer type used for encoder rounding offset parameters. */
+    VAEncMiscParameterTypeCustomRoundingControl = 17,
     /** \brief Buffer type used for FEI input frame level parameters */
     VAEncMiscParameterTypeFEIFrameControl = 18,
 } VAEncMiscParameterType;
@@ -2082,6 +2094,117 @@ typedef struct _VAEncMiscParameterParallelRateControl {
     uint32_t *num_b_in_gop;
 } VAEncMiscParameterParallelRateControl;
 
+/** per frame encoder quality controls, once set they will persist for all future frames
+  *till it is updated again. */
+typedef struct _VAEncMiscParameterEncQuality
+{
+    union
+    {
+        struct
+        {
+            /** Use raw frames for reference instead of reconstructed frames.
+              * it only impact motion estimation (ME)  stage, and will not impact MC stage
+              * so the reconstruct picture will can match with decode side */
+            uint32_t useRawPicForRef                    : 1;
+            /**  Disables skip check for ME stage, it will increase the bistream size
+              * but will improve the qulity */
+            uint32_t skipCheckDisable                   : 1;
+            /**  Indicates app will override default driver FTQ settings using FTQEnable.
+              *  FTQ is forward transform quantization */
+            uint32_t FTQOverride                        : 1;
+            /** Enables/disables FTQ. */
+            uint32_t FTQEnable                          : 1;
+            /** Indicates the app will provide the Skip Threshold LUT to use when FTQ is
+              * enabled (FTQSkipThresholdLUT), else default driver thresholds will be used. */
+            uint32_t FTQSkipThresholdLUTInput           : 1;
+            /** Indicates the app will provide the Skip Threshold LUT to use when FTQ is
+              * disabled (NonFTQSkipThresholdLUT), else default driver thresholds will be used. */
+            uint32_t NonFTQSkipThresholdLUTInput        : 1;
+            uint32_t ReservedBit                        : 1;
+            /** Control to enable the ME mode decision algorithm to bias to fewer B Direct/Skip types.
+              * Applies only to B frames, all other frames will ignore this setting.  */
+            uint32_t directBiasAdjustmentEnable         : 1;
+            /** Enables global motion bias. global motion also is called HME (Heirarchical Motion Estimation )
+              * HME is used to handle large motions and avoiding local minima in the video encoding process
+              * down scaled the input and reference picture, then do ME. the result will be a predictor to next level HME or ME
+              * current interface divide the HME to 3 level. UltraHME , SuperHME, and HME, result of UltraHME will be input of SurperHME,
+              * result of superHME will be a input for HME. HME result will be input of ME. it is a switch for HMEMVCostScalingFactor
+              * can change the HME bias inside RDO stage*/
+            uint32_t globalMotionBiasAdjustmentEnable   : 1;
+            /** MV cost scaling ratio for HME ( predictors.  It is used when
+              * globalMotionBiasAdjustmentEnable == 1, else it is ignored.  Values are:
+              *     0: set MV cost to be 0 for HME predictor.
+              *     1: scale MV cost to be 1/2 of the default value for HME predictor.
+              *     2: scale MV cost to be 1/4 of the default value for HME predictor.
+              *     3: scale MV cost to be 1/8 of the default value for HME predictor. */
+            uint32_t HMEMVCostScalingFactor             : 2;
+            /**disable HME, if it is disabled. Super*ultraHME should also be disabled  */
+            uint32_t HMEDisable                         : 1;
+            /**disable Super HME, if it is disabled, ultraHME should be disabled */
+            uint32_t SuperHMEDisable                    : 1;
+            /** disable Ultra HME */
+            uint32_t UltraHMEDisable                    : 1;
+            /** disable panic mode. Panic mode happened when there are extreme BRC (bit rate control) requirement
+              * frame size cant achieve the target of BRC.  when Panic mode is triggered, Coefficients will
+              *  be set to zero. disable panic mode will improve quality but will impact BRC */
+            uint32_t PanicModeDisable                   : 1;
+            /** Force RepartitionCheck
+             *  0: DEFAULT - follow driver default settings.
+             *  1: FORCE_ENABLE - enable this feature totally for all cases.
+             *  2: FORCE_DISABLE - disable this feature totally for all cases. */
+            uint32_t ForceRepartitionCheck              : 2;
+
+        };
+        uint32_t encControls;
+    };
+
+    /** Maps QP to skip thresholds when FTQ is enabled.  Valid range is 0-255. */
+    uint8_t FTQSkipThresholdLUT[52];
+    /** Maps QP to skip thresholds when FTQ is disabled.  Valid range is 0-65535. */
+    uint16_t NonFTQSkipThresholdLUT[52];
+
+    uint32_t reserved[VA_PADDING_HIGH];  // Reserved for future use.
+
+} VAEncMiscParameterEncQuality;
+
+/**
+ *  \brief Custom Encoder Rounding Offset Control.
+ *  Application may use this structure to set customized rounding
+ *  offset parameters for quantization.
+ *  Valid when \c VAConfigAttribCustomRoundingControl equals 1.
+ */
+typedef struct _VAEncMiscParameterCustomRoundingControl
+{
+    union {
+        struct {
+            /** \brief Enable customized rounding offset for intra blocks.
+             *  If 0, default value would be taken by driver for intra
+             *  rounding offset.
+             */
+            uint32_t    enable_custom_rouding_intra     : 1 ;
+
+            /** \brief Intra rounding offset
+             *  Ignored if \c enable_custom_rouding_intra equals 0.
+             */
+            uint32_t    rounding_offset_intra           : 7;
+
+            /** \brief Enable customized rounding offset for inter blocks.
+             *  If 0, default value would be taken by driver for inter
+             *  rounding offset.
+             */
+            uint32_t    enable_custom_rounding_inter    : 1 ;
+
+            /** \brief Inter rounding offset
+             *  Ignored if \c enable_custom_rouding_inter equals 0.
+             */
+            uint32_t    rounding_offset_inter           : 7;
+
+           /* Reserved */
+            uint32_t    reserved                        :16;
+        }  bits;
+        uint32_t    value;
+    }   rounding_offset_setting;
+} VAEncMiscParameterCustomRoundingControl;
 /**
  * There will be cases where the bitstream buffer will not have enough room to hold
  * the data for the entire slice, and the following flags will be used in the slice
--- a/va/va.h
+++ b/va/va.h
@@ -602,6 +602,27 @@ typedef enum
      * VAEncMiscParameterTypeDirtyRect.
      */
      VAConfigAttribEncDirtyRect       = 28,
+    /**
+     * \brief Parallel Rate Control (hierachical B) attribute. Read-only.
+     *
+     * This attribute conveys whether the encoder supports parallel rate control.
+     * It is a integer value 0 - unsupported, > 0 - maximum layer supported.
+     * This is the way when hireachical B frames are encoded, multiple independent B frames
+     * on the same layer may be processed at same time. If supported, app may enable it by
+     * setting enable_parallel_brc in VAEncMiscParameterRateControl,and the number of B frames
+     * per layer per GOP will be passed to driver through VAEncMiscParameterParallelRateControl
+     * structure.Currently three layers are defined.
+     */
+     VAConfigAttribEncParallelRateControl   = 29,
+     /**
+     * \brief Dynamic Scaling Attribute. Read-only.
+     *
+     * This attribute conveys whether encoder is capable to determine dynamic frame
+     * resolutions adaptive to bandwidth utilization and processing power, etc.
+     * It is a boolean value 0 - unsupported, 1 - supported.
+     * If it is supported,for VP9, suggested frame resolution can be retrieved from VACodedBufferVP9Status.
+     */
+     VAConfigAttribEncDynamicScaling        = 30,
 
     /**
      * \brief Encode function type for FEI.
@@ -664,9 +685,25 @@ typedef struct _VAConfigAttrib {
 #define VA_RC_CQP                       0x00000010
 /** \brief Variable bitrate with peak rate higher than average bitrate. */
 #define VA_RC_VBR_CONSTRAINED           0x00000020
+/** \brief Intelligent Constant Quality. Provided an initial ICQ_quality_factor,
+ *  adjusts QP at a frame and MB level based on motion to improve subjective quality. */
+#define VA_RC_ICQ			0x00000040
 /** \brief Macroblock based rate control.  Per MB control is decided
  *  internally in the encoder. It may be combined with other RC modes, except CQP. */
 #define VA_RC_MB                        0x00000080
+/** \brief Constant Frame Size, it is used for small tolerent  */
+#define VA_RC_CFS                       0x00000100
+/** \brief Parallel BRC, for hierachical B.
+ *
+ *  For hierachical B, B frames can be refered by other B frames.
+ *  Currently three layers of hierachy are defined:
+ *  B0 - regular B, no reference to other B frames.
+ *  B1 - reference to only I, P and regular B0 frames.
+ *  B2 - reference to any other frames, including B1.
+ *  In Hierachical B structure, B frames on the same layer can be processed
+ *  simultaneously. And BRC would adjust accordingly. This is so called
+ *  Parallel BRC. */
+#define VA_RC_PARALLEL                  0x00000200
 
 /**@}*/
 
@@ -1533,6 +1570,8 @@ typedef enum
     VAEncMiscParameterTypeTemporalLayerStructure   = 12,
     /** \brief Buffer type used for dirty region-of-interest (ROI) parameters. */
     VAEncMiscParameterTypeDirtyRect      = 13,
+    /** \brief Buffer type used for parallel BRC parameters. */
+    VAEncMiscParameterTypeParallelBRC   = 14,
     /** \brief Buffer type used for FEI input frame level parameters */
     VAEncMiscParameterTypeFEIFrameControl = 18,
 } VAEncMiscParameterType;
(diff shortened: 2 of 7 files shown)


xhaihao:
@xfengcarl  as you PR did some changes to the existent data structures, could you submit your PR against v2.0-branch instead master?

va/va.h
```diff
@@ -2350,6 +2351,8 @@ typedef struct _VAPictureParameterBufferH264
         unsigned int value;
     } pic_fields;
     unsigned short frame_num;
+    unsigned char num_ref_idx_l0_default_active_minus1;
+    unsigned char num_ref_idx_l1_default_active_minus1;
```
fhvwy: @xhaihao 
> The two flags are added for slice decoding base mode (VA_DEC_SLICE_MODE_BASE).

Can you explain what this mode does?

I don't believe these values can be used for anything unless you want to decode the whole slice header, since they can always take any value (including ones which would be invalid for num_ref_idx_lN_active_minus) and then be overridden.  But, there isn't enough information here to decode the whole slice header.

xfengcarl: it is from h264 spec. it will take effect when num_ref_idx_active_override_flag in slice header is false

yakuizhao: Agree with what xfengcarl said.
Based on H264 spec, the num_ref_idx_l0/l1_default_active_minus1 is the parameter that applies to the frame. Every slice can select the different num_ref_idx_l0/l1_active minus when the override flag exists. If the slice doesn't have the override flag, it can be derived from the num_ref_idx_l0/l1_default_active_minus1.

fhvwy: The point I am trying to make is that I don't think there is any reason to have these values because they are purely syntactic (optimising bit use in the slice header).

Do you disagree with any of the following?

(a)  The standard offers no constraint beyond a 0-31 range - since they can always be overidden in the slice header, they can take any value at all in that regardless of any other property of the stream.  In particular, they don't tell you anything about the number of references - they need not even be <16 when frame_mbs_only_flag is set.  Hence they are of no value directly.

(b)  They don't do anything if you have a VASliceParameterBufferH264 structure.  This supplies the values num_ref_idx_lN_active_minus1 without num_ref_idx_active_override_flag, so the default values for bitstream parsing are irrelevant.

(c)  Parsing the slice header can't make use of them by determining num_ref_idx_active_override_flag, because you don't know enough about the syntax elements present.  Specifically, bottom_field_pic_order_in_frame_present_flag is unknown, so delta_pic_order_cnt_bottom / delta_pic_order_cnt[1] may or may not be present and you can't parse anything after that point.

xhaihao: In slice decoding base mode, VASliceParameterBufferBase instead of  VASliceParameterBufferH264 is required, HW will parse the header. 

In addition, pic_order_present_flag is used to pass  bottom_field_pic_order_in_frame_present_flag to the driver. FFmpeg has done it in this way:

    pps->cabac             = get_bits1(gb);
    pps->pic_order_present = get_bits1(gb);
    pps->slice_group_count = get_ue_golomb(gb) + 1;

    .pic_order_present_flag                 = pps->pic_order_present,

xfengcarl: agree to remove these two field , there are no num_ref_idx_active_override_flag in slice parameters, so reference num always is determined by the values in slice parameter

xfengcarl: long format: application parse slice header and send slice parameter to driver, the num_ref_idx_active_override_flag is true by default.
short format: application will not parse slice header, it send slice header+slice data to driver directly. slice header will be parsed by driver

va/va.h
```diff
@@ -492,6 +492,20 @@ typedef enum
...
+     */
+    VAConfigAttribMaxPictureWidth     = 18,
+    /**
+     * \brief Maximum picture height. Read-only.
+     *
+     * This attribute determines the maximum picture height the driver supports
+     * for a given configuration.
+     */
+    VAConfigAttribMaxPictureHeight    = 19,
+    /**
```
xhaihao: VA has provided vaQuerySurfaceAttributes() to query all supported attributes for the specified config, including VASurfaceAttribMaxWidth / VASurfaceAttribMinHeight.  are the above 2 attributes needed?

va/va.h
fhvwy: @xhaihao  Argh, the name of that syntax element changed between standard versions - apologies for the confusion there.

I'm still unclear exactly how this feature can work, though.  Do you have any documentation or a working implementation?  I'm particularly curious about how the reference pictures will be handled.

va/va.h
xfengcarl: this is for encoder capability, the maximum resolution that  driver support. the surface attributes is the surface size

xhaihao: vaQuerySurfaceAttributes() has a parameter to specify the VA config, including encoding entrypoint/profile pair. once user get the maximum surface size, user can get the maximum resolution.

va/va.h
xhaihao: @fhvwy I don't have a working implementation or documentation so far. 

I think we can add some additional reserved bytes here. Once we have a workable driver with the new feature, we can reuse these reserved bytes and ABI won't be broken.

va/va.h
```diff
@@ -551,6 +592,23 @@ typedef struct _VAConfigAttrib {
...
 /** \brief Driver supports base mode for slice decoding */
 #define VA_DEC_SLICE_MODE_BASE         0x00000002
+
+/** @name Attribute values for VAConfigAttribDecJPEG */
+/**@{*/
+typedef union _VAConfigAttribValDecJPEG {
+    /** \brief Set to (1 << VA_ROTATION_xxx) for supported rotation angles. */
+    unsigned int rotation;
+    /** \brief Reserved for future use. */
+    unsigned int reserved[3];
```
xhaihao: Please add the comment that the reserved byte must be 0, otherwise we might have issue when re-using these bytes in the future.

va/va.h
```diff
@@ -380,6 +380,47 @@ typedef enum
      * See \c VA_DEC_SLICE_MODE_xxx for the list of slice decoding modes.
      */
     VAConfigAttribDecSliceMode		= 6,
+   /**
+     * \brief JPEG decoding attribute. Read-only.
+     *
+     * This attribute exposes a number of capabilities of the underlying
+     * JPEG implementation. The attribute value is partitioned into fields as defined in the
+     * VAConfigAttribValDecJPEG union.
```
xhaihao: I think VAConfigAttribValDecJPEG should be struct instead of union,  please see my comment below.

va/va.h
xhaihao: reserved[0] is not a reserved byte if VAConfigAttribValDecJPEG is union. So I think you should define VAConfigAttribValDecJPEG as struct instead of union. 

Please add the comment that the reserved byte must be 0, otherwise we will have issue when re-using these bytes in the future.

va/va.h
```diff
@@ -380,6 +380,47 @@ typedef enum
...
+     * all function calls that take in the config (e.g. vaQuerySurfaceAttributes())
+     * or context are in relation to the decoder, except those video processing
+     * function specified in the next paragraph.
+     *
+     * Once the decode config and context are created, the user must further
+     * query the supported processing filters using vaQueryVideoProcFilters(),
+     * vaQueryVideoProcFilterCaps(), vaQueryVideoProcPipelineCaps() by specifying
+     * the created decode context.  The user must provide processing information
+     * and extra processing output surfaces as "additional_outputs" to the driver
+     * through VAProcPipelineParameterBufferType.  The render_target specified
```
xhaihao: Which field in VAProcPipelineParameterBuffer is used as output surface? Could you add it in the comments?

va/va_dec_jpeg.h
```diff
@@ -70,6 +70,10 @@ typedef struct _VAPictureParameterBufferJPEGBaseline {
     }                   components[255];
     /** \brief Number of components in frame (Nf). */
     unsigned char       num_components;
+    /** \brief Input color space 0: YUV, 1: RGB, 2: BGR, others: reserved */
+    unsigned char       color_space;
+    /** \brief Set to VA_ROTATION_* for a single rotation angle reported by VAConfigAttribDecJPEG. */
+    unsigned int 	rotation;
 } VAPictureParameterBufferJPEGBaseline;
```
xhaihao: Are color space and rotation the features for JPEG baseline ?

va/va.h
```diff
@@ -523,6 +523,13 @@ typedef enum
      */
     VAConfigAttribEncQualityRange     = 21,
     /**
+     * \brief Encoding quantization attribute. Read-only.
+     *
+     * This attribute conveys whether the driver supports certain types of quantization methods
+     * for encoding (e.g. trellis).
```
xhaihao: It would be better to add a comment that " See \c VA_ENC_QUANTIZATION_xxx for the list of quantization methods", otherwise a reader will have no idea about the quantization methods.

va/va.h
```diff
@@ -1439,6 +1454,33 @@ typedef struct _VAEncMiscParameterBufferQualityLevel {
 } VAEncMiscParameterBufferQualityLevel;
 
 /**
+ * \brief Quantization settings for encoding.
+ *
+ * Some encoders support special types of quantization such as trellis, and this structure
+ * can be used by the app to control these special types of quantization by the encoder.
```
xhaihao: Should user send this misc parameter per frame/field? could you document it clearly ?

va/va.h
```diff
@@ -1439,6 +1454,33 @@ typedef struct _VAEncMiscParameterBufferQualityLevel {
...
+ */
+typedef struct _VAEncMiscParameterQuantization
+{
+    union
+    {
+    /* if no flags is set then quantization is determined by the driver */
```
xhaihao: Do you mean app doesn't pass VAEncMiscParameterQuantization to the driver? otherwise the flag is always set, either 1 or 0.

va/va.h
```diff
@@ -1439,6 +1454,33 @@ typedef struct _VAEncMiscParameterBufferQualityLevel {
...
+        struct
+        {
+	    /* \brief disable trellis for all frames/fields */
+            unsigned int disable_trellis : 1;
+	    /* \brief enable trellis for I frames/fields */
+            unsigned int enable_trellis_I : 1;
+	    /* \brief enable trellis for P frames/fields */
+            unsigned int enable_trellis_P : 1;
+	    /* \brief enable trellis for B frames/fields */
+            unsigned int enable_trellis_B : 1;
```
xhaihao: Which flag has higher priority, disable_trellis or enable_trellis_xxx ? e.g. disable_trellis = 1 and  enable_trellis_I = 1, how does the driver handle this case?

va/va.h
```diff
@@ -539,9 +651,25 @@ typedef struct _VAConfigAttrib {
 #define VA_RC_CQP                       0x00000010
 /** \brief Variable bitrate with peak rate higher than average bitrate. */
 #define VA_RC_VBR_CONSTRAINED           0x00000020
+/** \brief Intelligent Constant Quality. Provided an initial ICQ_quality_factor,
+ *  adjusts QP at a frame and MB level based on motion to improve subjective quality. */
+#define VA_RC_ICQ			0x00000040
 /** \brief Macroblock based rate control.  Per MB control is decided
  *  internally in the encoder. It may be combined with other RC modes, except CQP. */
 #define VA_RC_MB                        0x00000080
+/** \brief Constant Frame Size */
```
pengche1: VA_RC_MB is also controlled by VAEncMiscParameterRateControl.rc_flags.bits.mb_rate_control. should we delete VA_RC_MB here?

va/va.h
```diff
@@ -530,6 +530,13 @@ typedef enum
      */
     VAConfigAttribEncQuantization     = 22,
     /**
+     * \brief Encoding intra refresh attribute. Read-only.
+     *
+     * This attribute conveys whether the driver supports certain types of intra refresh methods
+     * for encoding (e.g. adaptive intra refresh or rolling intra refresh).
+     */
```
xhaihao: Please add a line of "See \c VA_ENC_INTRA_REFRESH_xxx for xxx" in the comment.

va/va.h
```diff
@@ -1413,6 +1438,51 @@ typedef struct _VAEncMiscParameterAIR
...
+        struct
+	/**
+	 * \brief Indicate if intra refresh is enabled in column/row.
+	 *
+	 * App should query VAConfigAttribEncIntraRefresh to confirm RIR support
+	 * by the driver before sending this structure. The following RIR restrictions
+	 * apply:
+	 *  - No field encoding.
+	 *  - No B frames.
+	 *  - No multiple references.
```
xhaihao: is the above a driver limitation only or the requirement for RIR? If it is a driver limitation,  please remove the above comments and add a mechanism to query the driver capabilities.

va/va.h
```diff
@@ -1413,6 +1438,51 @@ typedef struct _VAEncMiscParameterAIR
...
+	 */
+        {
+	    /* \brief enable RIR in column */
+            unsigned int enable_rir_column : 1;
+	    /* \brief enable RIR in row */
+            unsigned int enable_rir_row : 1;
+	    unsigned int reserved : 30;
```
xhaihao: Please use the same indentation style.

va/va.h
```diff
@@ -1256,6 +1275,68 @@ typedef enum
...
+    unsigned char       reserved0[3];
+    unsigned int        reserved;
+} VAProcessingRateParamsDec;
+
+typedef struct _VAProcessingRateParams {
+    union {
+        VAProcessingRateParamsEnc proc_buf_enc;
+        VAProcessingRateParamsDec proc_buf_dec;
+    };
+} VAProcessingRateParams;
```
xhaihao: Could you replace Params with Parameter to keep coding style consistency?

va/va.h
xfengcarl: @fhvwy ,  1. if there are Slice parameter, DPB is in the PPS, application  will parse slice haeader and set  active reference frame index is in slice parameter,  2. if there are no slice parameter from application, driver should parse the slice header and get the reference order (after ref_pic_list_modification or dec_ref_pic_marking

va/va.h
xfengcarl: @xhaihao you mean use VASurfaceAttribMaxWidth VASurfaceAttribMaxHeight to report maximum resolution ?  I just wonder whether surface resolution is encode/decode resolution.  sometimes, customer create a big surface (Maybe for VP usage) but the container YUV to encoder is smaller.

va/va.h
xfengcarl: seems structure is more suitable , will change it.

va/va.h
xfengcarl: @xhaihao output surface is sent to driver by vaBeginPicture

va/va.h
xfengcarl: accept, will add the comment

va/va.h
xhaihao: User can create a big surface for other usages. here the returned VASurfaceAttribMaxWidth / VASurfaceAttribMaxHeight is the maximum solution for the ***given*** config

va/va.h
xfengcarl: from VAEncMiscParameterQuantization definition, it should be a per-sequence parameter, because it set the enable/disable for  I/P/B separately.

va/va.h
xfengcarl: app will pass it, but if these field is zero, means the trellis is determined by driver.

va/va.h
xfengcarl: disable trellis is higher priority.

va/va.h
xfengcarl: this definition is for attribute . VAEncMiscParameterRateControl.rc_flags.bits.mb_rate_control is setting

va/va.h
```diff
@@ -1158,13 +1398,20 @@ typedef enum
...
     /** \brief Buffer type used for sending skip frame parameters to the encoder's
       * rate control, when the user has externally skipped frames. */
     VAEncMiscParameterTypeSkipFrame     = 9,
     /** \brief Buffer type used for region-of-interest (ROI) parameters. */
     VAEncMiscParameterTypeROI           = 10,
     /** \brief Buffer type used for temporal layer structure */
     VAEncMiscParameterTypeTemporalLayerStructure   = 12,
+    /** \brief Buffer type used for dirty region-of-interest (ROI) parameters. */
+    VAEncMiscParameterTypeDirtyROI      = 13,
+    /** \brief Buffer type used for parallel BRC parameters. */
```
xfengcarl: set ROI information through VAEncMiscParameterTypeROI ,  sometimes, we need a interface to inform driver that some areas changed, and need some optimization, maybe the changes is very lite,then we need DirtyROI

va/va.h
xfengcarl: accept, will add it

va/va.h
xfengcarl: accept

va/va.c
```diff
@@ -856,6 +856,20 @@ VAStatus vaQueryConfigAttributes (
...
+    VAConfigID config_id,
+    VAProcessingRateParams *proc_buf,
+    unsigned int *processing_rate	/* out */
+)
+{
+  VADriverContextP ctx;
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  return ctx->vtable->vaQueryProcessingRate( ctx, config_id, proc_buf, processing_rate);
```
xfengcarl: accept

va/va.h
xfengcarl: accept

va/va.h
xhaihao: According to the comments, the render_target specified by vaBeginPicture() is the decode output surface, not the processing output

va/va.h
xhaihao: How about if app doesn't pass it? Please document it clearly in va.h

va/va.h
xhaihao: if so,  disable_trellis = 1, means trellis isn't supported, disable_trellis=0 means the quantization is determined by the driver. Why is enable_trellis_I / enable_trellis_P / enable_trellis_B needed ?

va/va.h
xhaihao: Got it, please document it in va.h

va/va.h
xhaihao: I am confused too,  According to the comment,  Per MB control is decided  internally in the encoder, why is VAEncMiscParameterRateControl.rc_flags.bits.mb_rate_control needed?

va/va.h
xhaihao: Does VAEncMiscParameterTypeDirtyROI have dependency on VAEncMiscParameterTypeROI ?

va/va.h
```diff
@@ -588,6 +588,19 @@ typedef enum
...
+     * \brief Parallel Rate Control (hierachical B) attribute. Read-only.
+     * This attribute conveys whether the encoder supports parallel rate control.
+     * It is a boolean value 0 - unsupported, 1 - supported.
+     * This is the way when hireachical B frames are encoded, multiple independent B frames
+     * on the same layer may be processed at same time.
+     * If supported, app may enable it by setting enable_parallel_brc in
+     * VAEncMiscParameterRateControl,
+     * and the number of B frames per layer per GOP will be passed to driver
+     * through VAEncMiscParameterParallelRateControl structure.
+     * Currently three layers are defined.
```
xhaihao: How about if more than 3 layer are supported? Could you provide a mechanism to query the supported layer number.

va/va.h
```diff
@@ -632,9 +645,25 @@ typedef struct _VAConfigAttrib {
...
 /** \brief Variable bitrate with peak rate higher than average bitrate. */
 #define VA_RC_VBR_CONSTRAINED           0x00000020
+/** \brief Intelligent Constant Quality. Provided an initial ICQ_quality_factor,
+ *  adjusts QP at a frame and MB level based on motion to improve subjective quality. */
+#define VA_RC_ICQ			0x00000040
 /** \brief Macroblock based rate control.  Per MB control is decided
  *  internally in the encoder. It may be combined with other RC modes, except CQP. */
 #define VA_RC_MB                        0x00000080
+/** \brief Constant Frame Size */
+#define VA_RC_CFS                       0x00000100
```
xhaihao: How does app use CFS? Could you add more comments?

va/va.h
```diff
@@ -1464,6 +1495,7 @@ typedef struct _VAEncMiscParameterRateControl
     /* initial QP at I frames */
     unsigned int initial_qp;
     unsigned int min_qp;
+    unsigned int max_qp;
```
xhaihao: Is this flag the new BRC method related?

va/va.h
```diff
@@ -1477,10 +1509,14 @@ typedef struct _VAEncMiscParameterRateControl
              * The temporal layer that the rate control parameters are specified for.
              */
             unsigned int temporal_id : 8;
-            unsigned int reserved : 17;
+            unsigned int cfs_I_frames : 1; /* I frame also follows CFS */
+            unsigned int enable_parallel_brc    : 1;
+            unsigned int enable_dynamic_scaling : 1;
```
xhaihao: What is dynamic_scaling? could you document it in va.h

va/va.h
```diff
@@ -1294,6 +1294,7 @@ typedef enum
     VAEncMiscParameterBufferType	= 27,
     VAEncMacroblockParameterBufferType	= 28,
     VAEncMacroblockMapBufferType        = 29,
+    VAEncQpBufferType                   = 30,
```
xhaihao: How does user use this buffer? Could you document the buffer format in va.h? Otherwise please remove this buffer type.

va/va.h
```diff
@@ -2350,6 +2351,8 @@ typedef struct _VAPictureParameterBufferH264
         unsigned int value;
     } pic_fields;
     unsigned short frame_num;
+    unsigned char num_ref_idx_l0_default_active_minus1;
+    unsigned char num_ref_idx_l1_default_active_minus1;
 } VAPictureParameterBufferH264;
 
 /** H.264 Inverse Quantization Matrix Buffer */
```
xhaihao: Please update the commit log, I don't see NV21 forcc in your commit.

va/va.h
xfengcarl: yes, there are additional surface for post-process output surface, it 's in another commit, the pull request "add new vpp features", will combined the related changes in this PR. thanks

va/va.h
xfengcarl: accept

va/va_dec_jpeg.h
xfengcarl: @xhaihao these feature is post process feature, it is not related with JPEG itself.  vaapi profile definition only has baseline definition.

va/va.h
xfengcarl: accept

va/va.h
xfengcarl: it's a quality/performance related feature, it should be handled inside driver by the quality level. but if application want to customize their quality. they can enable/disable it.

va/va.h
xfengcarl: if the disable_trellis is zero, then driver should check enable_trellis_I/P/B, all is zero , it will be determined by driver., if the disable_trellis is 1, trellis will be disabled, and will not check enalbe_trellis_IPB anymore

va/va.h
xfengcarl: each MB's Qp is determined by driver, but  the feature is enabled/disabled by application. if disabled, the QP in one frame will be same. if enabled, the Qp of each MB in different area of one frame will changed by driver to improve quality and compress rate.

va/va.h
xfengcarl: sorry for misunderstanding, actually, dirty ROI is totally different with ROI, when there are dirty ROI, all of the ME will be disabled except dirty ROI rectangle area. it most likely is used by remote desktop. only dirty rectangle need to be do ME again.

va/va.h
xfengcarl: accept.

va/va.h
```diff
@@ -563,6 +563,16 @@ typedef enum
...
      */
     VAConfigAttribEncRateControlExt   = 26,
+    /**
+     * \brief Processing rate reporting attribute. Read-only.
+     *
+     * This attribute conveys whether the driver supports reporting of
+     * encode/decode processing rate based on certain set of parameters
+     * (i.e. levels, I frame internvals) for a given configuration.
+     * If this is supported, vaQueryProcessingRate() can be used to get
+     * encode or decode processing rate.
```
xfengcarl: accept

va/va.h
xhaihao: If so, could you add VAConfigAttribDecProcessing after the commit which adds new vpp features?  BTW why not ask  user to creates two VA contexts, one for decoding, another for VPP?

va/va_dec_jpeg.h
xhaihao: How does user know the valid color_space for JPEG decoding?

xhaihao: How about to define VAConfigAttribDecJPEG as read/write attribute? User can use VAConfigAttribDecJPEG to query the valid color space and rotation, then set the required value via VAConfigAttribDecJPEG when decoding JPEG image. If so, we won't  change VAPictureParameterBufferJPEGBaseline.

xhaihao: It would be better not to mix up two different features in a commit. could you please split this commit into 2 commits, one for VAConfigAttribDecJPEG, another for VAConfigAttribDecProcessing?

In addition, I don't think VAConfigAttribDecProcessing is necessary because user can create 2 contexts.

va/va.h
xhaihao: I see, could you please document it in va.h?

xhaihao: Could yo change the type of value to uint64_t so that we have more reserved bits for future extension? I will change existent data structures as well.

va/va.h
xhaihao: It is easy to confuse ROI with DirtyROI, could you rename DirtyROI to other-name, e.g. DirtyRegion and change the corresponding comments and variable name?

va/va.h
xfengcarl: so, user can create a bigger surface than  VASurfaceAttribMaxWidth & VASurfaceAttribMaxHeight , right? it sound a little weird.

va/va_dec_jpeg.h
xfengcarl: it will be a per-sequence setting? currently, it's a per-frame setting.

va/va.h
xfengcarl: sure, I can change it to be uint64_t, but I wonder whether there are too much bits required for quantization

va/va.h
xfengcarl: 1. yes. 2. it will not take effect for CQP mode. 
we are discussing two features
1. MBBRC feature ---- it can only be enabled with BRC mode, CQP is not BRC mode
2. perMB Qp setting ---- only for CQP mode without BRC

per-mb Qp

va/va.h
xhaihao: User must specify a ***config*** when calling vaQuerySurfaceAttributes(), so VASurfaceAttribMaxWidth/VASurfaceAttribMaxHeight is valid only for the ***given config***, e.g. for VAProfileH264Main/VAEntrypointVLD, VASurfaceAttribMaxWidth/VASurfaceAttribMaxHeight may be 1920/1080, but for VAProfileNone/VAEntrypointVideoProc, VASurfaceAttribMaxWidth/VASurfaceAttribMaxHeight  may be 4K/4K.  The API doesn't require VASurfaceAttribMaxWidth/VASurfaceAttribMaxHeight must be same for all ***config***

xfengcarl: This should be two different things, one is for codec capability (each codec would be different), one is for generic surface capability. A larger surface can be used for smaller pictures. BTW, is vaQuerySurfaceAttributes() required before creating the surface?

va/va_dec_jpeg.h
xfengcarl: 1.Jpeg itself doesn’t have notion of color space, the info is in user message. But for practical usage, you would need to have those info. Unlike other video codec, Jpeg can be coded in various formats but most common usage would be 420/422 or RGB which can be determined by output surface format. Without this info, we wouldn’t be able to convert to the correct output format, it’s not an Intel specific thing. 
2. Rotation is a nice feature that we can support it . Moving it out would make rotation generic and then we would need to have an attribute for each codec to indicate if rotation is supported or not. If we move it out to make it specific to Jpeg, that doesn’t make much sense

va/va.h
xfengcarl: will change the name to be DirtyRect. is it ok?

va/va.h
xfengcarl: will add such comments:
    /**
     * \brief Dynamic Scaling Attribute. Read-only.
     * This attribute conveys whether encoder is capable to determine dynamic frame
     * resolutions adaptive to bandwidth utilization and processing power, etc., as well
     * as capable to scale raw source and reference frames.
     * It is a boolean value 0 - unsupported, 1 - supported.
     * If it is supported, app may enable it by setting enable_dynamic_scaling in 
     * VAEncMiscParameterRateControl.
     */

xfengcarl: will change the code and add an attribute with comments
    /**
     * \brief Dynamic Scaling Attribute. Read-only.
     * This attribute conveys whether encoder is capable to determine dynamic frame
     * resolutions adaptive to bandwidth utilization and processing power, etc., as well
     * as capable to scale raw source and reference frames.
     * It is a boolean value 0 - unsupported, 1 - supported.
     * If it is supported, app may enable it by setting enable_dynamic_scaling in 
     * VAEncMiscParameterRateControl.
     */
     VAConfigAttribEncDynamicScaling        = 30,

va/va.h
xfengcarl: has changed the comment of VAConfigAttribEncParallelRateControl, it will return the maximum return layer number

va/va.h
xfengcarl: it's a BRC feature, can enabled with other BRC methods

va/va.h
xfengcarl: accept

va/va_dec_jpeg.h
xfengcarl: accept

va/va_enc_mpeg2.h
```diff
@@ -287,6 +287,17 @@ typedef struct _VAEncSliceParameterBufferMPEG2 {
     uint32_t                va_reserved[VA_PADDING_LOW];
 } VAEncSliceParameterBufferMPEG2;
 
+typedef struct _VAEncMiscParameterExtensionDataSeqDisplayMPEG2
+{
+    uint8_t extension_start_code_identifier;
```
sreerenjb: Why to include the "extension_start_code_identifier" ? Seems redundant to me..

va/va_enc_mpeg2.h
```diff
@@ -287,6 +287,17 @@ typedef struct _VAEncSliceParameterBufferMPEG2 {
     uint32_t                va_reserved[VA_PADDING_LOW];
 } VAEncSliceParameterBufferMPEG2;
 
+typedef struct _VAEncMiscParameterExtensionDataSeqDisplayMPEG2
```
sreerenjb: Is it the responsibility of the driver to include the Packed header? Or does the middleware supposed to insert them? In any case, we should comment it.

va/va.h
```diff
@@ -2003,6 +2015,142 @@ typedef struct _VAEncMiscParameterParallelRateControl {
 } VAEncMiscParameterParallelRateControl;
 
 /**
+ * \brief MB partition modes and 1/2 1/4 motion search configuration
+ *
+ * Specifies MB partition modes that are disabled. Specifies Half-pel
+ * mode and Quarter-pel mode searching
+ */
+typedef struct _VAEncMiscParameterSubMbPartPel
+{
```
sreerenjb: Is this generic enough to cover all codecs? I don't think so, It rather specific to h264. 
Even the union name "inter_sub_mb_partition_mask" is so h264 specific.
If we really want to use this for all codecs in future, at least the "reserved" field should be extended with more size (so that we can add other fields in future without breaking the ABI)and  rename "inter_sub_mb_partition_mask" as "inter_partition_mask"

How about using a single 32bit Flag so that we can easily do the bit masking for all combinations?

va/va.h
```diff
@@ -2003,6 +2015,142 @@ typedef struct _VAEncMiscParameterParallelRateControl {
...
+
+    /**
+     * \brief Precison of motion search
+     * 0:Integer mode searching
+     * 1:Half-pel mode searching
+     * 2:Reserved
+     * 3:Quarter-pel mode searching
+     */
+    uint32_t enable_sub_pel_mode;
+    uint8_t sub_pel_mode;
```
sreerenjb: Is it by default "Integer mode searching"? if so why we need "enable_sub_pel_mode"?

va/va.h
```diff
@@ -2003,6 +2015,142 @@ typedef struct _VAEncMiscParameterParallelRateControl {
...
+// per frame encoder controls, once set they will persist for all future frames
+// till it is updated again.
+typedef struct _VAEncMiscParameterEncQuality
+{
+    union
+    {
+        struct
+        {
+            // Use raw frames for reference instead of reconstructed frames.
+            uint32_t useRawPicForRef                    : 1;
```
sreerenjb: Wondering for what purpose this can be used!
Running only VME in FEI mode is something which can make use of this. But is it useful in general?

va/va.h
```diff
@@ -2003,6 +2015,142 @@ typedef struct _VAEncMiscParameterParallelRateControl {
...
+            //      3: scale MV cost to be 1/8 of the default value for HME predictor.
+            uint32_t HMEMVCostScalingFactor             : 2;
+            //disable HME
+            uint32_t HMEDisable                         : 1;
+            //disable Super HME
+            uint32_t SuperHMEDisable                    : 1;
+            //disable Ultra HME
+            uint32_t UltraHMEDisable                    : 1;
+            //disable panic mode
+            uint32_t PanicModeDisable                   : 1;
```
sreerenjb: What is panic mode? 
Is this Intel-specific?

va/va.h
xfengcarl: yes, we should merge the dependency PR first

va/va_enc_mpeg2.h
xfengcarl: application query whether it driver support packed header insertion by VAConfigAttribEncPackedHeaders.  if not support , driver will packed the header. this information is used for pack header inside driver

va/va_enc_mpeg2.h
xfengcarl: extension_start_code_identifier is used to distinguish the start code of the extension, as the spec, there are sequence display extension, quant matrix extension,sequence salable extension etc.

xfengcarl: Table 6-2. extension_start_code_identifier codes.

va/va.h
xfengcarl: it's for H264, will move it into va_enc_h264.h and rename to  VAEncMiscParameterSubMbPartPelH264

va/va.h
xfengcarl: by default , subpel mode is set inside driver, for  example, driver may set different value for different quality level. if enable_sub_pel_mode is 1, then application will control sub pel mode, else, will keep the default value.

va/va.h
xfengcarl: this structure is for quality tuning without FEI. we can set to useRawPicForRef during ME stage, and also can set FTQ enhanced skip estimation, Downscaling ME ....

va/va.h
xfengcarl: panic mode is not intel-specific, for the HRD , if BRC need a very small size frame, and the QP is 51 already. it may trigger panic , discarding all AC coefficients (CBP=0),  it will reduce a lot of detail information. I will add comment for this mode, I also can  change the parameter name to discardCoffforBRC  ect.

va/va_enc_mpeg2.h
sreerenjb: I know there is extension_start_code_identifier for all extensions, but my question is about the requirement of this filed in structure. It is very clearly mentioned as "ExtensionDataSeqDisplayMPEG2" so there only one possible predefined value for extension_start_code_identifier.  We don't really include the start_code(s) in any public structures, do we?

va/va.h
sreerenjb: Currently, middleware implementations only keep references(to make sure it doesn't get destroyed before other frames use it for reference) to the reconstructed surfaces. So,
If the user enables "useRawPicForRef", middleware supposed to keep references to the input raw surface and also to the reconstructed surfaces.  Or does it mean that the middleware is not required to fill the scratch surface in each picture parameter for the reconstructed surface?

va/va.h
sreerenjb: Got it, thanks. Please add more comments on fields like this. There are other vendors who have working drivers on top of VA-API :) How about "BrcPanicModeDisable"(if it is only for BRC). ?
Honestly, I am fine with any name if there are enough comments explaining what it means.

va/va.h
xfengcarl: it depend on driver implementation. if driver keep the map between raw surface & recon surface, when the useRawPicForRef is set, it will find the related input surface by the recon surface index.

va/va.h
xfengcarl: ok, will update the comments, thanks

va/va.h
```diff
@@ -1333,6 +1353,68 @@ typedef enum
...
+} VAProcessingRateParameter;
+
+/**
+ * \brief Queries processing rate for the supplied config.
+ *
+ * This function queries the processing rate based on parameters in
+ * \c proc_buf for the given \c config. Upon successful return, the processing
+ * rate value will be stored in \c processing_rate. Processing rate is
+ * specified as the number of macroblocks per second.
+ *
```
sreerenjb: Please add what is the expected code block for each codec since mb is h264 specific.

va/va_enc_mpeg2.h
xfengcarl: yes, so there are two method to define the structure, 1. define two structure , first one is structure Mpeg2SeqExtension{ uint8_t extension_start_code_identifier; uint8_t data[]} another structure is typedef structure ExtensionDisplay{...). 2. define extension_xxx directly with first byte is extension_start_code_identifier. using method 1 will let there are multiple data[] definition in the VAEncMiscParameterBuffer ( Misc type, Misc data, Extension type, extension data).   so I choose second one. I will add comments about the default value of extension_start_code_identifier.

va/va.h
xfengcarl: yes, accept

PR #70, opened 2017-06-16
add new parameter for H264 & VP8 encoder

xfengcarl:
add more vui parameters for H264 , such as fix bit rate ...
add more controls for VP8, reference order ...

### DIFF
--- a/va/va_enc_h264.h
+++ b/va/va_enc_h264.h
@@ -246,8 +246,14 @@ typedef struct _VAEncSequenceParameterBufferH264 {
             uint32_t log2_max_mv_length_horizontal          : 5;
             /** \brief Range: 0 to 16, inclusive. */
             uint32_t log2_max_mv_length_vertical            : 5;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            uint32_t fixed_frame_rate_flag                  : 1;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            uint32_t low_delay_hrd_flag                     : 1;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            uint32_t motion_vectors_over_pic_boundaries_flag: 1;
             /** \brief Reserved for future use, must be zero */
-            uint32_t reserved                               : 19;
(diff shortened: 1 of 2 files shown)


va/va_enc_h264.h
```diff
@@ -241,6 +241,12 @@ typedef struct _VAEncSequenceParameterBufferH264 {
             unsigned int log2_max_mv_length_horizontal          : 5;
             /** \brief Range: 0 to 16, inclusive. */
             unsigned int log2_max_mv_length_vertical            : 5;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            unsigned int fixed_frame_rate_flag                  : 1;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            unsigned int low_delay_hrd_flag                     : 1;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            unsigned int motion_vectors_over_pic_boundaries_flag: 1;
         } bits;
```
pengche1: It will be more readable to add reserved bits at the end of struct{}bits.

xfengcarl: accept

va/va_enc_vp8.h
```diff
@@ -128,7 +128,20 @@ typedef struct  _VAEncPictureParameterBufferVP8
             unsigned int no_ref_gf                      : 1;
             /* don't reference the alternate reference frame */
             unsigned int no_ref_arf                     : 1;
-            unsigned int reserved                       : 28;
+            /* The temporal id the frame belongs to. */
+            unsigned int temporal_id                    : 8;
+            /**
+            *  following two flags indicate the reference order
```
xuguangxin: Is 0 a valid value for first_ref/second_ref?People usually memset the structure and set some value need by them. If 0 is not a valid value for them. It may introduce inconvenience.

xfengcarl: temporary remove them

xfengcarl: it can be 0, 0 means it will take no effect

PR #73, opened 2017-06-19
configure fails if requirement can't be met for user's explicit request

xhaihao:
Notify user an error if user provides --enable-x11/--enable-glx/--enable-egl/--enable-wayland
however the requirement can't be met. drm has been checked mandatorily
in the script


configure.ac
```diff
@@ -227,11 +227,13 @@ AM_CONDITIONAL(USE_DRM, test "$USE_DRM" = "yes")
 
 # Check for X11
 USE_X11="no"
-if test "$enable_x11" = "yes"; then
-    USE_X11="yes"
-    PKG_CHECK_MODULES([X11],    [x11],    [:], [USE_X11="no"])
-    PKG_CHECK_MODULES([XEXT],   [xext],   [:], [USE_X11="no"])
-    PKG_CHECK_MODULES([XFIXES], [xfixes], [:], [USE_X11="no"])
+if test "x$enable_x11" != "xno"; then
+    PKG_CHECK_MODULES([X11],    [x11 xext xfixes],    [USE_X11="yes"], [:])
```
uartie: `XEXT` and `XFIXES` will now be included in the `X11_CFLAGS`.  Thus, `XEXT_CFLAGS` and `XFIXES_CFLAGS` need to be removed from `AM_CPPFLAGS` in `va/x11/Makefile.am` since they will no longer be defined.

configure.ac
```diff
@@ -227,11 +227,13 @@ AM_CONDITIONAL(USE_DRM, test "$USE_DRM" = "yes")
...
+
+    if test "x$USE_X11" = "xno" -a "x$enable_x11" = "xyes"; then
+       AC_MSG_ERROR([VA/X11 explicitly enabled, but can't find x11, xext or xfixes])
```
uartie: Should you include $X11_PKG_ERRORS in the error message instead (like you did for wayland)?

xhaihao:
@uartie Thanks for the comment, I updated the commit.

uartie:
LGTM, but i did not verify

xhaihao:
@uartie I did some testings with/without x11, wayland, glx and egl.

PR #74, opened 2017-06-19
Miscellaneous changes for libva2

fhvwy:
* Make logging callbacks usable in libraries - currently the global state only allows them to be used in end-user executables.
* Remove the H.264 FMO support which will never be used.
* Add namespace prefixes to globals which leak from the shared libraries.
* Some minor cleanup around all of that.

The Android parts aren't even build-tested (though hopefully that change was mechanical).

### DIFF
--- a/va/android/va_android.cpp
+++ b/va/android/va_android.cpp
@@ -26,6 +26,7 @@
 #include "sysdeps.h"
 #include "va.h"
 #include "va_backend.h"
+#include "va_internal.h"
 #include "va_trace.h"
 #include "va_fool.h"
 #include "va_android.h"
@@ -135,13 +136,11 @@ VADisplay vaGetDisplay (
         /* create new entry */
         VADriverContextP pDriverContext = 0;
         struct drm_state *drm_state = 0;
-        pDisplayContext = (VADisplayContextP)calloc(1, sizeof(*pDisplayContext));
+        pDisplayContext = va_newDisplayContext();
         pDriverContext  = (VADriverContextP)calloc(1, sizeof(*pDriverContext));
         drm_state       = (struct drm_state*)calloc(1, sizeof(*drm_state));
         if (pDisplayContext && pDriverContext && drm_state)
         {
-            pDisplayContext->vadpy_magic = VA_DISPLAY_MAGIC;          
-
             pDriverContext->native_dpy       = (void *)native_dpy;
             pDriverContext->display_type     = VA_DISPLAY_ANDROID;
             pDisplayContext->pDriverContext  = pDriverContext;
--- a/va/android/va_android.cpp
+++ b/va/android/va_android.cpp
@@ -165,9 +165,6 @@ VADisplay vaGetDisplay (
     return dpy;
 }
 
-#define CTX(dpy) (((VADisplayContextP)dpy)->pDriverContext)
-#define CHECK_DISPLAY(dpy) if( !vaDisplayIsValid(dpy) ) { return VA_STATUS_ERROR_INVALID_DISPLAY; }
-
 
 extern "C"  {
     extern int fool_postp; /* do nothing for vaPutSurface if set */
--- a/va/android/va_android.cpp
+++ b/va/android/va_android.cpp
@@ -166,8 +166,8 @@ VADisplay vaGetDisplay (
 
 
 extern "C"  {
-    extern int fool_postp; /* do nothing for vaPutSurface if set */
-    extern int trace_flag; /* trace vaPutSurface parameters */
+    extern int va_fool_postp; /* do nothing for vaPutSurface if set */
+    extern int va_trace_flag; /* trace vaPutSurface parameters */
 
     void va_TracePutSurface (
         VADisplay dpy,
--- a/va/va.c
+++ b/va/va.c
@@ -50,9 +50,6 @@
 
 #define DRIVER_EXTENSION	"_drv_video.so"
 
-#define CTX(dpy) (((VADisplayContextP)dpy)->pDriverContext)
-#define CHECK_DISPLAY(dpy) if( !vaDisplayIsValid(dpy) ) { return VA_STATUS_ERROR_INVALID_DISPLAY; }
-
 #define ASSERT		assert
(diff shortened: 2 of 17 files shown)


fhvwy:
To go with the FMO change, I'm wondering whether VAProfileH264Baseline should also be removed because it's similarly unsupportable?  That might have more compatibility issues, though.

va/va.c
```diff
@@ -114,75 +114,100 @@ int vaDisplayIsValid(VADisplay dpy)
...
-            vaSetErrorCallback(NULL);
-        }
-
-        if (strcmp(env_value, "1") == 0) {
-            vaSetInfoCallback(NULL);
-        }
+        ret = sscanf(env_value, "%d", &default_log_level);
+        if (ret < 1)
+            default_log_level = 2;
     }
```
xhaihao: It would be better to limit default_log_level to 0, 1, 2 when user set other values in the config file or environment.

va/va_trace.c
```diff
@@ -2672,9 +2672,6 @@ static void va_TraceVAPictureParameterBufferH264(
     va_TraceMsg(trace_ctx, "\tmb_adaptive_frame_field_flag = %d\n", p->seq_fields.bits.mb_adaptive_frame_field_flag);
     va_TraceMsg(trace_ctx, "\tdirect_8x8_inference_flag = %d\n", p->seq_fields.bits.direct_8x8_inference_flag);
     va_TraceMsg(trace_ctx, "\tMinLumaBiPredSize8x8 = %d\n", p->seq_fields.bits.MinLumaBiPredSize8x8);
-    va_TraceMsg(trace_ctx, "\tnum_slice_groups_minus1 = %d\n", p->num_slice_groups_minus1);
-    va_TraceMsg(trace_ctx, "\tslice_group_map_type = %d\n", p->slice_group_map_type);
-    va_TraceMsg(trace_ctx, "\tslice_group_change_rate_minus1 = %d\n", p->slice_group_change_rate_minus1);
     va_TraceMsg(trace_ctx, "\tpic_init_qp_minus26 = %d\n", p->pic_init_qp_minus26);
     va_TraceMsg(trace_ctx, "\tpic_init_qs_minus26 = %d\n", p->pic_init_qs_minus26);
     va_TraceMsg(trace_ctx, "\tchroma_qp_index_offset = %d\n", p->chroma_qp_index_offset);
```
xhaihao: I agree FMO is useless, but these fields are used in some upper-layer libraries, e.g. libyami, gstreamer-vaapi. how about adding a 'deprecated' indicator instead of removing these fields in this version?

fhvwy: Ok: will keep the fields with a deprecation marker, but still remove the map buffer comment and the trace output.

va/va.c
fhvwy: Ok: will set default_log_level to 2 unless it is explicitly 0 or 1.

va/android/va_android.cpp
```diff
@@ -135,13 +136,11 @@ VADisplay vaGetDisplay (
         /* create new entry */
         VADriverContextP pDriverContext = 0;
         struct drm_state *drm_state = 0;
-        pDisplayContext = (VADisplayContextP)calloc(1, sizeof(*pDisplayContext));
+        pDisplayContext = va_newDisplay();
```
chivakker: va_newDisplayContext for the function name?

va/va.c
```diff
@@ -229,36 +253,50 @@ void va_infoMessage(const char *msg, ...)
...
         free(dynbuf);
     }
     else if (len > 0)
-        va_log_info(buf);
+        dctx->info_callback(dctx->info_callback_user_context, buf);
 #endif
 }
 
-static bool va_checkVtable(void *ptr, char *function)
+VADisplayContextP va_newDisplay(void)
```
chivakker: va_newDisplayContext for the name as it won't get confused with it returning a vaDisplayID

va/va_internal.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef VA_INTERNAL_H
+#define VA_INTERNAL_H
+
+#define CTX(dpy) (((VADisplayContextP)dpy)->pDriverContext)
+#define CHECK_DISPLAY(dpy) if( !vaDisplayIsValid(dpy) ) { return VA_STATUS_ERROR_INVALID_DISPLAY; }
+
+void va_errorMessage(VADisplay dpy, const char *msg, ...);
+void va_infoMessage(VADisplay dpy, const char *msg, ...);
+
```
chivakker: can these definitions be moved to va_trace.h?

va/android/va_android.cpp
fhvwy: Yes, that would be clearer - I'll change it.

va/va_internal.h
fhvwy: I don't think they go there?  The direct logging facility is distinct from all of the tracing infrastructure, and the code is in va.c.

That said, the code could move into va_trace.c at the same time if you prefer (or a new file - both of the existing ones are pretty large already).

xhaihao: Yes, va_infoMessage() / va_errorMessage() is used in other components, I prefer not to move these functions to va_trace.h

fhvwy:
Updated with comments above.

Two additions:
* A macro to indicate deprecation, so that the user gets a warning when compiling code using deprecated fields rather than having to read the documentation to find them.
* A note about renamed fields, to avoid confusion like I had in #69.

xhaihao:
It would be better to deprecate VAProfileH264Baseline as well.  Beginning with GCC 6 it is easy to deprecate an enum item .  But I don't know how to deprecate an enum item with an old version of GCC.

fhvwy:
I don't think it's possible in older versions.  I'm not sure that matters too much, though - only getting the warning on newer gcc versions is better than never getting it at all.

How about:
```
#ifdef __GNUC__
#define va_deprecated __attribute__((deprecated))
#if __GNUC__ >= 6
#define va_deprecated_enum va_deprecated
#else
#define va_deprecated_enum
#endif
#else
#define va_deprecated
#define va_deprecated_enum
#endif
```
and then we can write:
```
    VAProfileH264Baseline va_deprecated_enum = 5,
```
to have the warning on GCC >= 6.

xhaihao:
OK, I am fine to add va_deprecated_enum.

fhvwy:
Updated with va_deprecated_enum.

PR #82, opened 2017-07-06
add attribute value for VAConfigAttribEncSliceStructure, redefine VA_…

xfengcarl:
…ENC_SLICE_STRUCTURE_ARBITRARY_ROW to non-zero value

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -648,12 +648,16 @@ typedef struct _VAConfigAttrib {
 
 /** @name Attribute values for VAConfigAttribEncSliceStructure */
 /**@{*/
-/** \brief Driver supports an arbitrary number of rows per slice. */
-#define VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS           0x00000000
 /** \brief Driver supports a power-of-two number of rows per slice. */
 #define VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS        0x00000001
 /** \brief Driver supports an arbitrary number of macroblocks per slice. */
 #define VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS    0x00000002
+/** \brief Dirver support 1 rows  per slice */
+#define VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS               0x00000004
+/** \brief Dirver support max encoded slice size per slice */
+#define VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE           0x00000008
+/** \brief Driver supports an arbitrary number of rows per slice. */
+#define VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS           0x00000010
 /**@}*/
 
 /** \brief Attribute value for VAConfigAttribEncJPEG */


va/va.h
```diff
@@ -597,12 +597,16 @@ typedef struct _VAConfigAttrib {
 
 /** @name Attribute values for VAConfigAttribEncSliceStructure */
 /**@{*/
-/** \brief Driver supports an arbitrary number of rows per slice. */
-#define VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS           0x00000000
 /** \brief Driver supports a power-of-two number of rows per slice. */
 #define VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS        0x00000001
 /** \brief Driver supports an arbitrary number of rows per slice. */
```
QuPengfei: comments should be " Driver supports an arbitrary number of MBs per slice" for #define VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS , right?

va/va.h
```diff
@@ -597,12 +597,16 @@ typedef struct _VAConfigAttrib {
...
 #define VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS        0x00000001
-/** \brief Driver supports an arbitrary number of rows per slice. */
+/** \brief Driver supports an arbitrary number of MBs per slice. */
 #define VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS    0x00000002
+/** \brief Dirver support 1 rows per slice */
+#define VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS               0x00000004
+/** \brief Dirver support max encoded slice size */
+#define VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE           0x00000008
+/** \brief Driver supports an arbitrary number of rows per slice. */
+#define VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS           0x00000010
```
xhaihao: Can all flags be OR'ed together? If not, could you please document it clearly?  How about to define VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS to 0x08 because VA_ENC_SLICE_STRUCTURE_POWER_OF_TWO_ROWS / VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS / VA_ENC_SLICE_STRUCTURE_EQUAL_ROWS / VA_ENC_SLICE_STRUCTURE_ARBITRARY_ROWS define slice structure per slice.

mleise: In English whenever he/she/it is doing something an 's' needs to be appended to the verb: "_[The]_ driver support**s**". Also: "1 row" without 's'.

Does `VA_ENC_SLICE_STRUCTURE_MAX_SLICE_SIZE` mean that the driver allows you to *change* the maximum slice size? In that case it is the only flag that can be combined with others, right? I would interpret the other 4 flags as `ARBITRARY_MACROBLOCKS > ARBITRARY_ROWS > POWER_OF_TWO_ROWS > EQUAL_ROWS` where each flag implicitly enables anything to the right of it. So a VAAPI client would test these flags from left to right and stop at the first match. I would consider it a driver error, if none of them was set.

But then it would better be represented with one of those bit-fields as was done for other attributes, with one bit for the `maxSliceSize` and a 2 (or more) bit unsigned int for the `sliceStructure`.

xhaihao:
@xfengcarl  could you help rebase the patch? I can't resolve the conflicts via github

PR #84, opened 2017-07-09
va: Add the fourcc of I420 format

mypopydev:
i965 will return the I420 in vaQuerySurfaceAttributes, but va.h didn't
add the pre-define fourcc

### DIFF
--- a/va/glx/va_glx_impl.c
+++ b/va/glx/va_glx_impl.c
@@ -205,7 +205,7 @@ static int check_extension3(const char *name)
     for(i = 0; i < nbExtensions; i++)
     {
         const GLubyte *strExtension = glGetStringi(GL_EXTENSIONS, i);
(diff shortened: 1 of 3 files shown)


fhvwy:
This is identical to the existing IYUV format?  I think we should decide on one name and use it consistently.  I'm not sure which it should be, though (i965 unhelpfully seems to use both) - if the answer is I420 then please deprecate IYUV when making this change so that it gets used uniformly in future.

xhaihao:
Yes, according to https://www.fourcc.org/yuv.php, IYUV is a duplicate format of I420 and it seems most user prefer I420.

xhaihao:
@mypopydev could you deprecate  IYUV in your commit? I'd like to merge this patch to v2.0-branch instead of master branch.

mypopydev:
Close this PR because this patch have merged to v2. 0-branch.

PR #85, opened 2017-07-09
Add a more flexible interface for import/export of DRM objects

### LINKED ISSUE #10: Cannot export as dmabuf a VASurface backed with non-contiguous memory
migrated from Bugzilla [#98285](https://bugs.freedesktop.org/show_bug.cgi?id=98285)
status ASSIGNED severity *enhancement* in component *core* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: Sean V Kelley

On 2016-10-17 10:10:12 +0000, Julien Isorce wrote:

> Currently the work-flow is:
> 
> VASurface -> vaDeriveImage -> VAImageBufferType -> VABufferID -> vaAcquireBufferHandle -> 1 dmabuf FD value.
> 
> This only work if the underlying memory is contiguous. But if the VASurface is backed by at least 2 non-contiguous memory (for example one for each NV12 plane) then it would require to export each plane with a different dmabuf FD value.
> 
> I think a solution could be to create to either create a new equivalent function vaAcquireBufferHandle2 that takes a plane id into parameter. Or a new member in VABufferInfo. Or a new function that vaGetPlaneBufferID(VAImageID img, int plane, VABufferID *out_buf).
> 
> Note that in vdpau it is supported:
> 
> struct VdpSurfaceDMABufDesc {
>    /* DMA-buf file descriptor */
>    uint32_t handle;
>    /* Width in pixel */
>    uint32_t width;
>    /* Height in pixel */
>    uint32_t height;
>    /* Offset in bytes */
>    uint32_t offset;
>    /* Stride in bytes */
>    uint32_t stride;
>    /* VDP_RGBA_FORMAT_* as defined in the VDPAU spec and above. */
>    uint32_t format;
> };
> 
> VdpStatus VdpVideoSurfaceDMABuf(
>    VdpVideoSurface               surface,
>    VdpVideoSurfacePlane          plane,
>    struct VdpSurfaceDMABufDesc * result)

On 2016-10-24 23:01:10 +0000, Sean V Kelley wrote:

> Yes, this is an isssue for use cases where we split off the y/u/v into separate planes and then we require three FD.  It's on my list of changes.
> 
> I agree, this is a needed change.

On 2016-10-25 02:07:01 +0000, haihao wrote:

> (In reply to Julien Isorce from comment # 0)
> > Currently the work-flow is:
> > 
> > VASurface -> vaDeriveImage -> VAImageBufferType -> VABufferID ->
> > vaAcquireBufferHandle -> 1 dmabuf FD value.
> > 
> > This only work if the underlying memory is contiguous. But if the VASurface
> > is backed by at least 2 non-contiguous memory (for example one for each NV12
> > plane) then it would require to export each plane with a different dmabuf FD
> > value.
> > 
> 
> Do you mean a VA surface is backed with two different buffer objects in your driver? In libva-intel-driver each surface is created with only one GEM buffer object, so the same dmabuf FD is exported for all planes. You can get the layer out via the derived image.

On 2016-10-26 09:06:47 +0000, Julien Isorce wrote:

> (In reply to Sean V Kelley from comment # 1)
> > Yes, this is an isssue for use cases where we split off the y/u/v into
> > separate planes and then we require three FD.  It's on my list of changes.
> > 
> > I agree, this is a needed change.
> 
> I am glad to hear that. Thx
> 
> (In reply to haihao from comment # 2)
> > Do you mean a VA surface is backed with two different buffer objects in your
> > driver? In libva-intel-driver each surface is created with only one GEM
> > buffer object, so the same dmabuf FD is exported for all planes. You can get
> > the layer out via the derived image.
> 
> Yes

On 2016-10-27 18:14:07 +0000, Sean V Kelley wrote:

> Julien,
> 
> Can you work under the existing constraints of VA Surface or do you need something more?
> 
> My requirements are different as they pertain to ChromeOS.
> 
> Thanks,
> 
> Sean

On 2016-10-27 20:31:32 +0000, Julien Isorce wrote:

> (In reply to Sean V Kelley from comment # 4)
> > Julien,
> > 
> > Can you work under the existing constraints of VA Surface or do you need
> > something more?
> > 
> 
> 
> Hi, I currently workaround it with an extra step: gstvaapipostproc to convert it to RGB since I know it will be one contiguous memory in that case. Then I export it a dmabuf and import into an EGLImage later on through gstglimagesink.
> 
> So currently in vaapi mesa gallium I only allow to derive_image for RGB surfaces. 
> 
> > My requirements are different as they pertain to ChromeOS.
> > 
> > Thanks,
> > 
> > Sean
> 
> I still need this change. What has changed since # 1 ? :)
> 
> Thx!

On 2016-10-28 21:11:42 +0000, Sean V Kelley wrote:

> Hi Julien,
> 
> In ChromeOS we would always have to convert a decoded video surface (YUV) to RGB format for composition.  That in itself incurs a cost.  Hardware overlays which supported YUV and newer platforms with display out support for YUV needed ways to work with graphics stack and video stack for textures. So we looked into Sub-bo handling for planar fb formats.  It was an idea about the kernel exposing the buffer objects of rendered surfaces (Y and UV) as contiguous in memory.  This would avoid conversion on the export side and also make GL import easier with existing hardware.  However, short the kernel managing that, we could modify the UMD to better expose the planes with non-contiguous memory.
> 
> The fundamental requirement was that we be able to import a single buffer into EGL so we can do texture sampling on it, but we also need to be able to reference
> individual buffers in the YUV buffer for the video decoder. 
> 
> Your issue is quite similar.
> 
> What we need are multi-planar functions/access for different bo handles in the sharing of non-contiguous memory by the UMD.
> 
> Sean

fhvwy:
This is an attempt to fix #10 and related problems around inability to describe surfaces in a way usable with external APIs.  Since it isn't fixable with current interfaces, we need to add something new to be able to do it and libva2 seems like a good opportunity for that.  Therefore, this proposes a much more flexible interface for import/export of DRM objects.

This is still looking for general comments, so I haven't actually implemented any of this yet.  I would intend to make something in Mesa and mpv to test it on AMD at least.

The descriptor is constructed with the intent that after export one can import directly to EGL using <https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import.txt> and <https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt>.

Wrt the complexity of the descriptor, we would like to be able to handle at least these three cases for NV12-like surfaces:
* One object, two layers (DRM_FORMAT_R8, DRM_FORMAT_RG88) each containing of one plane .
* Two objects, each a single layer (DRM_FORMAT_R8, DRM_FORMAT_RG88) with one plane.
* One object, one layer of DRM_FORMAT_NV12, containing two planes.

I define new a new pair of Acquire/Release functions to support this - the current VABufferInfo structure is insufficient to support the necessary information, so new functions seemed sensible.  They also act directly on surfaces rather than buffers of derived images - I believe this is the only case which anyone would ever use, though please do say if there is some other use-case which I'm not aware of.

Locking is left unclear here.  I think it would be cleaner if the export functions did not do any synchronisation, and instead request that the user calls vaSyncSurface() as necessary for their use-case (with behaviour undefined if they act on an object with another API during a render operation).  That would support mapping a single time and then synchronising as necessary without the need to keep creating and destroying objects, but there might be some other reason I'm not aware of why this causes problems.

(To compare, other cases of sharing between video and processing APIs tend to have separate mapping and synchronisation - see e.g. <https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/cl_khr_dx9_media_sharing.html> or <https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/cl_khr_d3d11_sharing.html>.)

Some other random questions:
* Would there be any value in adding a read/write distinction to import or export?
* Is any more detail needed for Mesa - for example, to support separated fields?
* How many planes do we need, at most?  (YUVA or RGBA planar formats need four.)
* Do we want to support any other memory types with this new API?  (For example, something similar could work for flink handles rather than PRIME fds if that has any value.  No idea about others - Android?)

All thoughts welcome.  I'm not attached to any particular detail here - feel free to suggest changing everything :)


xhaihao:
@fhvwy @scott-d-phillips has some commits for dma buffer modifier too
https://github.com/scott-d-phillips/intel-vaapi-driver/commit/5aebbb3043cda3626f62698f03d112a987eccc64
https://github.com/scott-d-phillips/libva/commit/52e006fe0126d8535d88d045872307d8e332386f#diff-ce40de7930eb5bb0c9f5801e5101ce29, could you have a look?

xhaihao:
> Locking is left unclear here. I think it would be cleaner if the export functions did not do any synchronisation, and instead request that the user calls vaSyncSurface() as necessary for their use-case (with behaviour undefined if they act on an object with another API during a render operation). That would support mapping a single time and then synchronising as necessary without the need to keep creating and destroying objects, but there might be some other reason I'm not aware of why this causes problems.

Agree with you that the export functions shouldn't do synchronisation

xhaihao:
> This is still looking for general comments, so I haven't actually implemented any of this yet. I would intend to make something in Mesa and mpv to test it on AMD at least.

AFAIK, Scott did some testing in gstreamer, please refer to https://bugzilla.gnome.org/show_bug.cgi?id=779146

fhvwy:
@scott-d-phillips  Some thoughts on that branch:

* Do we really want DRM-specific API calls here?  vaCreateSurfaces() can already hold arbitrary attributes, and an extended export function could easily support more than one type rather than being hardcoded to this one structure.
* For multiple plane formats, I don't think we should be using DRM_FORMAT_NV12 and other composed layouts - they seem to me to be intended for inflexible embedded cases with implicit conversion.  We should use R8/GR88 instead of NV12 (and R16/GR32 instead of P010, etc.) so that they interoperate directly with APIs like EGL and OpenCL.  Not sure on formats like YUYV which interleave both ways - it might be easier to avoid those in the first version (you generally can't represent them in the graphics APIs without hackery or nonstandard extensions anyway).
* As a consequence of that, I think the descriptor structure there is not sufficiently general: format needs to be per-part-of-object (called 'layer' in my version, better name welcome).  Similarly, the format modifier needs to be per-object rather than per-surface.  (This wants to be possible on import, even if the export format is fixed by the driver.)
* Object size is required by some import APIs.
* Freeing an object needs to be clarified - the current implementation seems to require the user to deduplicate the file descriptors before closing them.  (This motivated the separate objects array in what I wrote.)
* dma-buf is a kernel-internal name.  Calling them DRM objects (here referred to via PRIME fds and manipulated with libdrm) feels clearer to me, and fits with the existing naming in libva.

scott-d-phillips:
> For multiple plane formats, I don't think we should be using DRM_FORMAT_NV12 and other composed layouts - they seem to me to be intended for inflexible embedded cases with implicit conversion. We should use R8/GR88 instead of NV12 (and R16/GR32 instead of P010, etc.) so that they interoperate directly with APIs like EGL and OpenCL. Not sure on formats like YUYV which interleave both ways - it might be easier to avoid those in the first version (you generally can't represent them in the graphics APIs without hackery or nonstandard extensions anyway).

On the other hand, if you want to pass the surface to KMS for display, or to some other process for libva encoding then you do want NV12, P010, etc IIUC. Maybe this preference for the form of multi-planar export can be expressed with the API?

Otherwise agree with your points @fhvwy

fhvwy:
Can Y-tiled NV12 or P010 be used for scanout on Intel?  I didn't think they could.  In any case, if you want correct colours (with arbitrary colour spaces and HDR and whatever other features people dream up to mess with output) then you need to do a proper rendering to RGB for display, so I'm not sure how useful it could be (people typically use EGL+OpenGL for this, which is happy to render to KMS buffers).

For encoding, there isn't actually any hardware restriction on the plane locations on Intel is there?  The current driver does restrict it, but only because it has to be in a single object - the offsets within the object can be arbitrary, I think.  So, that should be fixable without too much difficulty.

I guess making it configurable would be fine with me, as long as the single-plane formats are accessible for the use-cases I care about.

scott-d-phillips:
> Can Y-tiled NV12 or P010 be used for scanout on Intel?

Yes, it can. Patches to add that support in the kernel are just now hitting the intel-gfx list.

> For encoding, there isn't actually any hardware restriction on the plane locations on Intel is there?

The hardware requires the address of the UV plane be greater than the address of the Y plane, so not just any two R8 and GR88 planes can be used as NV12.

fhvwy:
Ok, here is a new version, now fully working for export.

Driver support for Intel:
https://github.com/fhvwy/intel-vaapi-driver/commit/78a824cc9b19650d3f057c006e3b0b5d535c8ba8
Driver support for AMD (Mesa):
https://lists.freedesktop.org/archives/mesa-dev/2017-September/169953.html
Use in mpv for playback:
https://github.com/fhvwy/mpv/commit/0a505f9ff5f43ecb40abb99096204d1efbe47346

fhvwy:
* The read/write mode is included - Mesa has this so it seemed sensible to use it.  Intel will currently always return read/write objects.
* I've added a way to indicate that either separate planes (for further rendering) or composite objects (for scanout or similar immediate use) are desirable.  The Intel driver supports both modes (with some restrictions - there is no DRM format for P010 so it isn't usable as composite, while YUYV isn't representable as separate planes), while the Mesa driver supports separate planes only.
* Only export is implemented in either driver.  Import is specified to work via vaCreateSurfaces(), but not yet done.
* Additional memory types are still open.  I only have interest in DRM currently, but it should be usable for anything else.
* I haven't specified anything around making new objects with particular properties (this was suggested in the gstreamer discussion).  I think this use case is better served by making them with gbm or similar and then importing, rather than cluttering libva with a lot more options to control this?  Thoughts welcome on that.

xhaihao:
How about to add width and height for the exporting surface in VADRMPRIMESurfaceDescriptor? The 3rd party library might need these metadata.

fhvwy:
I was thinking they are already known because they are passed to vaCreateSurfaces() (for export by the creator, and you have to know them directly as an importer).  But yeah, that makes sense, I'll add them.

xhaihao:
Could you please change your PR to 'merge 1 commits to 01org:master'? All patches on v2.0-branch have been merged into master.

xhaihao:
https://bugzilla.gnome.org/show_bug.cgi?id=779146 is using @scott-d-phillips's definition which is a little different from yours, could you have a look?

fhvwy:
(Moved to #125.)

PR #86, opened 2017-07-17
add 2D buffer support

xfengcarl:
standard usage: permb  control buffer, it is always a 2D buffer related with MB Postion

Signed-off-by: Carl.Zhang <carl.zhang@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -2476,6 +2476,28 @@ VAStatus vaCreateBuffer (
 );
 
 /**
+ * Create a buffer for given width & height get unit_size, pitch, buf_id for 2D buffer
+ * for permb qp buffer, it will return unit_size for one MB or LCU and the pitch for alignments
+ * can call vaMapBuffer with this Buffer ID to get virtual address.
+ * e.g. AVC 1080P encode, 1920x1088, the size in MB is 120x68,but inside driver,
+ * maybe it should align with 256, and one byte present one Qp.so, call the function.
+ * then get unit_size = 1, pitch = 256. call vaMapBuffer to get the virtual address (pBuf).
+ * then read write the memory like 2D. the size is 256x68, application can only use 120x68
+ * pBuf + 256 is the start of next line.
+ * different driver implementation maybe return different unit_size and pitch
+ */
+VAStatus vaCreateBuffer2(
+    VADisplay dpy,
+    VAContextID context,
+    VABufferType type,
+    unsigned int width,
+    unsigned int height,
+    unsigned int *unit_size,
+    unsigned int *pitch,
+    VABufferID *buf_id
+);
+
+/**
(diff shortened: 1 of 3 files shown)


va/va.h
```diff
@@ -2180,6 +2180,30 @@ VAStatus vaCreateBuffer (
...
+ */
+VAStatus vaCreateBuffer2D (
+    VADisplay dpy,
+    VAContextID context,
+    VABufferType type,
+    unsigned int width,
+    unsigned int height,
+    VABuffer2D *buf2d
+);
+
```
xhaihao: Are all VABufferType types acceptable?

va/va_backend.h
```diff
@@ -140,7 +140,9 @@ struct VADriverVTable
 		unsigned int size,		/* in */
 		unsigned int num_elements,	/* in */
 		void *data,			/* in */
-		VABufferID *buf_id		/* out */
+                unsigned int *unit_size,
+                unsigned int *pitch,
+                VABufferID *buf_id
```
fhvwy: Please don't make this an incompatible driver API change by modifying the signature of the existing function - instead add the new function vaCreateBuffer2() in the driver API just like the external API.

va/va.h
```diff
@@ -2181,6 +2181,24 @@ VAStatus vaCreateBuffer (
...
+ * can call vaMapBuffer with this Buffer ID to get virtual address. 
+ */
+VAStatus vaCreateBuffer2(
+    VADisplay dpy,
+    VAContextID context,
+    VABufferType type,
+    unsigned int width,
+    unsigned int height,
+    unsigned int *unit_size,
+    unsigned int *pitch,
```
fhvwy: I think this needs a clearer description of what it does.  Is it that you supply width and height, and based on those and the buffer type the values pointed to by unit_size and pitch are written with something which you need to know when mapping the buffer?  (An example might help.)

xfengcarl: yes, when you create 2D buffer, you certainly need the pitch for the alignment information, unit_size is another one, for example, we need 4 bytes to represent something of one MB, the unit_size should be 4

va/va_backend.h
xfengcarl: add another function pointer in the v-table , right?

xhaihao: Yes, you should add another function pointer at the end of  the vtable.  your change breaks the compiling of other vaapi backend drivers

va/va.h
xhaihao: Agree with @fhvwy, could you update the comment?

va/va_backend.h
xfengcarl: accept

va/va.h
xfengcarl: accept, will add a example

PR #92, opened 2017-08-11
va: Add the fourcc of I420 format

mypopydev:
i965 will return the I420 in vaQuerySurfaceAttributes, but va.h didn't
add the pre-define fourcc, and will deprecate IYUV because it same as
I420 but most user perfer I420.


xhaihao:
@mypopydev there are some conflicts, could you resolve the conflicts?

PR #105, opened 2017-08-22
Bump a new VA API version

xhaihao:
Some new definitions were added for FEI, so bump a new VA API version


sreerenjb:
Just to confirm: Does the addition of a new entrypoint and whole encode structures still fall into the "micro version change", not the "minor version change" ??

xhaihao:
# - increment micro for any other change (new flag, new codec definition, etc.)

I think FEI is similar to a new codec in libva. When we add a new codec, we also need to add entrypoint and encode structures etc

PR #108, opened 2017-08-31
HEVC FEI support

xfengcarl:
HEVC FEI support, new HEVC FEI buffer in va.h and add va_fei_hevc.h

Signed-off-by: Kelvin Hu <kelvin.hu@intel.com>
Signed-off-by: xfengcarl <carl.zhang@intel.com>

### DIFF
--- a/va/va_fei_hevc.h
+++ b/va/va_fei_hevc.h
@@ -1,0 +1,246 @@
+/*
+ * Copyright (c) 2007-2017 Intel Corporation. All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sub license, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL INTEL AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * \file va_fei_hevc.h
+ * \brief The FEI encoding HEVC special API
+ */
+
+
+#ifndef  __VA_FEI_HEVC_H__
+#define  __VA_FEI_HEVC_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+#include "va_fei.h"
+
+/** \brief FEI frame level control buffer for HEVC */
+typedef struct _VAEncMiscParameterFEIFrameControlHEVC
+{
+    /* one of the VAConfigAttribFEIFunctionType values */
+    uint32_t      function; 
+    /** \brief CTB control input buffer. It is valid only when per_ctb_input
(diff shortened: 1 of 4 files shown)


xfengcarl:
because it include va_fei.h, so it cant pass build. but va_fei.h is not checked in now, not sure whether I should set different compare target.

uartie:
please rebase as v2.0-branch has va_fei.h now.

va/va_fei_hevc.h
```diff
@@ -0,0 +1,282 @@
...
+    // DWORD 4
+    uint32_t    reserved3                        : 16;  /* sse_classID_32x32_0~3 */
+    uint32_t    lcu_force_zero_coeff             : 1;
+    uint32_t    disable_sao_on_a_lcu_flag        : 1;
+    uint32_t    reserved4                        : 14;
+
+    // DWORD 5~7
+    uint32_t    reserved5[3];  /* padding zero DWORDs */
+} VAEncFEICTBCmdLongHevc;   // 32 bytes
+
```
xhaihao: Missed '}' for __cplusplus

va/va_fei_hevc.h
```diff
@@ -0,0 +1,282 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+#include "va_fei.h"
+
+/** \brief FEI frame level control buffer for Hevc */
+typedef struct _VAEncMiscParameterFEIFrameControlHevc
```
xhaihao: could you replace Hevc with HEVC for all data structure definitions in this file?  We use HEVC in other data structures, please keep the same name style.

va/va_fei_hevc.h
```diff
@@ -0,0 +1,282 @@
...
+
+    /** \brief number of MV predictors L0 and L1. the maximum number of motion vector predictor for a 16x16, 32x32 or
+     * 64x64 block is four, it must not be greater than maximum supported MV predictor,
+     **/
+    uint32_t      num_mv_predictors_l0      : 16;
+    uint32_t      num_mv_predictors_l1      : 16;
+
+    /** \brief control parameters */
+    uint32_t      search_path               : 8;
+    uint32_t      len_sp                    : 8;
```
xhaihao: what does 'sp' stand for? could you add more comment for this field or use a meaningful name?

va/va_fei_hevc.h
```diff
@@ -0,0 +1,282 @@
...
+ * it is output buffer of ENC and ENC_PAK modes, it's also input buffer of PAK mode.
+ * The CTB block is in raster scan order. Buffer size shall not be less than the number of CTB blocks
+ * multiplied by sizeof(VAEncFEICTBCmdHevc). Note that, when separate ENC and PAK is enabled,
+ * the exact layout of this buffer is needed for PAK input. App can reuse this buffer,
+ * or copy to a different buffer as PAK input, reserved elements must not be modified when used as PAK input.
+ **/
+typedef struct _VAEncFEICTBCmdHevc
+{
+    // DWORD 0
+    uint32_t    reserved0;
```
xhaihao: Could you move the reserved bytes to the end of this structure?

va/va_fei_hevc.h
```diff
@@ -0,0 +1,282 @@
...
+    uint32_t    split_flag_level1                : 4;
+    uint32_t    split_flag_level0                : 1;
+    uint32_t    reserved1                        : 3;
+    uint32_t    cu_count_minus1                  : 6;
+    uint32_t    last_ctb_of_tile_flag            : 1;
+    uint32_t    last_ctb_of_slice_flag           : 1;
+
+    // DWORD 2
+    uint32_t    current_lcu_Xaddr                : 16;
+    uint32_t    current_lcu_Yaddr                : 16;
```
xhaihao: could you use lowercase for each field?

va/va_fei_hevc.h
kelvinhu325: SP means Search Path, len_sp means Length of Search Path.

va/va_fei_hevc.h
kelvinhu325: No, this struct is HW PAK engine input layout, we can not move reserved0, otherwise HW PAK may gets hang.

xhaihao: so this buffer is hardware dependent, right?

kelvinhu325: yes, actually the first DWORD probable includes some debug info , which does not want customers know, it does not mean only a reserved DWORD.

kelvinhu325: Yes

sreerenjb:
Many things are unclear in this PR
-- it seem the va.h is missing the buffer type definitions of HEVC FEI.
-- I didn't see the definition of VAEncFEICURecordLongHevc
-- no defintion of VAEncFEICURecordHevc
-- no definitin of VAEncFEIDistortionBufferHevc
-- if we are going to put comment like this " Each CTB block has one CU record data with layout VAEncFEICURecordLongHevc for SKL and VAEncFEICURecordHevc for ICL+" then it is platform specific apis, we should move them under vendor/intel or something similar.

PR #110, opened 2017-09-04
Statistics function for FEI and non-FEI

kelvinhu325:
A pre-processing function for getting some statistics and motion vectors
is added, and some extra controls for Encode pipeline are provided. The
application can optionally call the statistics function to get motion
vectors and statistics like variances, distortions before calling Encode
function via this entry point.

Signed-off-by: Kelvin Hu <kelvin.hu@intel.com>
Signed-off-by: Jonathan Bian <jonathan.bian@intel.com>
Signed-off-by: Sreerenj Balachandran <sreerenj.balachandran@intel.com>
Signed-off-by: Zhong Li <zhong.li@intel.com>
Signed-off-by: Daniel Charles <daniel.charles@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -401,6 +401,23 @@ typedef enum
      * (VAEncFEIMVBufferType, VAEncFEIMBModeBufferType) are needed for PAK input.
      **/
     VAEntrypointFEI         = 11,
+    /**
+     * \brief VAEntrypointStats
+     *
+     * A pre-processing function for getting some statistics and motion vectors is added,
+     * and some extra controls for Encode pipeline are provided. The application can
+     * optionally call the statistics function to get motion vectors and statistics like
+     * variances, distortions before calling Encode function via this entry point.
+     *
+     * Checking whether Statistics is supported can be performed with vaQueryConfigEntrypoints().
+     * If Statistics entry point is supported, then the list of returned entry-points will
+     * include #VAEntrypointStats. Supported pixel format, maximum resolution and statistics
+     * specific attributes can be obtained via normal attribute query. One input buffer
+     * (VAStatsStatisticsParameterBufferType) and one or two output buffers
+     * (VAStatsStatisticsBufferType, VAStatsStatisticsBottomFieldBufferType (for interlace only)
+     * and VAStatsMVBufferType) are needed for this entry point.
+     **/
+    VAEntrypointStats       = 12,
 } VAEntrypoint;
 
 /** Currently defined configuration attribute types */
@@ -678,6 +695,14 @@ typedef enum
      * Currently it is for FEI entry point only.
      */
     VAConfigAttribFEIMVPredictors     = 33,
+    /**
+     * \brief Statistics attribute. Read-only.
+     *
+     * This attribute exposes a number of capabilities of the VAEntrypointStats entry
+     * point. The attribute value is partitioned into fields as defined in the
+     * VAConfigAttribValStats union. Currently it is for VAEntrypointStats only.
+     */
+    VAConfigAttribStats               = 34,
      /**
      * \brief Tile Support Attribute. Read-only.
      *
@@ -1550,6 +1575,13 @@ typedef enum
     VAEncFEIDistortionBufferType        = 45,
     VAEncFEIMBControlBufferType         = 46,
     VAEncFEIMVPredictorBufferType       = 47,
+    VAStatsStatisticsParameterBufferType = 48,
+    /** \brief Statistics output for VAEntrypointStats progressive and top field of interlaced case*/
+    VAStatsStatisticsBufferType         = 49,
+    /** \brief Statistics output for VAEntrypointStats bottom field of interlaced case*/
+    VAStatsStatisticsBottomFieldBufferType = 50,
+    VAStatsMVBufferType                 = 51,
+    VAStatsMVPredictorBufferType        = 52,
     /** Force MB's to be non skip for encode.it's per-mb control buffer, The width of the MB map
      * Surface is (width of the Picture in MB unit) * 1 byte, multiple of 64 bytes.
      * The height is (height of the picture in MB unit). The picture is either
--- a/va/va_fei.h
+++ b/va/va_fei.h
@@ -61,6 +61,102 @@ extern "C" {
 
 /**@}*/
 
+/** \brief Attribute value for VAConfigAttribStats */
+typedef union _VAConfigAttribValStats {
+    struct {
+        /** \brief Max number of past reference frames that are supported. */
+        uint32_t    max_num_past_references   : 4;
+        /** \brief Max number of future reference frames that are supported. */
+        uint32_t    max_num_future_references : 4;
+        /** \brief Number of supported output buffers for VAStatsStatisticsParameter->outputs */
+        uint32_t    num_outputs               : 3;
+        /** \brief Interlaced content is supported */
+        uint32_t    interlaced                : 1;
+        uint32_t    reserved                  : 20;
+    } bits;
+    uint32_t value;
+} VAConfigAttribValStats;
+
+typedef struct _VAPictureStats
+{
+    VASurfaceID picture_id;
+    /*
+     * see flags below.
+     */
+    uint32_t flags;
+} VAPictureStats;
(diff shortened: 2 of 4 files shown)


va/va.h
```diff
@@ -381,6 +381,25 @@ typedef enum
...
+     * \brief VAEntrypointStats
+     *
+     * Statistics for FEI and non-FEI. A pre-processing function for getting some
+     * statistics and motion vectors is added, and some extra controls for Encode
+     * pipeline are provided. The application can optionally call the statistics
+     * function to get motion vectors and statistics like variances, distortions before
+     * calling Encode function via this entry point.
+     *
+     * Checking whether Statistics is supported can be performed with vaQueryConfigEntrypoints()
+     * and the profile argument set to AVC profiles. If Statistics entry point is supported,
```
xhaihao: Does this function support AVC only?

va/va_fei_h264.h
```diff
@@ -318,6 +318,115 @@ typedef struct _VAEncFEIDistortionH264 {
...
+    uint32_t    adaptive_search	            : 1;
+    /** \brief indicate if future or/and past MV in mv_predictor buffer is valid.
+     * 0: MV predictor disabled
+     * 1: MV predictor enabled for past reference
+     * 2: MV predictor enabled for future reference
+     * 3: MV predictor enabled for both past and future references
+     **/
+    uint32_t    mv_predictor_ctrl           : 3;
+    uint32_t    mb_qp                       : 1;
+    uint32_t    ft_enable                   : 1;
```
xhaihao: What does ft stand for? Could you use a meaningful name?

va/va_fei_h264.h
```diff
@@ -318,6 +318,115 @@ typedef struct _VAEncFEIDistortionH264 {
...
+/** \brief Motion Vector and Statistics frame level controls.
+ * VAStatsStatisticsParameterBufferType for H264 16x16 block
+ **/
+typedef struct _VAStatsStatisticsParameterH264
+{
+   VAStatsStatisticsParameter stats_params;
+
+    uint32_t    frame_qp                    : 8;
+    /** \brief maximum number of Search Units, valid range is [1, 63] */
+    uint32_t    len_sp                      : 8;
```
xhaihao: How large is the search unit? what does sp stand for?

va/va.h
kelvinhu325: Currently yes, we only support VAEntrypointStats of AVC, and HEVC FEI will reuse VAEntrypointStats of AVC for saving big effort on VAEntrypointStats of HEVC.
For long term, we will add support to VAEntrypointStats of HEVC.

va/va.h
```diff
@@ -381,6 +381,25 @@ typedef enum
      * (VAEncFEIMVBufferType, VAEncFEIMBModeBufferType) are needed for PAK input.
      **/
     VAEntrypointFEI         = 11,
+    /**
+     * \brief VAEntrypointStats
+     *
+     * Statistics for FEI and non-FEI. A pre-processing function for getting some
```
sreerenjb: What does it mean my "Statistics for FEI and non-FEI" ? Ideally, VAEntrypointStats is a generic entrypoint for all codecs, If so why should we mention FEI, non-FEI terms here?

va/va.h
sreerenjb: Since "VAEntrypointStats" is a generic entrypoint, we shouldn't put the comment like "profile argument set to AVC profile".
Profile argument can be any profile and if the driver doesn't have support for that profile,  simply don't return the Stats entrypoint in vaQueryConfigEntrypoint().  Am I missing something?

va/va_fei.h
```diff
@@ -61,6 +61,102 @@ extern "C" {
...
+} VAConfigAttribValStats;
+
+typedef struct _VAPictureFEI
+{
+    VASurfaceID picture_id;
+    /*
+     * see flags below.
+     */
+    uint32_t flags;
+} VAPictureFEI;
```
sreerenjb: Just curious why is it named with "FEI" substring? why not "VAPictureStat"?

va/va_fei.h
```diff
@@ -61,6 +61,102 @@ extern "C" {
...
+#define VA_PICTURE_FEI_INVALID                   0x00000001
+#define VA_PICTURE_FEI_PROGRESSIVE               0x00000000
+#define VA_PICTURE_FEI_TOP_FIELD                 0x00000002
+#define VA_PICTURE_FEI_BOTTOM_FIELD              0x00000004
+/** \brief picutre surface content updated indicator.
+ * The picture surface content is updated, it means temporary buffer like downscaled pixel data in driver
+ * internal needs be forced freshing
+ **/
+#define VA_PICTURE_FEI_CONTENT_UPDATED           0x00000010
+
```
sreerenjb: Similar to the previous comment, why  VA_PICTURE_FEI_*  came as preferred name in favor of VA_PICTURE_STATS_*  ?

va/va_fei.h
```diff
@@ -61,6 +61,102 @@ extern "C" {
...
 
+/** \brief Attribute value for VAConfigAttribStats */
+typedef union _VAConfigAttribValStats {
+    struct {
+        /** \brief Max number of past reference frames that are supported. */
+        uint32_t    max_num_past_references   : 4;
+        /** \brief Max number of future reference frames that are supported. */
+        uint32_t    max_num_future_references : 4;
+        /** \brief Number of output surfaces that are supported */
+        uint32_t    num_outputs               : 3;
```
sreerenjb: Can you explain what exactly is this "num_outputs"? I don't understand the comment "Number of output surfaces that are supported". Is it for "the number of output buffers(stat and mv)"?

va/va_fei_h264.h
```diff
@@ -318,6 +318,115 @@ typedef struct _VAEncFEIDistortionH264 {
...
+     * 1: full search
+     * 2: diamond search
+     **/
+    uint32_t    search_path                 : 8;
+    uint32_t    reserved0                   : 8;
+
+    uint32_t    sub_mb_part_mask            : 7;
+    uint32_t    sub_pel_mode                : 2;
+    uint32_t    inter_sad                   : 2;
+    uint32_t    intra_sad                   : 2;
```
sreerenjb: I think we should add an explanation for all these fields, like "01 indicate half and 11 indicate quarter pixel interpolation" right?

va/va.h
xhaihao: Agree with Sree, the API should be generic, we shouldn't add such requirement in API definition to workaround driver limitation.

va/va.h
kelvinhu325: it only means VAEntrypointStats does not limit to only support FEI use case,  do you think it better to not mention FEI here?

va/va.h
kelvinhu325: At beginning, we use NonProfile for we want it no profile limitation, but you suggest it to be AVC profile since actually driver internal works as AVC Profile, so we change it, driver always requires a profile input, I still think it better to be Non profile than any profiles here, how do you think about it?

va/va_fei.h
kelvinhu325: yes, it should be number of output buffers for stats and mv

va/va_fei.h
kelvinhu325: Yes, VAPictureStats may be more accurate since it's only used for VAEntrypointStats

va/va_fei.h
kelvinhu325: Yes, VA_PICTURE_STATS_ may be more accurate since it's only used for VAEntrypointStats

va/va_fei_h264.h
kelvinhu325: The SP length range is [1, 63] as the comment, SP means search path

va/va_fei_h264.h
kelvinhu325: Yes, we should have comments for it like below,
00b: integer mode searching
01b: half-pel mode searching
10b: reserved
11b: quarter-pel mode searching

va/va_fei_h264.h
kelvinhu325: ft_enable means forward transform enable.

PR #112, opened 2017-09-06
New VAAPI definition for multi-frame processing

artem-shaporenko:
New VAAPI definition for multi-frame processing applicable for different entry points.

Signed-off-by: Artem Shaporenko artem.shaporenko@intel.com

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -1151,6 +1151,96 @@ VAStatus vaDestroyContext (
   return vaStatus;
 }
 
+VAStatus vaCreateMFContext (
+    VADisplay dpy,
+    VAMFContextID *mf_context    /* out */
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+    
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+    if(ctx->vtable->vaCreateMFContext == NULL)
+        vaStatus = VA_STATUS_ERROR_UNIMPLEMENTED;
+    else
+    {
+        vaStatus = ctx->vtable->vaCreateMFContext( ctx, mf_context);
+        VA_TRACE_ALL(va_TraceCreateMFContext, dpy, mf_context);
+    }
+
+    return vaStatus;
+}
+
+VAStatus vaMFAddContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+    
+    if(ctx->vtable->vaMFAddContext == NULL)
+        vaStatus = VA_STATUS_ERROR_UNIMPLEMENTED;
+    else
+    {
+        vaStatus = ctx->vtable->vaMFAddContext( ctx, context, mf_context);
+        VA_TRACE_ALL(va_TraceMFAddContext, dpy, context, mf_context);
+    }
+
+    return vaStatus;
+}
+
+VAStatus vaMFReleaseContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+    if(ctx->vtable->vaMFReleaseContext == NULL)
+        vaStatus = VA_STATUS_ERROR_UNIMPLEMENTED;
+    else
+    {
+        vaStatus = ctx->vtable->vaMFReleaseContext( ctx, context, mf_context);
+        VA_TRACE_ALL(va_TraceMFReleaseContext, dpy, context, mf_context);
+    }
+
+    return vaStatus;
+}
+
+VAStatus vaMFSubmit (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID *contexts,
+    int num_contexts
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+    CHECK_VTABLE(vaStatus, ctx, MFSubmit);
+    if(ctx->vtable->vaMFSubmit == NULL)
+        vaStatus = VA_STATUS_ERROR_UNIMPLEMENTED;
+    else
+    {
+        vaStatus = ctx->vtable->vaMFSubmit( ctx, mf_context, contexts, num_contexts);
+        VA_TRACE_ALL(va_TraceMFSubmit, dpy, mf_context, contexts, num_contexts);
+    }
+
+    return vaStatus;
+}
+
 VAStatus vaCreateBuffer (
     VADisplay dpy,
     VAContextID context,	/* in */
--- a/va/va.h
+++ b/va/va.h
@@ -1177,6 +1177,86 @@ VAStatus vaDestroyContext (
     VAContextID context
 );
 
+//Multi-frame context
+typedef VAGenericID VAMFContextID;
+/**
+ * vaCreateMFContext - Create a multi-frame context
+ *  interface encapsulating common for all streams memory objects and structures
+ *  required for single GPU task submission from several VAContextID's.
+ *  Allocation: This call only creates an instance, doesn't allocate any additional memory.
+ *  Support identification: Application can identify multi-frame feature support by ability
+ *  to create multi-frame context. If driver supports multi-frame - call successful,
+ *  mf_context != NULL and VAStatus = VA_STATUS_SUCCESS, otherwise if multi-frame processing
+ *  not supported driver returns VA_STATUS_ERROR_UNIMPLEMENTED and mf_context = NULL.
+ *  return values:
+ *  VA_STATUS_SUCCESS - operation successful.
+ *  VA_STATUS_ERROR_UNIMPLEMENTED - no support for multi-frame.
+ *  dpy: display adapter.
+ *  mf_context: Multi-Frame context encapsulating all associated context
+ *  for multi-frame submission.
+ */
+VAStatus vaCreateMFContext (
+    VADisplay dpy,
+    VAMFContextID *mf_context    /* out */
+);
+
+/**
+ * vaMFAddContext - Provide ability to associate each context used for
+ *  Multi-Frame submission and common Multi-Frame context.
+ *  Try to add context to understand if it is supported.
+ *  Allocation: this call allocates and/or reallocates all memory objects
+ *  common for all contexts associated with particular Multi-Frame context.
+ *  All memory required for each context(pixel buffers, internal driver
+ *  buffers required for processing) allocated during standard vaCreateContext call for each context.
+ *  Runtime dependency - if current implementation doesn't allow to run different entry points/profile,
+ *  first context added will set entry point/profile for whole Multi-Frame context,
+ *  all other entry points and profiles can be rejected to be added.
+ *  Return values:
+ *  VA_STATUS_SUCCESS - operation successful, context was added.
+ *  VA_STATUS_ERROR_OPERATION_FAILED - something unexpected happened - application have to close
+ *  current mf_context and associated contexts and start working with new ones.
+ *  VA_STATUS_ERROR_INVALID_CONTEXT - ContextID is invalid, means:
+ *  1 - mf_context is not valid context or
+ *  2 - driver can't suport different VAEntrypoint or VAProfile simultaneosly
+ *  and current context contradicts with previously added, application can continue with current mf_context
+ *  and other contexts passed this call, rejected context can continue work in stand-alone
+ *  mode or other mf_context.
+ *  VA_STATUS_ERROR_UNSUPPORTED_ENTRYPOINT - particular context being added was created with with
+ *  unsupported VAEntrypoint. Application can continue with current mf_context
+ *  and other contexts passed this call, rejected context can continue work in stand-alone
+ *  mode.
+ *  VA_STATUS_ERROR_UNSUPPORTED_PROFILE - Current context with Particular VAEntrypoint is supported
+ *  but VAProfile is not supported. Application can continue with current mf_context
+ *  and other contexts passed this call, rejected context can continue work in stand-alone
+ *  mode.
+ *  dpy: display adapter.
+ *  context: context being associated with Multi-Frame context.
+ *  mf_context: - multi-frame context used to associate contexts for multi-frame submission.
+ */
+VAStatus vaMFAddContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+);
+
+/**
+ * vaMFReleaseContext - Removes context from multi-frame and
+ *  association with multi-frame context.
+ *  After association removed vaEndPicture will submit tasks, but not vaMFSubmit.
+ *  Return values:
+ *  VA_STATUS_SUCCESS - operation successful, context was removed.
+ *  VA_STATUS_ERROR_OPERATION_FAILED - something unexpected happened.
+ *  application need to destroy this VAMFContextID and all assotiated VAContextID
+ *  dpy: display
+ *  mf_context: VAMFContextID where context is added
+ *  context: VAContextID to be added
+ */
+VAStatus vaMFReleaseContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+);
+
 /**
  * Buffers 
  * Buffers are used to pass various types of data from the
@@ -2654,10 +2734,38 @@ VAStatus vaRenderPicture (
  * The server should start processing all pending operations for this 
  * surface. This call is non-blocking. The client can start another 
  * Begin/Render/End sequence on a different render target.
+ * if VAContextID used in this function previously successfully passed
+ * vaMFAddContext call, real processing will be started during vaMFSubmit
  */
 VAStatus vaEndPicture (
     VADisplay dpy,
     VAContextID context
+);
+
+/**
+ * Make the end of rendering for a pictures in contexts passed with submission.
+ * The server should start processing all pending operations for contexts.
+ * All contexts passed should be associated through vaMFAddContext
+ * and call sequence Begin/Render/End performed.
+ * This call is non-blocking. The client can start another
+ * Begin/Render/End/vaMFSubmit sequence on a different render targets.
+ * Return values:
+ * VA_STATUS_SUCCESS - operation successful, context was removed.
+ * VA_STATUS_ERROR_INVALID_CONTEXT - mf_context or one of contexts are invalid
+ * due to mf_context not created or one of contexts not assotiated with mf_context
+ * through vaAddContext.
+ * VA_STATUS_ERROR_INVALID_PARAMETER - one of context has not submitted it's frame
+ * through vaBeginPicture vaRenderPicture vaEndPicture call sequence.
+ * dpy: display
+ * mf_context: Multi-Frame context
+ * contexts: list of contexts submitting their tasks for multi-frame operation.
+ * num_contexts: number of passed contexts.
+ */
+VAStatus vaMFSubmit (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID * contexts,
+    int num_contexts
 );
 
 /*
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -436,8 +436,31 @@ struct VADriverVTable
             VABufferID          buf_id          /* in */
         );
 
+        VAStatus (*vaCreateMFContext) (
+            VADriverContextP ctx,
+            VAMFContextID *mfe_context    /* out */
+        );
+
+        VAStatus (*vaMFAddContext) (
+            VADriverContextP ctx,
+            VAMFContextID mf_context,
+            VAContextID context
+        );
+
+        VAStatus (*vaMFReleaseContext) (
+            VADriverContextP ctx,
+            VAMFContextID mf_context,
+            VAContextID context
+        );
+
+        VAStatus (*vaMFSubmit) (
+            VADriverContextP ctx,
+            VAMFContextID mf_context,
+            VAContextID *contexts,
+            int num_contexts
+        );
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[64];
+        unsigned long reserved[60];
 };
 
 struct VADriverContext
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -1438,6 +1438,65 @@ void va_TraceDestroyContext (
     UNLOCK_CONTEXT(pva_trace);
 }
 
+void va_TraceCreateMFContext (
+    VADisplay dpy,
+    VAMFContextID *mf_context    /* out */
+)
+{
+    DPY2TRACECTX(dpy, VA_INVALID_ID, VA_INVALID_ID);
+    TRACE_FUNCNAME(idx);
+    if (mf_context) {
+        va_TraceMsg(trace_ctx, "\tmf_context = 0x%08x\n", *mf_context);
+        trace_ctx->trace_context = *mf_context;
+    } else
+        trace_ctx->trace_context = VA_INVALID_ID;
+}
+
+void va_TraceMFAddContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+)
+{
+    DPY2TRACECTX(dpy, mf_context, VA_INVALID_ID);
+
+    TRACE_FUNCNAME(idx);
+    va_TraceMsg(trace_ctx, "\tmf_context = 0x%08x\n", mf_context);
+    va_TraceMsg(trace_ctx, "\tcontext = 0x%08x\n", context);
+}
+
+void va_TraceMFReleaseContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+)
+{
+    DPY2TRACECTX(dpy, mf_context, VA_INVALID_ID);
+
+    TRACE_FUNCNAME(idx);
+    va_TraceMsg(trace_ctx, "\tmf_context = 0x%08x\n", mf_context);
+    va_TraceMsg(trace_ctx, "\tcontext = 0x%08x\n", context);
+}
+
+void va_TraceMFSubmit (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID *contexts,
+    int num_contexts
+)
+{
+    int i;
+
+    DPY2TRACECTX(dpy, mf_context, VA_INVALID_ID);
+
+    TRACE_FUNCNAME(idx);
+    va_TraceMsg(trace_ctx, "\tmf_context = 0x%08x\n", mf_context);
+
+    for(i = 0; i < num_contexts; i++){
+        va_TraceMsg(trace_ctx, "\t\tcontext[%d] = 0x%08x\n", i, contexts[i]);
+    }
+}
+
 void va_TraceCreateBuffer (
     VADisplay dpy,
     VAContextID context,	/* in */
--- a/va/va_trace.h
+++ b/va/va_trace.h
@@ -121,6 +121,34 @@ void va_TraceDestroyContext (
 );
 
 DLL_HIDDEN
+void va_TraceCreateMFContext (
+    VADisplay dpy,
+    VAContextID *mf_context	/* out */
+);
+
+DLL_HIDDEN
+void va_TraceMFAddContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+);
+
+DLL_HIDDEN
+void va_TraceMFReleaseContext (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID context
+);
+
+DLL_HIDDEN
+void va_TraceMFSubmit (
+    VADisplay dpy,
+    VAMFContextID mf_context,
+    VAContextID *contexts,
+    int num_contexts
+);
+
+DLL_HIDDEN
 void va_TraceCreateBuffer (
     VADisplay dpy,
     VAContextID context,	/* in */


artem-shaporenko:
Multi-frame processing is an optimization, that allows to run several tasks inside one batch buffer to better reuse EU and fixed function such as media sampler, as well as decrease CPU overhead for batch buffer submission.

# VAStatus vaCreateMFContext (VADisplay dpy, VAMFContextID *mf_context);

## Description:
Entry point to create new Multi-Frame context interface encapsulating common for all streams memory objects and structures required for single batch buffer submission from several VAContextID's.
Allocation: This call only creates an instance, doesn't allocate any additional memory.
Support identification: Application can identify multi-frame feature support by ability to create multi-frame context. If driver supports multi-frame - call successful, mf_context != NULL and VAStatus = VA_STATUS_SUCCESS, otherwise if multi-frame processing not supported driver returns VA_STATUS_ERROR_UNIMPLEMENTED and mf_context = NULL.

## Arguments:
VADisplay dpy - display adapter.
VAMFContextID *out_context - MF context encapsulating all associated encoder context for multi-frame submission.

## Return value:
VA_STATUS_SUCCESS - operation successful.
VA_STATUS_ERROR_UNIMPLEMENTED - no support for multi-frame.

# VAStatus vaMFAddContext (VADisplay dpy, VAContextID context, VAMFContextID mfe_context);

## Description:
Provide ability to associate each encoder context used for submission and common MF context.
Query - try to add context to understand if it is supported.
Allocation: this call allocates and/or reallocates all memory objects common for all streams associated with particular MF context.
All memory required for each encoder context(pixel buffers, per stream buffers such as PAKObj, motion vector buffers, predictor buffers, etc.) allocated during standard vaCreateContext call for each stream.
Runtime dependency - if current implementation doesn't allow to run different entry points/profile, first context added will set entry point/profile for whole MF Context, all other entry points and profiles will be rejected to be added.

## Arguments:
VADisplay dpy - display adapter.
VAContextID context - context being associated with Multi-Frame context.
VAMFContextID mf_context - multi-frame context used to associate other contexts and for multi-frame submission.

## Return value:
VA_STATUS_SUCCESS - operation successful, context was added.
VA_STATUS_ERROR_OPERATION_FAILED - something unexpected happened.
VA_STATUS_ERROR_INVALID_CONTEXT - ContextID is invalid - entry point and/or profile contradicts with previously added.
VA_STATUS_ERROR_UNSUPPORTED_ENTRYPOINT - particular VAContextID created with VAConfigID associated with VAEntrypoint that is not supported, as an example VAEntrypointVLD not supported(as decoder doesn't make sense for MF operation), VAEntrypointVideoProc and/or VAEntrypointFEI not supported at a moment, etc.
VA_STATUS_ERROR_UNSUPPORTED_PROFILE - Current context with Particular VAEntrypoint is supported, but VAProfile is not supported(so for example H264 encode or FEI is supported, but H265/Mpeg2 not supported at a moment)

## Limitations:
Number of encoder contexts per MF context - should not be limited.
Base on implementation - can limit: different VAEntrypoint/VAProfile(codec) inside one context


# VAStatus vaMFReleaseContext (VADisplay dpy, VAContextID context, VAMFContextID mf_context);
## Description:
Provides ability to remove association between MF context and particular encode context. This is not necessary, can be removed, as the same can be done through vaDestroyContext, however such support brings hidden from App, unclear logic in driver.

## Arguments:
VADisplay dpy - display adapter.
VAContextID context - encoder context being removed from MFE context.
VAMFEContextID mfe_context - MFE context used to associate encode context and for submission.

## Return value:
VA_STATUS_SUCCESS - operation successful, context was removed.
VA_STATUS_ERROR_OPERATION_FAILED - something unexpected happened.


# VAStatus vaMFSubmit (VADisplay dpy, VAMFContextID mf_context, VAContextID *contexts, int num_contexts);

## Description:
Provides ability to submit frames from multiple contexts streams for execution.

## Arguments:
VADisplay dpy - display adapter.
VAMFContextID mf_context - MF context used for submission.
VAContextID *contexts - contexts with frames ready for submission.
int num_contexts - number of context in contexts.

## Return value:
VA_STATUS_SUCCESS - operation successful, context was removed.
VA_STATUS_ERROR_INVALID_CONTEXT - mf_context or one of contexts are invalid due to mf_context not created or one of contexts not assotiated with mf_context through vaAddContext.
VA_STATUS_ERROR_INVALID_PARAMETER - one of context has not submitted it's frame through vaBeginPicture vaRenderPicture vaEndPicture call sequence.

# Schematic code flow examples:
## Initialization:

VADisplay dpy;
VAMFContextID mfe_context;
VAContextID contexts[N];
....
//initialize display, etc.
...
vaCreateMFContext(dpy, &mf_context);
.....
for N streams
{
    vaCreateConfig(dpy,...., &stream_config)            
    vaCreateContext(dpy, stream_config, ...., &context[i])
    vaAddContext(dpy, context[i], mf_context)    // Add this context into the MF context
}

## Runtime:

VAContextID fContexts[M]; //frame contexts - to submit, M - max number of frames to submit, number of overall contexts associated with MF can be bigger than maximum number of frames to submit
....
// prepare streams ready to submit frames, fill contexts vector.
....
for M frames        // take frames ready for submission from differen streams, this can happen not 
{
    vaBeginPicture(dpy, fContexts[i], in_surf)
    vaRenderPicture(dpy, fContexts[i], buffers, buffer_count)
    vaEndPicture(dpy, fContexts[i])             // Delay real submission untill vaMFSubmit()
}
vaMFSubmit(dpy, mf_context, contexts, M)  // Submit through MF context

## Destruction:

for each stream 
{
    vaReleaseContext(dpy, context, mf_context);    // remove context 
    vaDestroyContext(dpy, context);
}
vaDestroyContext(mf_context);

xhaihao:
1. Could you please consider a more generic API? e.g. maybe a driverX  can support  Multi-frame processing for decode, 

2. Could you update the descriptor for vaMFAddContext/vaMFReleaseContext/vaMFSubmit etc? Currently these functions are limited for encoder context only. e.g.

> Provide ability to associate each encoder context used for submission and common MF context
> AContextID context - encoder context being removed from MFE context.

3. Can two VA contexts which have different profile/entrypoint pairs be added into the same multi-frame context? e.g. one context for VPP and one context for HEVC encoding. I think it is reasonable to require both MF context and VA context work have the same profile/entrypoint pair. How about to add parameters to pass profile/entrypoint to the driver when calling vaCreateMFContext(), or add a parameter to pass VA config ID to the driver? In this way, the driver can return the right value upon the input profile/entrypoint pair or VA config when calling vaCreateMFContext() and vaMFAddContext() etc.

4. Batch buffer, EU etc should be platform dependent, We should add a generic API in libva, could you update the comment with some common words?

va/va.c
```diff
@@ -1121,6 +1125,78 @@ VAStatus vaDestroyContext (
...
+    VAMFContextID *mf_context    /* out */
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+
+    vaStatus = ctx->vtable->vaCreateMFContext( ctx, mf_context);
```
xhaihao: Please check ctx->vtable->vaCreateMFContext against NULL first.

va/va.c
```diff
@@ -1121,6 +1125,78 @@ VAStatus vaDestroyContext (
...
+    VAMFContextID mf_context
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+
+    vaStatus = ctx->vtable->vaMFAddContext( ctx, context, mf_context);
```
xhaihao: Please check ctx->vtable->vaMFAddContext against NULL first and the similar check for other new APIs below

va/va.h
```diff
@@ -1130,6 +1130,50 @@ VAStatus vaDestroyContext (
     VAContextID context
 );
 
+//Multi-frame context
+typedef VAGenericID VAMFContextID;
+/**
+ * vaCreateMFContext - Create a multi-frame context
+ *  Multi-frame context allow to run tasks from different
+ *  contexts in single batch buffer for performance optimization.
```
xhaihao: the underlying HW might not support batch buffer, could you use some common words?

va/va.h
```diff
@@ -1130,6 +1130,50 @@ VAStatus vaDestroyContext (
...
+ *  After association is done, current context will not submit 
+ *  render targets during vaEndPicture, but during vaMFSubmit.
+ *  dpy: display
+ *  context: VAContextID to be added
+ *  mf_context: VAMFContextID where context is added
+ */
+VAStatus vaMFAddContext (
+    VADisplay dpy,
+    VAContextID context,
+    VAMFContextID mfe_context
```
xhaihao: please replace mfe_context with mf_context in this commit, we should use the same name style.

xhaihao:
There are some trailing whitespace, could you please delete those trailing whitespaces when updating your patch?

> .git/rebase-apply/patch:136: trailing whitespace.
>  *  After association is done, current context will not submit
> .git/rebase-apply/patch:170: trailing whitespace.
>  * Make the end of rendering for a pictures in contexts passed with submission.
> .git/rebase-apply/patch:174: trailing whitespace.
>  * This call is non-blocking. The client can start another
> .git/rebase-apply/patch:178: trailing whitespace.
>  * contexts: list of contexts submitting their tasks for multi-frame operation.
> warning: 4 lines add whitespace errors.

va/va.c
artem-shaporenko: FIxed

va/va.c
artem-shaporenko: FIxed

va/va.h
artem-shaporenko: FIxed

va/va.h
artem-shaporenko: FIxed

artem-shaporenko:
Thanks Haihao, I fixed issues. Also updated descriptions to remove uncertainty for different feature support.
About API itself:
1. It can support decoder or anything(any entry point) created through vaCreateContext depending on underling driver/hw implementation.
2. Fixed
3. By design - yes any entrypoint/profile can be combined together, but this depend on driver implementation, so one implementation can require same entrypoint/profile and reject adding contexts with different ones, another implementation can support different profiles and/or entry point. So it doesn't make sense to add entrypoint/profile to vaCreateMFContext
4. Fixed

xhaihao:
Just for curious, do you have a case that adds a decoder context and a encoder context  in the same Multi-Frame context in practice?

xhaihao:
How does application do when vaMFAddContext returns an error, 
1) Continue to add the remaining contexts to MF context
or 
2) Release the MF context at once, 
or
3) Don't add the remaining context and submit the current MFcontext?

Which one is right ?

artem-shaporenko:
Added notes to vaAddContext error code description - can't continue work if VA_STATUS_ERROR_OPERATION_FAILED, can continue in other cases.

artem-shaporenko:
>>Just for curious, do you have a case that adds a decoder context and a encoder context in the same Multi-Frame context in practice?
Not now

xhaihao:
@sreerenjb @lizhong1008 @xuguangxin Could you help to review the patches? How do you think if adding multi-frame processing in gstreamer-vaapi, FFmpeg and libyami?

xhaihao:
@artem-shaporenko Could you update the comment in va.h too when you update the description on github?

artem-shaporenko:
@xhaihao Do you think it'll be useful to put full description into comments in va.h?

artem-shaporenko:
"gstreamer-vaapi, FFmpeg and libyami"
ffmpeg will add support for sure(at least MSDK path) - we are working on it.

xhaihao:
@artem-shaporenko  We can use doxygen to generate the document for user in libva, so you should document  how to use multi frame processing clearly in va.h.

va/va.h
```diff
@@ -1130,6 +1130,86 @@ VAStatus vaDestroyContext (
...
+ *  and other contexts passed this call, rejected context can continue work in stand-alone
+ *  mode or other mf_context.
+ *  VA_STATUS_ERROR_UNSUPPORTED_ENTRYPOINT - particular VAContextID created with VAConfigID
+ *  associated with VAEntrypoint that is not supported, as an example VAEntrypointVLD not supported
+ *  (as decoder doesn't make sense for Multi-Frame operation), VAEntrypointVideoProc and/or
+ *  VAEntrypointFEI not supported at a moment, etc. Application can continue with current mf_context
+ *  and other contexts passed this call, rejected context can continue work in stand-alone mode.
+ *  VA_STATUS_ERROR_UNSUPPORTED_PROFILE - Current context with Particular VAEntrypoint is supported
+ *  but VAProfile is not supported(so for example H264 encode or FEI is supported, but H265/Mpeg2 not
+ *  supported at a moment). Application can continue with current mf_context and other contexts passed
```
lizhong1008: “H264 encode or FEI is supported”，is this a typo? should be "H264 encoder of FEI is supported" or "H264 encode without FEI is supported"?

lizhong1008: I don't think it is a good idea to list  which profile/entrypoint is not supported or not support in the API file. It means you must update  this comments when driver behavior has been changed.  
Why don't define a vaQueryXXX interface (just like vaQueryConfigEntrypoints) to access drive limitation and capacity? Before calling vaCreateMFContext and vaMFAddContext, we can query firstly, then we can know what feature is supported and avoid many inexplicable error.

va/va.h
```diff
@@ -2502,6 +2582,32 @@ VAStatus vaEndPicture (
     VAContextID context
 );
 
+/**
+ * Make the end of rendering for a pictures in contexts passed with submission. 
+ * The server should start processing all pending operations for contexts.
+ * All contexts passed should be associated through vaMFAddContext
+ * and call sequence Begin/Render/End performed.
+ * This call is non-blocking. The client can start another 
+ * Begin/Render/End/vaMFSubmit sequence on a different render targets.
```
lizhong1008: As current definition of VAEndPicture, if VAEndPicture is called, the server should start processing all pending operations. Seems it will conflict with vaMFSubmit, because before vaMFSubmit, VAEndPicture has processed current frame. The mean it requires change current implementation of VAEndPicture, or VAEndPicture should not be called for the case of multi-frame. Am I right?

lizhong1008:
@artem-shaporenko, 
1. Multi-frame processing is a good feature, but seems this pull request has not described it well. Let me take an example, "Number of contexts per Multi-Frame context - should not be limited."  How we know what is the maximum number of context? 
  We have a libva sync meeting, do you think it is a good idea to invert you to present this feature? It should can help us to know more details. 
 
2. There is a rule, if a new feature is introduced to VAAPI, there should be  an open source driver which has already implemented it (and would better a open source middware has verified it), otherwise it would be merged.  
So, what feature of multi-frame has been verified by driver and middleware? decoder, encoder, vpp or the three combined case? Only encoder case verified, right? It is risky for other cases. Let me take h264 decoder as example, h264 has maximal 16 reference pictures, will it decoded error if multi-frames are send but missed some reference picture? It is also risky for multi-frame case combined encoder and vpp.

va/va.h
artem-shaporenko: >>“H264 encode or FEI is supported”，is this a typo? should be "H264 encoder of FEI is supported" or "H264 encode without FEI is supported"?
No It's not a typo, it is just an example, I can change to exact entry point names so it is more clear
>>I don't think it is a good idea to list which profile/entrypoint is not supported or not support in the API file. It means you must update this comments when driver behavior has been changed.
One  more time it is not listing of supported features, but examples(covered by 'for example').
>>I don't think it is a good idea to list which profile/entrypoint is not supported or not support in the API file. It means you must update this comments when driver behavior has been changed.
As by architecture and assumed design there are not issue to continue with single context mode without Multi-Frame, if Multi-Frame is not supported - why add excessive APIs?
>> inexplicable error
Errors are explicable as described.

va/va.h
artem-shaporenko: It doesn't require change to vaEndPicture, but to context working with multiframe. 
So context knows that it is working through multi-frame and doesn't perform submission during vaEndPicture

artem-shaporenko:
@lizhong1008 Yes we can meet and discuss.
>>Multi-frame processing is a good feature, but seems this pull request has not described it well. Let me take an example, "Number of contexts per Multi-Frame context - should not be limited." How we know what is the maximum number of context?
We have a libva sync meeting, do you think it is a good idea to invert you to present this feature? It should can help us to know more details.

Number of context added is not limited, so any number of context can be added.
>>There is a rule, if a new feature is introduced to VAAPI, there should be an open source driver which has already implemented it (and would better an open source middleware has verified it), otherwise it would be merged.
So, what feature of multi-frame has been verified by driver and middleware? decoder, encoder, vpp or the three combined case? Only encoder case verified, right? It is risky for other cases. Let me take h264 decoder as example, h264 has maximal 16 reference pictures, will it decoded error if multi-frames are send but missed some reference picture? It is also risky for multi-frame case combined encoder and vpp.

It will come together with iHD driver opensource implementation and Media SDK implementation over it. 

Multi-Frame API not dependent on reference frame and anything else, depending on driver implementation it can support particular vaEntryPoint/Profiles/etc.  Any issues with parameters for decode/vpp/encode either reference picture or anything else should be reported during vaBeginPicture/vaRenderPicture/vaEndPicture, vaMFSubmit only report issues associated with multi-frame submission. Such architecture allows to keep minimum changes for both driver and application.
Somehow missed this explanation - will add.

xhaihao:
@fhvwy could you help to provide your input from FFmpeg side?

va/va.h
lizhong1008: "One more time it is not listing of supported features, but examples(covered by 'for example')"
Yes, I know they are examples, but I think they are examples of some driver (iHD driver) implementation status (Please correct me if I am wrong). 
"so for example H264 encode or FEI is supported, but H265/Mpeg2 not supported at a moment",maybe h264/mpeg2 are supported one day,  will it be necessary to change this comment?

lizhong1008: "As by architecture and assumed design there are not issue to continue with single context mode without Multi-Frame, if Multi-Frame is not supported - why add excessive APIs?"
Here the  vaQueryXXX interface I mean to to get Multi-Frame-Processing (MFP) capacity not for singe context mode. It looks like there are many limitations for iHD driver implementation of MFP as your examples in the comments here.

va/va.h
lizhong1008: "doesn't perform submission during vaEndPicture" seems conflict with the definition of VAEndPicture: " if VAEndPicture is called, the server should start processing all pending operations". Am I right? Any driver implementation should follow and make sure without any conflict with API definition.

va/va.h
artem-shaporenko: Ok, yes, current iHD driver doesn't support mpeg2 and H265, will it be better to remove example comment or change it to something else not related to current iHD driver support, like "HEVC supported, but Mpeg2 not supported" or put particular entry point name?

Limitations are shown by particular implementation, we can introduce query interface, but how does this help, how this function will be used from an application, what is benefit over current suggested path?

va/va.h
artem-shaporenko: Yes, it conflicts with definition of vaEndPicture, but the same time all using vaMFSubmit will be aware of this change to behavior.

lizhong1008: so maybe there are two ways to avoid this conflict: 1. change the definition of VAEndPicture, add one more comment to indicate vaEndPicture won't process the pending operations in the MFP cases. 2. In the MFP case, no need to call vaEndPicture.

artem-shaporenko: Isn't it enough to have comment for vaMFSubmit?

va/va.h
artem-shaporenko: In general I don't object adding some query function, but it is not necessarily required here as can work without it - can be added later as improvement. Posted some code examples below to show why I don't think query function will be really useful in general.

va/va.h
artem-shaporenko: Posted example below why vaEndPicture behavior should be the way it described in vaMFSubmit

artem-shaporenko:
Regarding query and vaEndPicture, here are simple schematic code examples(processing is shown as serial, in real case it most likely will be multi-thread, but general code flow is something like below):

1. Showing code for multiple context before adding MFP support.
2. Added MFP support - showing current expected example 
3. Added Query function - just only MFContext logic

# 1. multiple context before adding MFP support.
VADisplay dpy;
VAContextID contexts[N];//all available contexts
VAStatus sts = VA_STATUS_SUCCESS;
....
//initialize display, etc.
.....
for( int i=0; i<N; i++)
{
    VAConfigID stream_config;
    ...
    //fill config
    ...
    sts = vaCreateConfig(dpy,...., &stream_config);
    CHECKSTS(sts);
    sts = vaCreateContext(dpy, stream_config, ...., &contexts[i]);
    CHECKSTS(sts);
}

....
// prepare streams ready to submit frames, fill contexts vector.
....
//runtime submission
for( int i=0; i<N; i++ )
{
    sts = vaBeginPicture(dpy, contexts[i], in_surf);
    CHECKSTS(sts);
    sts = vaRenderPicture(dpy, contexts[i], buffers, buffer_count);
    CHECKSTS(sts);
    sts = vaEndPicture(dpy, contexts[i]); // Delay real submission untill vaMFSubmit()
    CHECKSTS(sts);
}

# 2. MFP flow example with changes in bold, clearly visible - MFP is added in addition to current code and not breaking logic for contexts working not through MFP

VADisplay dpy;
**VAMFContextID mf_context;**
VAContextID contexts[N];//all available contexts
**bool added_contexts[N];**
VAStatus sts = VA_STATUS_SUCCESS;
**bool mfSupported;**
....
//initialize display, etc.
...
**sts = vaCreateMFContext(dpy, &mf_context);
mfSupported = (sts == VA_STATUS_SUCCESS);
.....**
for( int i=0; i<N; i++)
{
    VAConfigID stream_config;
    ...
    //fill config
    ...
    sts = vaCreateConfig(dpy,...., &stream_config);
    CHECKSTS(sts);
    sts = vaCreateContext(dpy, stream_config, ...., &contexts[i]);
    CHECKSTS(sts);
    **sts = vaAddContext(dpy, &contexts[i], mf_context); // Add this context into the Multi-Frame context
    added_contexts[i] = (sts == VA_STATUS_SUCCESS); // save context state - if it is available for multi-frame or not.**
}

**VAContextID submitContexts[N]; //frame contexts - to submit, M - max number of frames to submit, number of overall contexts associated with Multi-Frame can be bigger than maximum number of frames to submit**
....
// prepare streams ready to submit frames, fill contexts vector.
....
//runtime
**int M=0;**
for( int i=0; i<N; i++ )
{
    sts = vaBeginPicture(dpy, contexts[i], in_surf);
    CHECKSTS(sts);
    sts = vaRenderPicture(dpy, contexts[i], buffers, buffer_count);
    CHECKSTS(sts);
    sts = vaEndPicture(dpy, contexts[i]); // Delay real submission untill vaMFSubmit() if context is applicable
    CHECKSTS(sts);
    **if(added_contexts[N])
        submitContexts[M++] = contexts[i];**
}

**if(mfSupported)
{
    sts = vaMFSubmit(dpy, mf_context, contexts, M); // Submit through Multi-Frame context
}**

# 3. Added Query function - just only MFContext logic, addion from(2) shown in bold. Clearly visible that any query function added will be just an addition above context management logic, can be useful for debug at some cases, but in general complicating code and requires additional support from driver.

VADisplay dpy;
VAMFContextID mf_context;
VAContextID contexts[N];//all available contexts
bool added_contexts[N];
VAStatus sts = VA_STATUS_SUCCESS;
bool mfSupported;
**MFQueryParam query_param;//some list of paraters to report**
....
//initialize display, etc.
...
**sts = vaQueryMF(dpy, &query_param);//some vaQueryXXX function.**
...
**//check multi-frame operation support based on query functionality
mfSupported = CheckFoo(query_param);**
...
**if(mfSupported)
{**
    sts = vaCreateMFContext(dpy, &mf_context);
    mfSupported = **mfSupported &&** (sts == VA_STATUS_SUCCESS);
**}**
.....
for( int i=0; i<N; i++)
{
    VAConfigID stream_config;
    ...
    //fill config
    ...
    sts = vaCreateConfig(dpy,...., &stream_config);
    CHECKSTS(sts);
    sts = vaCreateContext(dpy, stream_config, ...., &contexts[i]);
    CHECKSTS(sts);
    **if(mfSupported)
    {**
        sts = vaAddContext(dpy, &contexts[i], mf_context); // Add this context into the Multi-Frame context
        added_contexts[i] = (sts == VA_STATUS_SUCCESS); // save context state - if it is available for multi-frame or not.
    **}**
}

VAContextID submitContexts[N]; //frame contexts - to submit, M - max number of frames to submit, number of overall contexts associated with Multi-Frame can be bigger than maximum number of frames to submit
....
// prepare streams ready to submit frames, fill contexts vector.
....
//runtime
int M=0;
for( int i=0; i<N; i++ )
{
    sts = vaBeginPicture(dpy, contexts[i], in_surf);
    CHECKSTS(sts);
    sts = vaRenderPicture(dpy, contexts[i], buffers, buffer_count);
    CHECKSTS(sts);
    sts = vaEndPicture(dpy, contexts[i]); // Delay real submission untill vaMFSubmit() if context is applicable
    CHECKSTS(sts);
    if(added_contexts[N])
        submitContexts[M++] = contexts[i];
}

if(mfSupported)
{
    sts = vaMFSubmit(dpy, mf_context, contexts, M); // Submit through Multi-Frame context
    CHECKSTS(sts);
}

fhvwy:
@xhaihao  I'm not sure there is any use-case for this API in ffmpeg.  From my reading (though do correct me if I'm wrong), it is made for maximising performance in tightly-coupled applications where a number of streams want to run pretty much in lock-step.  But, libavcodec is all built around keeping streams like this independent.

sreerenjb:
@fhvwy good point !
Initially, I thought the proposal is to encode a bunch of frames together (single stream encode use case) and in this case, we could have to treat it similar to B-frame encode use case where the framework needs to collect the raw-data-frames and rearrange. But it does not seem to be the case here.

I guess we will be having the same issue in GStreamer too, but it is not an impossible scenario. I can think of a complex multi-element bin which can act as a single plugin for application developer :)

sreerenjb:
@artem-shaporenko  I think the idea of a Query interface is necessary as @xhaihao and @lizhong1008 mentioned. It is not just for checking whether the underlined driver has support for MFContext but should be capable enough to return a status based on input Entrypoint/Profile combinations.
Because it is expensive to do "createConfig+create vacontext+addtoMFContext+destroyVAContextifNotCompatible" just to check whether driver has capability. 
An eg use case based on GStreamer: GStreamer has a tool called "gst-inspect" which query the plugin element and display the capabilities of that element to the user. This is not supposed to do any encode/decode/vpp or other expensive operations. And this is not just for "gst-inspect" but there are other scenarios like "autoplugging of different elements in a pipeline based on capabilities of that particular element".

There should be a way to query the driver capability for any feature.

sreerenjb:
The order of input arguments in vaMFAddContext() and vaMFReleaseContext () APIs seems bit strange to me. I guess the order should be like this:
"VADisplay dpy, VAMFContextID mfe_context,  VAContextID context"

sreerenjb:
@artem-shaporenko 
Can you comment about the synchronization? Does the application suppose to follow the vaSyncSurface() on the input to make sure the operation is completed?


Is there any option in Github for "reply to specific comment" ?

xhaihao:
> Is there any option in Github for "reply to specific comment" ?

You may select the corresponding text then press 'r' to reply to the specific comment.

artem-shaporenko:
@sreerenjb for mediasdk - we switch from vaSyncSurface to vaMapBuffer to synchronize encoder as it is essential for component to be synchronized by output, but not input, so for decoder or VPP it should be synchronized by output surface, encoder- by bitstream buffer.  vaSyncSurface for encoder is pretty bad in case you have several encoders running with the same source(e.g. multiple different bitrates output) - you need to wait for all encoders to finish processing before return control for application - this breaking performance. 

About compatibility -  vaSyncSurface still can work good way with Multi-frame path - you synchronize each encoder after submission within it's context the same way it was done before multi-frame added.

artem-shaporenko:
@xhaihao @lizhong1008 @sreerenjb I agree query interface can be useful, in some cases and ned to be added. This will take some time to elaborate proper interface for that and as current API allow you to get information about support an gracefully switch between single-frame and multi-frame processing, let's finalize it first, and add query as additional, separate pull request from current one.

artem-shaporenko:
@sreerenjb 

>>Because it is expensive to do "createConfig+create vacontext+addtoMFContext+destroyVAContextifNotCompatible" just to check whether driver has capability.
An eg use case based on GStreamer: GStreamer has a tool called "gst-inspect" which query the plugin element and display the capabilities of that element to the user. This is not supposed to do any encode/decode/vpp or other expensive operations. And this is not just for "gst-inspect" but there are other scenarios like "autoplugging of different elements in a pipeline based on capabilities of that particular element".

you are doing "createConfig+create vacontext" anyway for running it currently with single frame, you don't need "destroyVAContextifNotCompatible" as you still can use VAContext if it is not compatible, the only additional call you need is addtoMFContext which is free if it is not compatible, low cost if it is compatible(and if compatible it will be just used).

artem-shaporenko:
resolve conflict with merge to v2.0

sreerenjb:
> you are doing "createConfig+create vacontext" anyway for running it currently with single frame, you don't need "destroyVAContextifNotCompatible" as you still can use VAContext if it is not compatible, the only additional call you need is addtoMFContext which is free if it is not compatible, low cost if it is compatible(and if compatible it will be just used).

The use case is different, let me try to explain, this is a GStreamer specific scenario.

In order to support MFC , gstreamer-vaapi has to implement a "mutli-encoder" element, let's call it as "vaapimulitiencoder".
Now if I run a command "gst-inspect-1.0 vaapimultiencoder", it should display all the capabilities of this specific plugin element.  And for that, the element should do something like this
vaQueryMFCCapabilities (VADisplay, VAMFContext, entrypoint_combination_list, profile_combination_list). If the driver returns a SUCCESS, the "vaapimutliencoder" element advertise(and also create a few GStreamer specific things internally) the capability of those combos provided in entrypoint/profile list. Assume the "vaapimultiencoder" supports only two combos 1) h264enc+mpeg2enc   2) h264enc+hevcenc , In this case, the Query API get invoked twice with required entrypoint/profile combos. The driver just needs to return a SUCCESS/NOT-SUCCESS based on a platform-specific static table, no expensive operations at all.
Eventually, vaCreateContext/vaAddMFContext happens only once depends on user's choice of input streams.

Now assume there is no Query interface,
the "vaapimultiencoder" has to create a three VAConfig and three VAContext (h264,mpeg2 and hevc) just to advertise the capabilities of this element.

The above scenario can happen each time when an actual pipeline starts running too (because it needs to negotiate the capability with multiple elements added in the pipeline) and it will affect the "start-up" time of the pipeline significantly.

Note: The vaQueryMFCCapabilities() is just a random thought, we need something better :)

xhaihao:
@artem-shaporenko could you please change your PR to 'merge 5 commits to 01org:master'? All patches on v2.0-branch have been merged into master.

artem-shaporenko:
@sreerenjb for your multi-encoder element, what you will do if can't get multi-frame support?  Just not run? Why? The idea of multi-frame is not to give capability of multi-stream processing, but improve multi-stream performance. So if you have "gst-inspect-1.0 vaapimultiencoder" it should be able to run with or without multi-frame support. So with adding multi-frame there you will be able just to improve performance.

artem-shaporenko:
@xhaihao how to do this? Just create new pull request to master, or there is capability to change target branch from github?

xhaihao:
@artem-shaporenko no, you needn't create a new PR. There is a 'edit' button on the top-right corner, you may change your PR after click 'edit' button.

artem-shaporenko:
Thanks! I saw I can change name, but not branch.

artem-shaporenko:
@xhaihao - Changed, rebased to master
@lizhong1008 updated comments
@sreerenjb changed parameters order for Add/ReleaseContext

sreerenjb:
> @sreerenjb for your multi-encoder element, what you will do if can't get multi-frame support? Just not run? Why? The idea of multi-frame is not to give capability of multi-stream processing, but improve multi-stream performance. So if you have "gst-inspect-1.0 vaapimultiencoder" it should be able to run with or without multi-frame support. So with adding multi-frame there you will be able just to improve performance.

Okay, this is more into the GStreamer internals now :)

There are multiple ways it can work from GStreamer perspective: for eg:
1: Query the hardware/driver and never register(in GStreamer plugin registry) an element called "vaapimultiencoder" if there is no mfcontext support
2: Query the hardware/driver and work as a wrapper for already existing single encoder if there is only single entrypoint/profile supported for a combo.

The problem is that it is impossible to integrate the idea of parallel encoding into the design architecture of a single encoder. We may be able to add something, but I am pretty sure it will end up as an absolute hack. Even if we add something like that, it won't get approval from upstream since it requires changes in base classes, most probably architectural changes in GStreamer core/base library etc.
So the only way to support a feature like "optimized(or not) parallel encoding" in a single element (which can be a bin of multiple elements too) is to provide a new element like "vaapimultiencoder".
It doesn't mean that we need to duplicate everything from single encoders but still, requires significant changes.

BTW, GStreamer is just an example. Even in general perspective, we should have this feature for any media framework. As you can see every other feature that we add in libva is following the same design: There should be a way to query the feature before implementing it. Especially for MFC it is a *must needed* thing.

lizhong1008:
@artem-shaporenko , current comment looks better and the API definition is clearer.

artem-shaporenko:
@sreerenjb Yes I agree query interface required, but as it is not blocking  - my suggestion is  - to keep it out of current PR and add in next one, so it is not block current implementation of iHD driver.

artem-shaporenko:
@xhaihao, @lizhong1008 , @sreerenjb, any other concern on API? 
Query and tool in va_utils will come in later pull requests.

lizhong1008:
@artem-shaporenko  LGTM for current API definition, will review your pull request  of sample code.

artem-shaporenko:
@lizhong1008 What LGTM mean?

artem-shaporenko:
Ok, got it - "looks good for me"

lizhong1008:
@artem-shaporenko , yes it means "looks good to me"

artem-shaporenko:
@xhaihao if there are no concerns can we merge this PR into v2.0-next so it can be used for unified libva in iHD driver development?

artem-shaporenko:
combined into single commit to avoid API change in different commits.
@xhaihao please update v2.0 next branch

va/va_backend.h
```diff
@@ -133,6 +133,30 @@ struct VADriverVTable
...
+	);
+
+	VAStatus (*vaMFSubmit) (
+		VADriverContextP ctx,
+		VAMFContextID mf_context,
+		VAContextID *contexts,
+		int num_contexts
+	);
+
 	VAStatus (*vaCreateBuffer) (
```
xhaihao: Please add the new callback function pointers after the existing callback functions and reduce the reserved bytes. otherwise it will break compatibility between libva and driver.

artem-shaporenko: Done

va/va_backend.h
```diff
@@ -436,8 +436,31 @@ struct VADriverVTable
...
+
+        VAStatus (*vaMFSubmit) (
+            VADriverContextP ctx,
+            VAMFContextID mf_context,
+            VAContextID *contexts,
+            int num_contexts
+        );
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[64];
+        unsigned long reserved[63];
```
xhaihao: The new array should be an array of 60 unsigned long integers because you add 4 callback functions.

sreerenjb:
@artem-shaporenko @xhaihao 

I still have few more concerns.

I would like to take a step back and think what we really trying to achieve here: IIUC, the intention is to utilize the extra hardware blocks more efficiently.
Let's talk about this based on a simple use case: assume the Platform has 4 VMEs and we are trying to achieve maximum efficiency in the parallel encoding of 4 streams.
Here there is no way middleware components can know the capability of actual hardware(how many units available, cost of hw thread instantiation and how hw is utilizing them etc).
Which means, only the underlined driver can do optimization based on HW capability. If that is the case, I am wondering what is the real use case
of these APIs? Middleware has no idea about the cases(combinations of operations) where hw can actually do optimization in parallel operations,
It can only perform some random job scheduling with the assumption of hardware can perform better, isn't it?
Or do we need to assume that hw can always perform better if submit the jobs together? Then the next question is, how many jobs in single go can give better results :)

On the other hand, if the driver really knows the hw capabilities,  why shouldn't it possible to schedule tasks in a better way without having
any specific APIs? Let's take the previous example itself. The first encode job submitted by user space is running on first VME block and rest of them are idle(assuming hw not instantiated parallel thread for other VMEs). Now if the user requesting for a second encode (either from a same process or as a different process), driver should be internally able to monitor the hardware
utilization and assigns the second VME block for the new encode job.

Of course the new apis bring the benefits of "more gpu tasks in single batchbuffer and decrease the cpu overload". My question is, how much performance difference
it can really bring comparing with the idea of "driver internally handle all resource utilization without new APIs but with multiple batch buffers".

Please correct me If I am wrong about the understanding of hw block utilization.

fhvwy:
@artem-shaporenko  Can you clarify the behaviour of this API with respect to pipelining and dependencies between operations?  To offer some specific examples, do the following two cases work:

Internal dependency:
* Create a scaler and an encoder context and add them to an MF context.
* Submit frames to the scaler, submit the scaler output to the encoder (in the same way that pipelining currently works).
* Submit the MF context.

External dependency:
* Create two each of scaler and encoder contexts.
* Create two MF contexts, add the two scalers to one of them and the two encoders to the other.
* Submit frames to both scalers, submit the scaler outputs to the encoders.
* Submit the two MF contexts, in either order.

artem-shaporenko:
@sreerenjb you are right that performance improvement will depend on VME amount(and other factors depending on target HW). This is why we need to put more effort to implement proper query function that report enough data for different situation so application/middleware can properly manage multi-frame submission. In general talking about performance - anyway developer need to test how it works on different HW to make sure performance is in expected range. So the same is without multi-frame, you need to test how many parallel transcodings you need to do on different HW and don't have any report from driver about it.

va/va_backend.h
artem-shaporenko: Yes, sorry. done, not 60, but 56 as long is 32 bit, but pointer is 64 bit for linux 64

artem-shaporenko:
@xhaihao Can you please merge latest update for VTable changes into v2.0-next branch?

va/va_backend.h
```diff
@@ -436,8 +436,31 @@ struct VADriverVTable
...
+
+        VAStatus (*vaMFSubmit) (
+            VADriverContextP ctx,
+            VAMFContextID mf_context,
+            VAContextID *contexts,
+            int num_contexts
+        );
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[64];
+        unsigned long reserved[56];
```
xhaihao: You added 4 hook functions only, so the reserved bytes should be 60 * sizeof(unsigned long) now, not 56 * sizeof(unsigned long).

va/va.c
```diff
@@ -439,6 +442,7 @@ static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
                     CHECK_VTABLE(vaStatus, ctx, BeginPicture);
                     CHECK_VTABLE(vaStatus, ctx, RenderPicture);
                     CHECK_VTABLE(vaStatus, ctx, EndPicture);
+                    CHECK_VTABLE(vaStatus, ctx, MFSubmit);
                     CHECK_VTABLE(vaStatus, ctx, SyncSurface);
```
xhaihao: The new functions are not mandatory to implement, could you remove the checks for the new functions ?

va/va.c
```diff
@@ -1151,6 +1155,90 @@ VAStatus vaDestroyContext (
...
+    VADisplay dpy,
+    VAMFContextID *mf_context    /* out */
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+    CHECK_VTABLE(vaStatus, ctx, CreateMFContext);
```
xhaihao: Here vaStatus is set to VA_STATUS_ERROR_UNKNOWN if vaCreateMFContext, however vaCreateMFContext is not a mandatory function, so please return VA_STATUS_ERROR_UNIMPLEMENTED instead of VA_STATUS_ERROR_UNKNOWN.  The same change should be applied to other new functions

va/va.c
```diff
@@ -1225,7 +1225,7 @@ VAStatus vaMFSubmit (
 )
 {
     VADriverContextP ctx;
-    VAStatus vaStatus;
+    VAStatus vaStatus = VA_STATUS_SUCCESS;
 
     CHECK_DISPLAY(dpy);
     ctx = CTX(dpy);
```
xhaihao: It is not needed if returning VA_STATUS_ERROR_UNIMPLEMENTED at once when ctx->vtable->vaCreateMFContext is NULL, 

BTW Could you please to remove all trailing whitespaces in this patch, then squash the two commits into one commit?

va/va_backend.h
artem-shaporenko: unsigned long is 32 bit isn't it?

artem-shaporenko: I changed that to be [60], but please make sure you won't get it broken

va/va.c
artem-shaporenko: Done

va/va.c
artem-shaporenko: Done

va/va_backend.h
xhaihao: unsigned long is 32bit on 32bit OS and 64bit on 64bit OS

va/va.c
```diff
@@ -1151,6 +1151,96 @@ VAStatus vaDestroyContext (
...
+    VAMFContextID *mf_context    /* out */
+)
+{
+    VADriverContextP ctx;
+    VAStatus vaStatus;
+    
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+    if(ctx->vtable->vaCreateMFContext == NULL)
+        vaStatus = VA_STATUS_ERROR_UNIMPLEMENTED;
```
xhaihao: *mf_context may be a random value for this case, so va_TraceCreateMFContext shouldn't be called when ctx->vtable->vaCreateMFContext is a NULL pointer.

artem-shaporenko: agree, fixed

PR #114, opened 2017-09-11
Some comment update

xhaihao:
* comment about surface and context
* comment about VA-API
* comment about num_ref_idx_l0_active_minus1 and num_ref_idx_l1_active_minus1
* comment about packed header

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -615,11 +615,26 @@ typedef struct _VAConfigAttrib {
 /**@{*/
 /** \brief Driver does not support any packed headers mode. */
 #define VA_ENC_PACKED_HEADER_NONE       0x00000000
-/** \brief Driver supports packed sequence headers. e.g. SPS for H.264. */
+/**
+ * \brief Driver supports packed sequence headers. e.g. SPS for H.264.
+ *
+ * Application must provide it to driver once this flag is returned through
+ * vaGetConfigAttributes()
+ */
 #define VA_ENC_PACKED_HEADER_SEQUENCE   0x00000001
-/** \brief Driver supports packed picture headers. e.g. PPS for H.264. */
+/**
+ * \brief Driver supports packed picture headers. e.g. PPS for H.264.
+ *
+ * Application must provide it to driver once this falg is returned through
+ * vaGetConfigAttributes()
+ */
 #define VA_ENC_PACKED_HEADER_PICTURE    0x00000002
-/** \brief Driver supports packed slice headers. e.g. \c slice_header() for H.264. */
+/**
+ * \brief Driver supports packed slice headers. e.g. slice_header() for H.264.
+ *
+ * Application must provide it to driver once this flag is returned through
+ * vaGetConfigAttributes()
+ */
 #define VA_ENC_PACKED_HEADER_SLICE      0x00000004
 /**
  * \brief Driver supports misc packed headers. e.g. SEI for H.264.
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -864,14 +881,14 @@ VAStatus vaQueryConfigAttributes (
 /**
  * Contexts and Surfaces
  *
- * Context represents a "virtual" video decode pipeline. Surfaces are render 
- * targets for a given context. The data in the surfaces are not accessible  
- * to the client and the internal data format of the surface is implementatin 
- * specific. 
+ * Context represents a "virtual" video decode, encode or video processing
+ * pipeline. Surfaces are render targets for a given context. The data in the
+ * surfaces are not accessible to the client except derived image is supported
```
chivakker: client except if derived image is supported

va/va.h
```diff
@@ -864,14 +881,14 @@ VAStatus vaQueryConfigAttributes (
...
+ * and the internal data format of the surface is implementatin specific.
```
chivakker: is implementation

va/va.h
```diff
@@ -864,14 +881,14 @@ VAStatus vaQueryConfigAttributes (
...
  *
- * Surfaces will be bound to a context when the context is created. Once
- * a surface is bound to a given context, it can not be used to create  
- * another context. The association is removed when the context is destroyed
+ * Surfaces are provided as a hint of what surfaces will be used when the context
+ * is created. A surface may be shared between different contexts, e.g. a surface
+ * is used as the output of a decode context and the input of a video process context.
```
chivakker: it should describe who and how the surface ownership is maintained, i.e. what happens when a different context uses and destroys them

xhaihao:
@chivakker thanks for the comments, I updated the corresponding commit, could you please review again?

va/va.h
```diff
@@ -1126,7 +1129,7 @@ VAStatus vaDestroySurfaces (
  *  picture_height: coded picture height
  *  flag: any combination of the following:
  *    VA_PROGRESSIVE (only progressive frame pictures in the sequence when set)
- *  render_targets: render targets (surfaces) tied to the context
+ *  render_targets: a hint for render targets (surfaces) tied to the context
```
fhvwy: Making the render targets optional is a driver API change.  Is that definitely ok everywhere?

chivakker: It would be up to the driver to make them tied to the context.  intel-vaapi-driver doesn't force that.

fhvwy: A driver which relies on all render targets being specified in advance will be broken if they are made optional and the user then does not supply them.  This is not just about tying the surfaces to the context, they might also be required for the context creation itself - perhaps because the targets must be known directly, or because some properties of the surfaces might be needed.

fhvwy: Mesa does some slightly subtle things with it here: <https://cgit.freedesktop.org/mesa/mesa/tree/src/gallium/state_trackers/va/context.c>.  I'm not sure that the VPP test is doing anything useful, but it looks like that max_references test for H.265 might.  The surfaces themselves aren't used.

xhaihao: * Even in mesa, render_targets is not used directly in vaCreateContext(). 
* Mesa developers also think max_references cannot be based on num_render_targets
> commit b4fb6d76161d86c67697cf28a221d7913b8d084d
> Author: Julien Isorce <j.isorce@samsung.com>
> Date:   Thu Nov 26 08:29:54 2015 +0000
>
>    st/va: delay decoder creation until max_references is known
>
>    In general max_references cannot be based on num_render_targets.
* It is weird for me VPP and other profiles have different requirement on render_targets in Mesa
* The old API doesn't allow two contexts use the same surface at the same time, so render_targets is required when creating a context. 
> Once a surface is bound to a given context, it can not be used to create another context

xhaihao: It would be better to ask mesa developer to remove the max_references test for H.265.

PR #115, opened 2017-09-12
Use a flexible array member

xhaihao:
ISO C/C++ forbids zero-size array, but flexible array member is
supported since C99.

This fixes https://github.com/01org/libva/issues/76


xhaihao:
@uartie Could you approve the PR so 'Review required' is in green?

PR #116, opened 2017-09-13
Add reserved bytes to most structures

xhaihao:
The reserved bytes must be set to 0, then we can add new fields to those
structures in future without worrying about the API compatibility too
much.

### DIFF
--- a/va/va_dec_hevc.h
+++ b/va/va_dec_hevc.h
@@ -190,6 +190,8 @@ typedef struct  _VAPictureParameterBufferHEVC
      */
     uint32_t                st_rps_bits;
 
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                va_reserved[VA_PADDING_MEDIUM];
 } VAPictureParameterBufferHEVC;
 
 /**
@@ -321,6 +323,8 @@ typedef struct  _VASliceParameterBufferHEVC
     uint8_t                 five_minus_max_num_merge_cand;
     /**@}*/
 
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                va_reserved[VA_PADDING_LOW];
 } VASliceParameterBufferHEVC;
 
 
@@ -379,6 +383,9 @@ typedef struct _VAIQMatrixBufferHEVC
      * with sizeID = 3 and matrixID in the range of 0 to 1, inclusive.
      */
     uint8_t                 ScalingListDC32x32[2];
+
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                va_reserved[VA_PADDING_LOW];
 } VAIQMatrixBufferHEVC;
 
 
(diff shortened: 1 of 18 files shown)


fhvwy:
"__reserved" is in the implementation namespace, and plausibly something a compiler could use in a language extension (e.g. <https://msdn.microsoft.com/en-us/library/windows/desktop/aa383701(v=vs.85).aspx#__reserved>).  Maybe just call it "reserved", or "va_reserved" if that could collide?

va/va_enc_hevc.h
```diff
@@ -316,6 +316,9 @@ typedef struct _VAEncSequenceParameterBufferHEVC {
     uint8_t     max_bytes_per_pic_denom;
     /** \brief Same as the HEVC bitstream syntax element. */
     uint8_t     max_bits_per_min_cu_denom;
+
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                __reserved[VA_PADDING_LOW];
```
yakuizhao: It will be better to reserve more bytes for HEVC.
For the SPS parameter: it can be PADDING_MEDIA.

va/va_enc_hevc.h
```diff
@@ -517,6 +520,9 @@ typedef struct _VAEncPictureParameterBufferHEVC {
         } bits;
         uint32_t        value;
     } pic_fields;
+
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                __reserved[VA_PADDING_LOW];
```
yakuizhao: This fields can be HIGH for the further extension.

va/va_enc_vp9.h
```diff
@@ -506,6 +510,8 @@ typedef struct  _VAEncPictureParameterBufferVP9
      */
     uint32_t    skip_frames_size;
 
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                __reserved[VA_PADDING_LOW];
```
yakuizhao: It is similar to HEVC.
It will be better to reserve more bytes for VP9. 
The EncPictureParameterBufferVP9 can be VA_PADDING_MEDIUM.

va/va_dec_vp9.h
```diff
@@ -191,6 +191,9 @@ typedef struct  _VADecPictureParameterBufferVP9
      */
     uint8_t                 bit_depth;
 
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                __reserved[VA_PADDING_LOW];
+
```
yakuizhao: This can be configured as MEDIUM.

va/va_dec_hevc.h
```diff
@@ -190,6 +190,8 @@ typedef struct  _VAPictureParameterBufferHEVC
      */
     uint32_t                st_rps_bits;
 
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                __reserved[VA_PADDING_LOW];
```
yakuizhao: This can be configured as MEDIUM.

xhaihao:
@fhvwy @yakuizhao I updated the patch, could you review again?

PR #120, opened 2017-09-19
va_internal:  list va_internal on noinst_HEADERS

chivakker:
va_internal.h is not to be installed

Signed-off-by: Daniel Charles <daniel.charles@intel.com>

### DIFF
--- a/va/Makefile.am
+++ b/va/Makefile.am
@@ -63,6 +63,7 @@ libva_source_h = \
 libva_source_h_priv = \
 	sysdeps.h		\
 	va_fool.h		\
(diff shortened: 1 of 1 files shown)


oliver-sang:
the reason of this change is that va_internal.h missed in 'make dist'. then the tar-ball created by 'make dist' actually becomes useless because the source codes within it are not compileable any longer.

regarding install, will headers in libva_source_h_priv be installed? I saw another list libva_source_h.

oliver-sang:
sorry, just found the change has already commited.

PR #121, opened 2017-09-20
vafool: add return value check and clean the build warning.

mypopydev:
check read return value and clean the build warning.


va/va_fool.c
```diff
@@ -290,7 +291,8 @@ static int va_FoolFillCodedBufEnc(struct fool_context *fool_ctx)
     }
     if (fd != -1) {
         fool_ctx->segbuf_enc = realloc(fool_ctx->segbuf_enc, file_stat.st_size);
-        read(fd, fool_ctx->segbuf_enc, file_stat.st_size);
+        size = read(fd, fool_ctx->segbuf_enc, file_stat.st_size);
+        assert(size == file_stat.st_size);
```
chivakker: instead of asserting it could be returning an error. 

ditto on the other places.

mypopydev: Maybe give a warning message when size != file_stat.st_size, but need to add a new va_warnMessage() function in va.c.

fhvwy:
See https://github.com/01org/libva/commit/f5d40df0636c3767e0a47156d0c53987f86acb53 on the v2 branch.  That doesn't actually depend on v2, it could be merged earlier if useful.

mypopydev:
I had not noticed you have submitted the fix. @fhvwy

PR #71, opened 2017-06-16
add new vpp features

xfengcarl:
new features:Deblocking filter, FRC filter, Total Color Correction, Non-Linear Anamorphic Scaling.Image Stabilization...
new color standard: sRGB stRGB, xvYCC601, xvYCC709, BT.2020

Signed-off-by: xfengcarl <carl.zhang@intel.com>

PR #127, opened 2017-10-01
RFC: Add more complete colour properties for use in VPP

fhvwy:
Currently only a very limited set of colour properties are actually
supported, and colours are therefore often incorrect after passing
through VPP (especially when converting to/from RGB).  This attempts
to define general colour properties which can be used to cover all
videos which might be encountered.

In particular, it is intended to cover all detail which can be found
in the VUI section of an H.264/H.265 stream.  As in those standards,
some values are defined with reference to ISO/IEC 23001-8 and H.273.
It therefore implicitly supports many newer HDR standards which are
not currently expressible, such as PQ, HDR10 and HLG.

No implementation yet - I'm looking for thoughts on improving the colour support and something like this seems a reasonable starting point.  Just adding more colour standard values is not scalable as standards proliferate (especially for HDR) and we end up with videos combining elements of many different ones, so we need a more general approach something like this.


va/va_vpp.h
```diff
@@ -456,10 +584,14 @@ typedef struct _VAProcPipelineParameterBuffer {
...
     uint32_t        output_background_color;
     /**
-     * \brief Requested output color primaries.
+     * \brief Requested output color standard.
      */
     VAProcColorStandardType output_color_standard;
     /**
+     * \brief Requested output color properties.
+     */
+    VAProcColorStandardType output_color_properties;
```
lu-zero: VAProcColorProperties I guess.

@kodabb might chime in regarding the need of more fields.

wm4:
This seems good and useful.

kodabb:
seems like a good idea, also maybe mention 23091-9

va/va_vpp.h
```diff
@@ -286,22 +286,152 @@ typedef enum _VAProcColorBalanceType {
...
+    VAProcChromaSampleLocationCenter = 1,
+    /** Top-left. */
+    VAProcChromaSampleLocationTopLeft = 2,
+    /** Top. */
+    VAProcChromaSampleLocationTop = 3,
+    /** Bottom-left. */
+    VAProcChromaSampleLocationBottomLeft = 4,
+    /** Bottom. */
+    VAProcChromaSampleLocationBottom = 5,
+} VAProcChromaSampleLocation;
```
xhaihao: @xfengcarl do you have any comment for this PR,  you have a similar patch for chroma sample location.

xhaihao: @xfengcarl what's your comment?

xfengcarl: there are same definition in PR71. check it.

fhvwy: I'd be fine with adopting the same flag structure for chroma siting and full/limited range as there if preferred.  Unclear whether it should still be inside the ColorProperties structure, though - are the flags going to be used anywhere else?  (Currently you have `input_surface_flag` and `output_surface_flag` in the same roles as `surface_color_properties` and `output_color_properties` above.)

xfengcarl: you will integrate the color properties together, so you want to remove input_surface_flag & output_surface_flag from the PR71?

jackie-guo: In xfengcarl's code, the bit fields in input_surface_flag and output_surface_flag are defined as following.
bit[31] - encrypted or not
bit[16] - full range or reduced range
bit[1~0] - vertical chromasiting
bit[3~2] - horizontal chromasiting

What about using the remaining bit fields in input_surface_flag and output_surface_flag for colour_primaries,  transfer_characteristics and matrix_coefficients. Below are examples (Assume 3-bit is enough for each one).
bit[6~4] - colour_primaries
bit[9~7] - transfer_characteristics
bit[12~10] - matrix_coefficients

fhvwy: @jackie-guo  Ideally they should be 8 bits to match use in MPEG-2 (uimsbf(8) in sequence display extension), H.264 (u(8) in VUI) and H.265 (also u(8) in VUI).  They could be smaller at the moment (largest current value is 22), but that might cause problems later as future versions of the standard specify more values.

While that just about fits if we cram them in to the surface flags, I think it would be cleaner for them to be standalone uint8_t values.  (A bitfield setup could also work?)

@xfengcarl  I think so?  If `input_surface_flag` and `output_surface_flag` are going to contain other things unrelated to colour then do keep them, but I would move the colour properties to the new structure so that they are together.

jackie-guo: Although some ITU standards have defined colour_primaries/transfer_characteristics/matrix_coefficients, but some other ISO standards or some company standards have not defined colour_primaries/transfer_characteristics/matrix_coefficients. So current driver only takes color range and chromasiting into consideration, does not take colour_primaries/transfer_characteristics/matrix_coefficients into consideration.

kodabb: that's a wrong approach IMO, the driver should take the MAXIMUM common denominator and then proactively fill in the information that is missing (if needed), not exclude necessary fields from its API

also what are the ISO standards that you refer to lacking these properties? they are usually pretty accurate

xfengcarl: @fhvwy do you have concerns about using union instead of structure?

xfengcarl: I add a  structure in #71 
typedef struct _VAProcColorProperties {
    /** Chroma sample location.\c VA_CHROMA_SITING_VERTICAL_XXX | VA_CHROMA_SITING_HORIZONTAL_XXX */
    uint8_t chroma_sample_location;
    /** Chroma sample location. \c VA_SOURCE_RANGE_XXX*/
    uint8_t color_range;
    uint8_t reserved[6];
} VAProcColorProperties;
and there are 6 reserved bytes to be used , you can combine your pr with mine smoothly
thanks

xhaihao: Thanks for the new definition. The solution looks fine to me.

va/va_vpp.h
```diff
@@ -286,22 +286,152 @@ typedef enum _VAProcColorBalanceType {
...
+ * RGB data always use the identity matrix (matrix_coefficients = 0).
+ */
 typedef enum _VAProcColorStandardType {
     VAProcColorStandardNone = 0,
-    /** \brief ITU-R BT.601. */
+    /** \brief Explicitly specified color properties.
+     *
+     * Use corresponding color properties section.
+     */
+    VAProcColorStandardExplicit,
```
xhaihao: Could you move VAProcColorStandardExplicit to the end of union? Otherwise an application built against an old version of libva won't work as expected when VAProcColorStandardBT601 is used.

va/va_vpp.h
fhvwy: @xfengcarl  Yep, that looks good to me.  I'll try to review your branch again soon, and then rebase this on top of it.

va/va_vpp.h
fhvwy: Will do.  (A leftover from the unstable ABI.  The structure size change which was also missing isn't needed, because that will now be inherited from #71.)

xhaihao:
@fhvwy is there an update for this PR?

xhaihao:
@fhvwy  will you have a chance to update this PR?

fhvwy:
@xhaihao  Here's an updated version.  This has been mostly on hold because I don't have a real implementation to compare against, but if someone else is interested in adding HDR and similar support then this should be sufficient.

There is one outstanding question here: do you know what the colour primaries and transfer function are used by stRGB?

va/va_vpp.h
fhvwy: @FurongZhang  Set the appropriate ```color_standard``` field to ```VAProcColorStandardExplicit``` to indicate that the explicit fields are used, then set the fields in the matching ```color_properties``` structure to your values - ```colour_primaries``` to 14 and ```transfer_characteristics``` to 16.  The ```matrix_coefficients``` field will also need to be set in this case - for RGB it will be 0, while for YUV I believe it uses the BT.2020 NCL value which is 9.

FurongZhang: @fhvwy Then, I agree with you on this. Is it possible for you to merge this PR to VAAPI? We could use this for HDR color.

PR #128, opened 2017-10-02
Add a route for drivers to send messages to the API message callbacks

fhvwy:
The important thing here is to make logging easier from drivers.  Currently the only logging route they have is stderr, which isn't necessarily visible anywhere and can causes problems for end users when messages appear on the console - drivers therefore avoid logging anything at all except in exceptional circumstances.  With this mechanism, hopefully drivers can log more useful information for application developers to assist development and debugging, particularly around the more complex parameter structures which have a large number of poorly-documented requirements.

### DIFF
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -552,7 +552,48 @@ struct VADriverContext
 
     char *override_driver_name;
 
-    unsigned long reserved[41];         /* reserve for future add-ins, decrease the subscript accordingly */
+    void *pDisplayContext;
+
+    /**
+     * Error callback.
+     *
+     * This is set by libva when the driver is opened, and will not be
+     * changed thereafter.  The driver can call it with an error message,
+     * which will be propagated to the API user through their error
+     * callbacks, or sent to a default output if no callback is available.
+     *
+     * It is expected that end users will always be able to see these
+     * messages, so it should be called only for serious errors.  For
+     * example, hardware problems or fatal configuration errors.
+     *
+     * @param pDriverContext  Pointer to the driver context structure
+     *                        being used by the current driver.
+     * @param message  Message to send to the API user.  This must be a
+     *                 null-terminated string.
+     */
+    void (*error_callback)(VADriverContextP pDriverContext,
+                           const char *message);
+    /**
+     * Info callback.
+     *
+     * This has the same behaviour as the error callback, but has its
+     * own set of callbacks to the API user.
+     *
+     * It should be used for informational messages which may be useful
+     * for an application programmer or for debugging.  For example, minor
+     * configuration errors, or information about the reason when another
+     * API call generates an error return.  It is not expected that end
+     * users will see these messages.
+     *
+     * @param pDriverContext  Pointer to the driver context structure
+     *                        being used by the current driver.
+     * @param message  Message to send to the API user.  This must be a
+     *                 null-terminated string.
+     */
+    void (*info_callback)(VADriverContextP pDriverContext,
+                          const char *message);
+
+    unsigned long reserved[38];         /* reserve for future add-ins, decrease the subscript accordingly */
 };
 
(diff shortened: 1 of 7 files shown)


sreerenjb:
This seems to be a good idea , +1

xhaihao:
>> It is intended to be used to replace all use of logging to stdout/stderr in drivers

Do you mean we use ctx->info_callback() or ctx->error_callback() only  and needn't redefine the callbacks in the drivers?

va/va_backend.h
```diff
@@ -529,6 +529,36 @@ struct VADriverContext
...
+     * These should be used for informational messages which may be useful
+     * for an application programmer or for debugging.  For example, minor
+     * configuration errors, or information about the reason when another
+     * API call generates an error return.  It is not expected that end
+     * users will see these messages.
+     */
+    void (*info_callback)(VADriverContextP pDriverContext,
+                          const char *message);
+
     unsigned long reserved[41];         /* reserve for future add-ins, decrease the subscript accordingly */
```
xhaihao: Please decrease the subscript of the field 'reserved' below for compatibility.

fhvwy:
@xhaihao  Yes: these callbacks are set by libva when the driver is opened and should not be changed in the driver.  When the driver calls them, libva forwards the messages on to the user callbacks set by vaSetErrorCallback and vaSetInfoCallback.

va/va_backend.h
fhvwy: There is no release to be compatible with yet.  (The reserved size is presumably going to be reset before the release of libva2?)

xhaihao:
Could you more comments in .h about the usage of the new callbacks?

sreerenjb:
@fhvwy this is not exactly related to this PR but a minor mistake from your old patch to enable callbacks:
https://github.com/01org/libva/blob/master/va/va.c#L180 
is returning "old error_callback" from vaSetInfoCallback instead of "old info callback"

fhvwy:
@xhaihao  Updated; hopefully it is now a bit clearer.

@sreerenjb  Thanks for spotting that!  I've included a fix here.

xhaihao:
@fhvwy Could you fix the returning of old info callback in a new PR? I'd like to merge the fix firstly.

lizhong1008:
Good idea.  One question, it is necessary to add one more log level, for example: debug_callback()?

fhvwy:
@lizhong1008  I don't think so.  The two classes of messages are, to my mind, "for end users, when errors occur" (error) and "for API users, to help to diagnose problems" (info).

To try to offer some examples, I'm imagining that error messages are things like:
* "Driver initialisation failed: versions are not compatible"
* "GPU hang happened while decoding: input stream is probably corrupt"
* "Failed to render surface: X11 connection lost"
* "HRD buffer overflow while encoding: output stream bitrate may be incorrect"

Whereas info messages are more like:
* "vaCreateConfig: Profile 19 entrypoint 6 not supported on this hardware"
* "vaQuerySurfaceAttributes: Invalid surface 0xffffffff"
* "vaCreateSurfaces: Failed to create surfaces: unsupported format P016"
* "vaRenderPicture: Unknown buffer type 44497"

Is there anything which an API user might want to know which is not covered by either of these?  I think lower-level information than this is likely to be only suitable for driver developers, and therefore probably shouldn't be passing through this API.

lizhong1008:
@fhvwy , got it. Yes, lower-level information is mainly for driver developers.  I asked such a question because I noted few log message is provided in i965 driver. Looks like it should be handled by driver internally instead of passing through VAAPI.

lizhong1008:
+1

xhaihao:
@fhvwy could you help to rebase your messages branch? I resolved the conflicts on github,  but rebasing your commit on top of the base branch is still blocked.

PR #130, opened 2017-10-10
Fix returning of old info callback

fhvwy:
Split from https://github.com/01org/libva/pull/128 as requested by @xhaihao.


mypopydev:
LGTM

PR #131, opened 2017-10-30
Expose the encode passes interface.

hongbolv:
Report the encode passes number status to application through vaapi.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -2434,6 +2434,16 @@ VAStatus vaBufferSetNumElements (
 #define VA_CODED_BUF_STATUS_AIR_MB_OVER_THRESHOLD	0xff0000
 
 /**
+ * \brief The coded buffer segment status contains frame encoding passes number
+ *
+ * This is the mask to get the number of encoding passes from the coded
+ * buffer segment status. 
+ * NUMBER_PASS(bit24~bit27): the number for encoding passes executed for the coded frame.
+ * 
+ */
+#define VA_CODED_BUF_STATUS_NUMBER_PASSES_MASK          0xf000000
+
+/**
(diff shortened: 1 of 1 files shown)


va/va.h
```diff
@@ -2434,6 +2434,13 @@ VAStatus vaBufferSetNumElements (
 #define VA_CODED_BUF_STATUS_AIR_MB_OVER_THRESHOLD	0xff0000
 
 /**
+ * \brief The frame encoding PAK Pass Number
+ *
+ * This flag indicates that the how many PAK passes during frame encoding
+ */
+#define VA_CODED_BUF_STATUS_NUMBER_PASSES               0xf000000
+
```
xhaihao: Could you describe how does an application use this flag in detail?

va/va.h
```diff
@@ -2433,6 +2433,16 @@ VAStatus vaBufferSetNumElements (
 #define VA_CODED_BUF_STATUS_FRAME_SIZE_OVERFLOW         0x1000
 #define VA_CODED_BUF_STATUS_AIR_MB_OVER_THRESHOLD	0xff0000
 
+/**
+ * \brief The coded buffer segment status contains frame encoding passes number
+ *
+ * This is the mask to get the number of encoding passes from the coded
+ * buffer segment status. 
+ * NUMBER_PASS(bit24~bit28): the number for encoding passes executed for the coded frame.
```
xhaihao: Are you sure this mask has 5 bits ? If so, the mask should be 0x1f000000

xhaihao: What is the default value if the backend driver doesn't set the pass number in the coded buffer?

PR #133, opened 2017-11-06
android: remove useless vaPutSurface

xuguangxin:
vaPutSurface only used in libva-utils. The actual application/omx
implementation never use it.
It will introduce many header files in android framework.Android change
rapidly, the directory or file name changed in versionto version. It's
hard to catch up and maintain support for many versionof android.


xhaihao:
I'm OK to remove vaPutSurface() from libva-android if vaPutSurface() is not used at all in actual applications on Android.  Could you ask other Android folks help to review the patch?

xhaihao:
Any comment from Android developers?

xuguangxin:
@XiaosongWei, could you help + 1 for me?
thanks

PR #139, opened 2017-11-12
va_drm: dlopen correct version of libva-x11

sebastinas:
Signed-off-by: Sebastian Ramacher <sramacher@debian.org>


sebastinas:
Looking at the code there: why doesn't libva-drm2 simply link against libva-x11?

xhaihao:
@sebastinas  libva-drm can be used without libva-x11.

Issue #144, opened 2017-11-21
define missing enum in case


ceyusa:
compiling with clang got these warnings:

```
va_str.c:31:13: warning: enumeration value 'VAProfileH264Baseline' not handled in switch [-Wswitch]         switch (profile) {                                                                                  
            ^                                                                                           va_str.c:79:13: warning: enumeration values 'VAConfigAttribFEIFunctionType', 'VAConfigAttribFEIMVPredictors', and 'VAConfigAttribTypeMax' not handled in switch [-Wswitch]                                      
    switch (configAttribType) {                                                                       
            ^                                                                                           
va_str.c:104:13: warning: 7 enumeration values not handled in switch: 'VAEncQPBufferType', 'VAEncFEIMVBu
fferType', 'VAEncFEIMBCodeBufferType'... [-Wswitch]                                                     
    switch (bufferType) {                                                                               
            ^                                                                        
3 warnings generated.                               
```

PR #145, opened 2017-11-21
va: Add missing enums to string conversions

ceyusa:
Those enum were recently added and their conversion to strings were
missing.

Fixes: #144


uartie:
Maybe we should add `-Werror=switch` to the default project compiler flags to prevent this in the future?

uartie:
...and `-Werror=switch-enum` to force all cases to be handled?

uartie:
Of course, the latter flag causes a bunch of errors in va_trace.c

xhaihao:
@uartie agree to add the flags, could you provide the patches?

xhaihao:
@xfengcarl could you update your PRs to handle the added new buffer types? Otherwise user will experience the same after merging your PRs.

PR #153, opened 2017-11-22
Make older versions of GCC happy

xhaihao:
Some older versions of GCC fail to build libva since -Wall -Werror compiler
flags are used for stricter compilation

  CC       va_fglrx.lo
va_fglrx.c: In function 'VA_FGLRXGetClientDriverName':
va_fglrx.c:243:33: error: 'ADL_Main_Control_Destroy' may be used uninitialized in this function [-Werror=maybe-uninitialized]

From the code path, ADL_Main_Control_Destroy has been initialized if is_adl_initialized is 1.
however the older versions of GCC are not smart enough for an automatic variable, so I ignore
the error using -Wno-maybe-uninitialized

This fixes https://github.com/01org/libva/issues/152

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -301,8 +301,8 @@ if test "$USE_DRM:$USE_X11:$USE_WAYLAND" = "no:no:no"; then
     AC_MSG_ERROR([Please select at least one backend (DRM, X11, Wayland)])
 fi
 
-AC_SUBST([AM_CFLAGS], ["-Wall -Werror"])
-AC_SUBST([AM_CXXFLAGS], ["-Wall -Werror"])
+AC_SUBST([AM_CFLAGS], ["-Wall -Werror -Wno-maybe-uninitialized"])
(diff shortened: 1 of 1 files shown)


fhvwy:
There is `-Wno-error=maybe-uninitialized` to stop the warning being an error rather than disabling it entirely.  Also note that adding this option will break building with gcc < 4.6, which don't support it - that's probably fine, but it might want to be checked at configure rather than build time (still-supported RHEL/CentOS 6 use gcc 4.4).

(As noted on the driver PR, I would still recommend disabling `-Werror` for releases, because this sort of problem will happen again.)

uartie:
@xhaihao Should we revert the "strict flags" commit and just have CI set them via user CFLAGS instead?

uartie:
Please consider #155 instead

xhaihao:
@uartie It's fine for me if you can add these flags for CI only. I will revert commit 6ca99be

xhaihao:
Close this PR and use https://github.com/01org/libva/pull/155 instead

Issue #151, opened 2017-11-22
vaPutSurface XError race condition with X11 pixmap


Ferocity:
Using vaPutSurface with a X11 pixmap instead a X11 window drawable
could cause a X11 abort through an unhandled XError.

I tracked down the problem to a Drawable detection mechanism in va/x11/va_dricommon.c.
XGetWindowAttributes does as expected throw an async XError then a pixmap is used, but
error handler may resetted to early => X11 abort called.

A possible fix is to use a XSync before untrap.
Any Performance concerns ?

Greetings,
Christian

```
diff --git a/va/x11/va_dricommon.c b/va/x11/va_dricommon.c
index 379e3d4..f85a134 100644
--- a/va/x11/va_dricommon.c
+++ b/va/x11/va_dricommon.c
@@ -55,6 +55,7 @@ is_window(Display *dpy, Drawable drawable)
 
     x11_trap_errors();
     XGetWindowAttributes(dpy, drawable, &wattr);
+    XSync(dpy, False);
     return x11_untrap_errors() == 0;
 }
```

xhaihao:
I'm fine with your fix, could you submit a PR?

Ferocity:
Today I have seen that XSync has not fixed this problem. Maybe I only have changed the timing on my last try.
I will do some more investigations on this problem. XGetWindowAttributes does already wait on these errors (BadWindow) as I have seen inside the code. 
So this problem seems not be related to libva code and could be closed.

Sorry for any inconvenience,

PR #154, opened 2017-11-23
X11: sync xerrors before unset error handler

Ferocity:
fix for #151 vaPutSurface XError race condition with X11 pixmap


Ferocity:
XGetWindowAttributes itself has a synchronization mechanism, so XSync doesn't help

Issue #168, opened 2018-01-10
when building with meson va_fei_hevc.h is not installed


ceyusa:
Thus, preproprocessing va.h will fail because va_fei_hevc.h is not available.

PR #169, opened 2018-01-10
meson: add va_fei_hevc.h

ceyusa:
Fixes: #168


ceyusa:
That failure in travis is not because of the patch

PR #176, opened 2018-01-25
android: fix typo introduced by 4cc9a74b

xuguangxin:


xhaihao:
@uartie is it possible to check building on Android in intel/ci?

Issue #79, opened 2017-06-29
vaInitialize fails under Gnome Shell Wayland


vanvugt:
vaInitialize fails when Xwayland is running. So VA-API never works at all in Gnome Shell on Wayland.

It appears the problem is libva only supports DRI2. And Xwayland only supports DRI3.

I understand you could call this a bug in the video player apps for not choosing native wayland first, but this problem is prevalent in every single video player app that supports VA-API I have encountered (gst-play-1.0, totem, mpv). So trying to fix each video player isn't really a solution that scales (also my first attempt at doing so was rejected by mpv).

chivakker:
Did you compile libva and intel-vaapi-driver with Wayland support?

stuarta:
@chivakker  I'm seeing what looks like this issue with the stock libva shipped with fedora 26 (libva-1.8.3-1.fc26.x86_64), and the corresponding libva-intel-driver from rpmfusion (libva-intel-driver-1.8.3-2.fc26.x86_64).

I raised https://bugzilla.redhat.com/show_bug.cgi?id=1491186 for fedora, and we have done some investigation there, there is also a simplified reproducer.

Regards
Stuart

vanvugt:
I did not compile libva. Just using the standard Ubuntu installation. We have it working now, including Wayland.

The problem and the solution are both detailed in the first comment already, so you might want to read that...

For Ubuntu what we did to work around the issue was to prioritize Wayland over X11 in both mpv and gstreamer-opengl (to aid gstreamer-vaapi). So that way video players don't attempt to use X11 with Xwayland and then give up. They use native Wayland instead.

vanvugt:
The proper solution for libva would be to support DRI3, so that libva doesn't fail when talking to Xwayland.

xhaihao:
I filed https://github.com/01org/libva/issues/122 to track the support for DRI3.

vanvugt:
While strictly accurate, it might be harsh and unnecessary for me to suggest libva needs DRI3 support.

[...]

PR #180, opened 2018-01-30
Enable DRI3 support for libva

jayesht:
vaInitialize used to fail when Xwayland is running blocking VA-API in Gnome Shell on Wayland.
Root cause was that libva only supported DRI2 whereas Xwayland only supported DRI3.
https://github.com/intel/libva/issues/79 

This pull request enables support for DRI3 protocol for Direct Rendering via libva. With this libva supports standard dri3 apis which dri clients can utilize.

The thing to take care of is that under DRI3 clients allocate themselves their render buffers instead of relying on the X Server for doing the allocation. 

Use --enable-dri3 flag to enable this feature.

### DIFF
--- a/va/Makefile.am
+++ b/va/Makefile.am
@@ -103,6 +103,9 @@ libva_x11_la_LDFLAGS		= $(LDADD)
 libva_x11_la_DEPENDENCIES	= libva.la x11/libva_x11.la
 libva_x11_la_LIBADD		= libva.la x11/libva_x11.la \
 	$(LIBVA_LIBS) $(X11_LIBS) $(XEXT_LIBS) $(XFIXES_LIBS) $(DRM_LIBS) -ldl
+if USE_DRI3
+libva_x11_la_LIBADD		+= $(DRI3_LIBS) -ldl
+endif
 endif
 
 if USE_GLX
--- a/va/x11/va_dri3.c
+++ b/va/x11/va_dri3.c
@@ -1,0 +1,217 @@
+/*
+ * Copyright (c) 2018 Intel Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <fcntl.h>
+
+#include "va_dri3.h"
+
+Pixmap VA_DRI3_create_pixmap(Display *dpy,
+                             Drawable draw,
+                             int width, int height, int depth,
+                             int fd, int bpp, int stride, int size)
+{
+    xcb_connection_t *c = XGetXCBConnection(dpy);
+    if (fd >= 0) {
+        xcb_pixmap_t pixmap = xcb_generate_id(c);
+        xcb_dri3_pixmap_from_buffer(c, pixmap, draw, size, width, height,
+                                    stride, depth, bpp, fd);
+        return pixmap;
+    }
+    return 0;
+}
+
+int VA_DRI3_create_fd(Display *dpy, Pixmap pixmap, int *stride)
+{
+    xcb_connection_t *c = XGetXCBConnection(dpy);
+    xcb_dri3_buffer_from_pixmap_cookie_t cookie;
+    xcb_dri3_buffer_from_pixmap_reply_t *reply;
+
+    cookie = xcb_dri3_buffer_from_pixmap(c, pixmap);
+    reply = xcb_dri3_buffer_from_pixmap_reply(c, cookie, NULL);
+    if (!reply) {
+        return -1;
+    }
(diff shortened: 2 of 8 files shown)


jayesht:
@xhaihao @sreerenjb please review this PR.

xhaihao:
Thanks for the patch series. Do you have another PR for the driver? I saw some hook functions are added in struct dri_state.

jayesht:
I do not have corresponding patch set for intel-vaapi-driver. I placed hook functions so that it enables easier driver integration later on.
Though, with this PR vaInitialise() returns success on Fedora 27 on gnome shell.

xhaihao:
vaInitialise() returning success doesn't mean libva+intel-vaapi-driver work well on gnome shell, could you work out a PR for driver as well?

jayesht:
Sure. Will take it up.

va/drm/va_drm.h
```diff
@@ -40,6 +40,23 @@
...
+    { "i915",       4, "i965"   }, // Intel OTC GenX driver
+    { "pvrsrvkm",   8, "pvr"    }, // Intel UMG PVR driver
+    { "emgd",       4, "emgd"   }, // Intel ECG PVR driver
+    { "hybrid",     6, "hybrid" }, // Intel OTC Hybrid driver
+    { "nouveau",    7, "nouveau"  }, // Mesa Gallium driver
+    { "radeon",     6, "r600"     }, // Mesa Gallium driver
+    { "amdgpu",     6, "radeonsi" }, // Mesa Gallium driver
+    { NULL, }
+};
+
```
fhvwy: No - this definitely shouldn't be in a public header.  It probably shouldn't be in a header at all, because the static structure will be duplicated in every file which includes it.

va/x11/va_dri3.c
```diff
@@ -0,0 +1,194 @@
...
+{
+    xcb_connection_t *c = XGetXCBConnection(dpy);
+    xcb_dri3_open_cookie_t cookie;
+    xcb_dri3_open_reply_t *reply;
+
+    if (!VA_DRI3_exists(c))
+        return -1;
+
+    cookie = xcb_dri3_open(c, root, provider);
+    reply = xcb_dri3_open_reply(c, cookie, NULL);
```
fhvwy: Leaks reply.

va/x11/va_dri3.c
```diff
@@ -0,0 +1,194 @@
...
+            if (drm_version->name_len >= m->key_len &&
+                strncmp(drm_version->name, m->key, m->key_len) == 0)
+                break;
+        }
+        drmFreeVersion(drm_version);
+
+        if (!m->name)
+            return 0;
+
+        driverName = strdup(m->name);
```
fhvwy: This leaks.  You probably only want to strdup() once?

va/Makefile.am
```diff
@@ -103,6 +103,10 @@ libva_x11_la_LDFLAGS		= $(LDADD)
 libva_x11_la_DEPENDENCIES	= libva.la x11/libva_x11.la
 libva_x11_la_LIBADD		= libva.la x11/libva_x11.la \
 	$(LIBVA_LIBS) $(X11_LIBS) $(XEXT_LIBS) $(XFIXES_LIBS) $(DRM_LIBS) -ldl
+if USE_DRI3
+libva_x11_la_LIBADD		+= $(X11_XCB_LIBS) $(XCB_LIBS) \
+	$(XCB_DRI3_LIBS) -ldl
```
fhvwy: These variables aren't mentioned anywhere else - I think you want DRI3_LIBS here.

va/x11/va_dri3.c
```diff
@@ -0,0 +1,217 @@
...
+int VA_DRI3_create_fd(Display *dpy, Pixmap pixmap, int *stride)
+{
+    xcb_connection_t *c = XGetXCBConnection(dpy);
+    xcb_dri3_buffer_from_pixmap_cookie_t cookie;
+    xcb_dri3_buffer_from_pixmap_reply_t *reply;
+
+    cookie = xcb_dri3_buffer_from_pixmap(c, pixmap);
+    reply = xcb_dri3_buffer_from_pixmap_reply(c, cookie, NULL);
+    if (!reply) {
+        free(reply);
```
ceyusa: nitpick: if reply is NULL, there's no need to call free

va/x11/va_dri3.c
```diff
@@ -0,0 +1,217 @@
...
+
+    if (!VA_DRI3_exists(c)) {
+        return -1;
+    }
+
+    cookie = xcb_dri3_open(c, root, provider);
+    reply = xcb_dri3_open_reply(c, cookie, NULL);
+
+    if (!reply) {
+        free(reply);
```
ceyusa: ditto

va/x11/va_dri3.c
kwizart: My understanding is that the free(reply) call should be removed because reply can be either NULL or a struct and if !reply means it can be only NULL.

Is there anything preventing this PR (beyond this free(reply) change ?)

alyst:
What is the status of this?

kuboosoft:
Please merge it!

xhaihao:
Sorry for slow response. I'd like to know it works or not before merging it. Did anyone try this patch with https://github.com/intel/intel-vaapi-driver/pull/369?

va/x11/va_dricommon.h
```diff
@@ -83,6 +88,16 @@ struct dri_state
...
+                    int width, int height, int depth,
+                    int fd, int bpp, int stride, int size);
+    int
+    (*create_fence)(VADriverContextP ctx, Pixmap pixmap,
+                         struct dri3_fence *fence);
+    void (*fence_sync)(VADriverContextP ctx, struct dri3_fence *fence);
+    void (*fence_free)(VADriverContextP ctx, struct dri3_fence *fence);
 #endif
     /** \brief Reserved bytes for future use, must be zero */
     unsigned long  va_reserved[16];
```
xhaihao: Please reduce the reserved bytes to keep the size of struct dri_state unchanged.

va/x11/va_x11.c
```diff
@@ -139,7 +153,11 @@ static VAStatus va_DisplayContextGetDriverName (
...
     else
         return VA_STATUS_ERROR_UNKNOWN;
-    
+
+#ifdef HAVE_VA_DRI3
+    vaStatus = va_DRI3GetDriverName(pDisplayContext, driver_name);
+    if (vaStatus != VA_STATUS_SUCCESS)
+#endif
     vaStatus = va_DRI2GetDriverName(pDisplayContext, driver_name);
     if (vaStatus != VA_STATUS_SUCCESS)
```
xhaihao: It will have compatibility issue if the driver supports DRI2 only but both DRI3 and DRI2 can work.

sjnewbury: Shouldn't VA_STATUS_SUCCESS be false if the driver doesn't support DRI3 or does va_DRI3GetDriverName() always return true if the X server supports DRI3 but the driver doesn't?

va/x11/va_dricommon.h
sjnewbury: Is it a bug that the "#ifndef ANDROID" doesn't account for a different resulting size difference when building for Android or is it accepted to have a smaller struct in that case??

chivakker:
@xhaihao what's the latest on this?

uartie:
Original submitter is no longer active and has abandoned this PR.  Someone else will need to take ownership and resubmit in a new PR, please.

PR #187, opened 2018-02-13
va: add new VA STATUS ERROR

ChenXiaomin36:
Add new VA STATUS ERROR to indicate allocated buffer size is
not enough for input or output. For example, when actual bitstream
size for one frame is larger than the size of allocated coded buffer,
then this new VA STATUS ERROR could be returned to indicate it. 

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -27,7 +27,7 @@
 # - reset micro version to zero when minor version is incremented
 # - reset minor version to zero when major version is incremented
 m4_define([va_api_major_version], [1])
-m4_define([va_api_minor_version], [0])
+m4_define([va_api_minor_version], [1])
 m4_define([va_api_micro_version], [0])
 
 m4_define([va_api_version],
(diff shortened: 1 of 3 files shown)


va/va.h
```diff
@@ -210,6 +210,8 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_STATUS_ERROR_HW_BUSY	                0x00000022
 /** \brief An unsupported memory type was supplied. */
 #define VA_STATUS_ERROR_UNSUPPORTED_MEMORY_TYPE 0x00000024
+/** \brief Indicate allocated buffer size is not enough for input or output. */
```
mypopydev: when add a new error type, need to update const char *vaErrorStr(VAStatus error_status) https://github.com/intel/libva/blob/b2fdaf4eb465269656417eb03699ea47d0e532bd/va/va.c#L534 at the same time

PR #188, opened 2018-02-13
add qp adjustment control for HEVC CQP mode

XinfengZhang:

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -791,7 +791,7 @@ typedef struct _VAConfigAttrib {
  *  adjusts QP at a frame and MB level based on motion to improve subjective quality. */
 #define VA_RC_ICQ			0x00000040
 /** \brief Macroblock based rate control.  Per MB control is decided
- *  internally in the encoder. It may be combined with other RC modes, except CQP. */
+ *  internally in the encoder. It may be combined with other RC modes. */
 #define VA_RC_MB                        0x00000080
 /** \brief Constant Frame Size, it is used for small tolerent  */
 #define VA_RC_CFS                       0x00000100


fhvwy:
Can you explain more about exactly what this field is intended to do?  It doesn't look like it should be in `seq_fields` because there is no sequence parameter set field of that name.

XinfengZhang:
@fhvwy sorry for late response. it's CNY. this parameter is only for CQP,  in motion estimation stage, we can get the complexities of the picture, with these data, we can set different QP for different areas  to get better quality and higher compression rate.  it is not exist in HEVC spec. just an enhancement for HEVC CQP

fhvwy:
@XinfengZhang  I'm pretty sure it shouldn't be in the sequence parameter structure, then.  Instead it sounds like it should be a rate control option, though one that applies when bitrate control is not being used (like CRF of libx264).  The quantisation parameter really should be constant when in Constant Quantisation Parameter mode (and H.265 may enforce this because `cu_qp_delta_enabled_flag` need not be set), so maybe you want a new rate control mode (a new `VA_RC_*` value) for this?  Perhaps some sort of constant-quality mode, with this option as one of its parameters?

XinfengZhang:
@fhvwy cu_qp_delta_enabled_flag should just control the delta Qp for the CU inside CTB, right? I am not sure whether it also present the Qp difference of CTB. this feature is for CQP, I wonder whether a new BRC method is suitable

fhvwy:
The change in this form isn't backward compatible, because a v2.1 driver supporting MB RC in VBR mode will now appear to support MB RC in CQP mode when it doesn't.  It probably needs to be a new flag if you do this.

Wrt cu_qp_delta_enabled_flag in H.265, the (luma) QP can't be changed at all inside a slice if it isn't set.  Does your scheme want to modify the QP?  If it does then it definitely can't be used with CQP mode, and a new mode would be required instead.

XinfengZhang:
just like @fhvwy  mentioned, no strong reason for this change

PR #190, opened 2018-02-26
va: add new VA STATUS ERROR

ChenXiaomin36:
Add new VA STATUS ERROR to indicate allocated buffer size is
not enough for input or output.


mypopydev:
LGTM

PR #192, opened 2018-02-27
va: add emulation prevention bytes insertion minimum offset interface

ChenXiaomin36:
The interface indicates the minimum offset to insert emulation prevention
bytes. Bytes before the offset will not include emulation prevention
bytes.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1754,9 +1754,14 @@ typedef struct _VAEncPackedHeaderParameterBuffer {
     uint32_t                bit_length;
     /** \brief Flag: buffer contains start code emulation prevention bytes? */
     uint8_t               has_emulation_bytes;
+    /** \brief Minimum offset for emulation prevention bytes insertion,
+     * bytes before the offset will not include emulation prevention bytes.
+     * Taking effect when has_emulation_bytes is set to zero.
+     */
+    uint32_t                emulation_bytes_insertion_minimum_offset;
 
     /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_LOW];
+    uint32_t                va_reserved[VA_PADDING_LOW - 1];
 } VAEncPackedHeaderParameterBuffer;
 
 /**


va/va.h
```diff
@@ -1754,9 +1754,13 @@ typedef struct _VAEncPackedHeaderParameterBuffer {
     uint32_t                bit_length;
     /** \brief Flag: buffer contains start code emulation prevention bytes? */
     uint8_t               has_emulation_bytes;
+    /** \brief Minimum offset for emulation prevention bytes insertion,
+     * bytes before the offset will not include emulation prevention bytes.
+     */
+    uint32_t                emulation_bytes_insertion_minimum_offset;
```
fhvwy: I'm guessing that for compatibility reasons this value is only used if has_emulation_bytes is zero?  Maybe clarify that.

ChenXiaomin36: We know that if 0x00 0x00 0x01 occurs inside the slice data, encoder should insert 0x03 in front of 0x01. Application packs SPS, PPS, and slice header, and it sets corresponding VAEncPackedHeaderParameterBuffer. If application doesn't insert 0x03 for NALU start codes emulation prevention, it should set has_emulation_bytes to 0. And then driver will take charge of insertion. There are two cases, one is only one NALU (SPS/PPS/slice header) is included in VAEncPackedHeaderParameterBuffer, then emulation prevention byte (0x03) insertion should be after the first three bytes (0x00 0x00 0x01) at least. The other case is that when two NALUs (SVC prefix NALU + slice header) is in one VAEncPackedHeaderParameterBuffer, then emulation prevention byte should be inserted after at least the whole SVC prefix NALU and the second NALU start codes. The new added emulation_bytes_insertion_minimum_offset is used for both cases, which indicates how many bytes should be skipped to insert 0x03.

ChenXiaomin36: update the comment for clarifying when emulation_bytes_insertion_minimum_offset is used.

ChenXiaomin36:
do you have further concern for this pull request? @fhvwy @XinfengZhang

xhaihao:
I just realized that you are using VAEncPackedHeaderParameterBuffer to pass two different NALUs in the same buffer in your case 2. Per your usage, it requires the middle-ware to handle emulation prevention byte insertion for the first NALU. If so, why the middle-ware not insert  emulation prevention byte for all NALUs when packing the headers in the same way? It needn't hardware to insert emulation prevention byte any more if the middle-ware handles emulation prevention byte insertion. 

If you driver follows the header submitting sequence, you may use Packed Raw data header to pass your SVC prefix NALU. 

>  Packed raw data header can be used by the client to insert a header into the bitstream data buffer at the point it is passed,

For case 1, it is easy for driver to identify the start code.

dmitryermilov:
Imagine the case of SVC prefix NALU + slice header in one buffer. Emulation prevention bytes can't be inserted by VAAPI application in slice headers since the application can't be sure that 0x0 0x0 0x1 sequence won't appear on the boundary between slice header and encoded slice itself. So the application can't take responsibility to insert emulation prevention bytes in slice headers.

xhaihao:
I see @dmitryermilov, but why not submit packed buffer for each nalu one by one?

dmitryermilov:
Yes, I agree. E.g. MSDK submits SVC prefix NALU and packed slice header in two separate buffer. As far as I understand this API change is for applications which use the same buffer due to some reasons.

xhaihao:
@dmitryermilov @ChenXiaomin36 Actually case 2 is not expected when adding interface for packed header, a buffer is expected to pass a packed header to driver. e.g.  the comment for packed raw data header mentions inserting **a header** into the bitstream.

xhaihao:
@dmitryermilov @ChenXiaomin36 is there any other comment? I will close this PR if no more comment.

ChenXiaomin36:
No further comment. Thanks.

PR #196, opened 2018-03-07
va: Add VA_SURFACE_ATTRIB_USAGE_HINT_EXPORT

pkerling:
This can be used to tell the driver to prepare the surface such
that it can easily be used for export to third-party APIs.
For example, the driver might choose to decode to frames instead
of fields since only frames can be exported.

AMD driver could make use of this so it doesn't have to weave for export.
See https://bugs.freedesktop.org/show_bug.cgi?id=105145 for more motivational background

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1321,6 +1321,9 @@ typedef struct _VASurfaceAttribExternalBuffers {
 #define VA_SURFACE_ATTRIB_USAGE_HINT_VPP_WRITE 	0x00000008
 /** \brief Surface used for display. */
 #define VA_SURFACE_ATTRIB_USAGE_HINT_DISPLAY 	0x00000010
+/** \brief Surface used for export to third-party APIs, e.g. via
+ *  vaExportSurfaceHandle(). */
+#define VA_SURFACE_ATTRIB_USAGE_HINT_EXPORT 	0x00000020
 
 /**@}*/
 


xhaihao:
Just for the curious, does your application work with intel driver if this flag is set in your application? As it is a hint only, so the driver may ignore the user's setting.

fhvwy:
(That attribute has only ever been used by the iHD driver.)

I would be vary of adding this sort of thing as a requirement, because it will fail surprisingly on one driver only if the hint hasn't been set.  IMO this seems like another argument for making everything inside Mesa progressive by default and only converting surfaces to the separated-fields format when needed by the deinterlacer (since everything else can work with progressive, indeed most things (encoder, export, H.265 and JPEG decoders) all require it anyway so the separated-fields are mostly huge pain).  I pushed for this a bit when adding vaExportSurfaceHandle(), but with Christian's suggestion of the magical surface format change I didn't have any actual failing case because of it and therefore gave up.  Maybe you want to look at it again now?

pkerling:
> Just for the curious, does your application work with intel driver if this flag is set in your application? As it is a hint only, so the driver may ignore the user's setting.

Yes, as @fhvwy said neither mesa nor intel-vaapi-driver use any of the surface usage hints at present. 

> I would be vary of adding this sort of thing as a requirement, because it will fail surprisingly on one driver only if the hint hasn't been set.

This is a performance optimization to avoid requiring the driver to weave the first few fields until the decoder hits the surfaces again that are then marked as progressive. Everything should work fine without, so I think the usage hint is appropriate.

>  IMO this seems like another argument for making everything inside Mesa progressive by default and only converting surfaces to the separated-fields format when needed by the deinterlacer (since everything else can work with progressive, indeed most things (encoder, export, H.265 and JPEG decoders) all require it anyway so the separated-fields are mostly huge pain).

You could do that and it does sound like a good idea to me, but then you would at least need to be able to unweave already decoded frames when kicking in the deinterlacer. And again use a hint (like `VA_SURFACE_ATTRIB_USAGE_HINT_VPP_READ`?) to avoid having to unweave when you know from the start that you want to deinterlace.

In order to solve the problem fully and get rid of the workaround I'm using (always doing a vaapi postproc pass even when not deinterlacing), mesa would also have to retain the functionality to weave the frames when switching off deinterlacing dynamically.

So basically the (or rather, my) problem stays the same, but yes we wouldn't need this hint :-) Or maybe no hint at all, since interlaced video is mostly 1080i and below and so the performance impact of having to unweave a few 1080i frames is smaller than having to weave a few 2160p frames.

PR #202, opened 2018-03-28
Add 'fast_intra_mode' in VAEncMiscParameterFEIFrameControlHEVC to do fast intra prediction

dlin2:
Add 'fast_intra_mode' in VAEncMiscParameterFEIFrameControlHEVC to do fast intra prediction

Signed-off-by: dlin2 <decai.lin@intel.com>

### DIFF
--- a/va/va_fei_hevc.h
+++ b/va/va_fei_hevc.h
@@ -165,7 +165,10 @@ typedef struct _VAEncMiscParameterFEIFrameControlHEVC
     uint32_t      search_window             : 8;
     /** \brief number of internal MV predictors for IME searches */
     uint32_t      max_num_ime_search_center : 3;
-    uint32_t      reserved2                 : 5;
+    /** \brief fast intra prediction enabling bit. It is used as a trade-off between speed and quality.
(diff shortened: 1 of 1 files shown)


va/va_fei_hevc.h
```diff
@@ -165,7 +165,9 @@ typedef struct _VAEncMiscParameterFEIFrameControlHEVC
     uint32_t      search_window             : 8;
     /** \brief number of internal MV predictors for IME searches */
     uint32_t      max_num_ime_search_center : 3;
-    uint32_t      reserved2                 : 5;
+    /** \brief fast intra prediction enabling bit */
+    uint32_t      fast_intra_mode           : 1;
```
xhaihao: How is an application or middleware aware of this capability of the underlying driver? What's the result if application set this flag but the driver doesn't support fast intra mode?

dlin2: There will be the corresponding interface in MSDK API, and also have detail descriptions in MSDK doc. Driver will ignore the flag and the result will be no change if setting the flag in the unsupported driver.
I will update the code comments for more detailed descriptions.

PR #206, opened 2018-04-26
Misc changes around formats and documentation

fhvwy:
Fixes some stuff missed in d6fd111e, then general cleanup and documentation.

The fourcc documentation patch in particular wants to be checked carefully to make sure I haven't made any mistakes in the format definitions.


va/va.h
```diff
@@ -1836,7 +1836,8 @@ typedef struct _VAEncMiscParameterRateControl
     /* initial QP at I frames */
     uint32_t initial_qp;
     uint32_t min_qp;
-    uint32_t basic_unit_size;
+    /** @deprecated This value is unused. */
+    va_deprecated uint32_t basic_unit_size;
```
xhaihao: basic_unit_size was used in https://github.com/freedesktop/pvr-driver (although no one is using this driver now), maybe other HWs can use it.

fhvwy: Can you explain what it does so it can be documented, then?  From what I can get from the source code of that driver it appears to be a measurement in blocks (H.264 -> macroblocks) which gets passed into a binary blob to do something unclear.

(That driver doesn't even compile with anything recent - it depends on an early 0.something version VAAPI from before any API compatibility was maintained.)

xhaihao: Actually I don't know what it does, I just guess some HWs support basic unit size based rate control so basic_unit_size is required for these HWs.

fhvwy: Ok: I've dropped this commit and modified the following one which updates the documentation appropriately.

PR #207, opened 2018-05-02
add histogram output support for decode

XinfengZhang:

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -748,6 +748,17 @@ typedef enum
      * support for QP info for buffer #VAEncQpBuffer.
      */
     VAConfigAttribQPBlockSize            = 37,
+    /**
+    * \brief Histogram reporting attribute. Read-only.
+    *
+    * This attribute conveys whether the driver supports reporting of
+    * decoded frame histogram, /c VAConfigAttribHistogram is used to specify supported
+    * hisogram information, zero means no histogram support.
+    * If it is supported, use /c VAHistogramBufferType,
+    * to get the histogram, use /c VAEncMiscParameterTypeHistogram /c VAEncMiscParameterHistorgram
+    * to set histogram parameters.
+    */
+    VAConfigAttribHistogram             = 38,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -959,6 +974,25 @@ typedef union _VAConfigAttribValEncJPEG {
...
+/** \brief Specifying histogram channels */
+#define VA_HISTOGRAM_NONE                             0x00000000
+/** \brief Driver supports channel Y histogram report  */
+#define VA_HISTOGRAM_CHANNEL_Y                                0x00000001
+/** \brief Driver supports channel U histogram report  */
+#define VA_HISTOGRAM_CHANNEL_U                                0x00000002
+/** \brief Driver supports channel V histogram report  */
+#define VA_HISTOGRAM_CHANNEL_V                                0x00000004
+/** \brief Driver supports channel R histogram report  */
+#define VA_HISTOGRAM_CHANNEL_R                                0x00000008
```
cwang105: for R, G, B, do we need to give distinct values for Y, U, V?

XinfengZhang: this is for the caps reporting, we can t report 0x7 only, customer dont know whether we support  RGB or YUV from value 0x7

va/va.h
```diff
@@ -748,6 +748,26 @@ typedef enum
...
+    * If histogram reporting is not supported, driver should return
+    * VA_HISTOGRAM_CHENNAL_NONE.
+    */
+    VAConfigAttribHistogramChannels      = 38,
+    /**
+     * \brief Histogram Bin Count reporting attribute. Read-only.
+     * This attribute conveys the number of bins per channel if the
+     * driver supports histogram reporting.
+     */
+    VAConfigAttribHistogramBinCount      = 39,
```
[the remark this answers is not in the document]
XinfengZhang: yes

va/va.h
```diff
@@ -980,6 +991,31 @@ typedef union _VAConfigAttribValEncJPEG {
...
+    struct {
+        /** \brief bin count for given channel. */
+        uint32_t bin_count : 16;
+        /** \brief driver support channel */
+        uint32_t channel   : 8;
+        /** \brief frequence value size */
+        uint32_t reserved  : 8;
+    } bits;
+    uint32_t value;
+} VAConfigAttribHistogramBin;
```
xhaihao: It would be better to keep the same name style, _VAConfigAttribHistogram and VAConfigAttribHistogram, or _VAConfigAttribHistogramBin and VAConfigAttribHistogramBin

va/va.h
```diff
@@ -980,6 +991,31 @@ typedef union _VAConfigAttribValEncJPEG {
...
+    VAHistogramChannelA
+}VAHistogramChannelType;
+
+/** \brief Attribute value for VAConfigAttribHistogram */
+typedef union _VAConfigAttribHistogram{
+    struct {
+        /** \brief bin count for given channel. */
+        uint32_t bin_count : 16;
+        /** \brief driver support channel */
+        uint32_t channel   : 8;
```
xhaihao: does it support multiple channels?

XinfengZhang: yes, for the caps report, will return several VAConfigAttribHistograms , also application can send multiple VAEncMiscParameterHistorgram

XinfengZhang:
@xhaihao  do you still have other concern? if no , I will merge it

PR #208, opened 2018-05-08
Support content protection.

AppleWang001:
Add interfaces to support content protection for secure decode/encode/vpp

Signed-off-by: AppleWang001 <apple.wang@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -129,6 +129,7 @@ extern "C" {
  * 	- \ref api_fei
  * 	- \ref api_fei_h264
  * 	- \ref api_fei_hevc
+ * - \ref api_cp
  */
 
 /**
@@ -1630,6 +1631,10 @@ typedef enum
     /** decode stream out buffer, intermedia data of decode, it may include MV, MB mode etc.
       * it can be used to detect motion and analyze the frame contain  */
     VADecodeStreamoutBufferType             = 56,
+    /**\brief encryption parameters buffer for content protection usage*/
+    VAEncryptionParameterBufferType = 57,
+    /**\brief CENC status paramter*/
+    VACencStatusParameterBufferType = 58,
     VABufferTypeMax
 } VABufferType;
 
@@ -3172,6 +3177,22 @@ VAStatus vaBufferSetNumElements (
 #define VA_CODED_BUF_STATUS_SINGLE_NALU                 0x10000000	
 
 /**
+ * \brief The coded buffer segment contains valid IV. 
+ *
+ * This flag indicates that the coded buffer segment contains
+ * valid enryption_iv[]. 
+ */
+#define VA_CODED_BUF_STATUS_VALID_IV               0x20000000	
+
+/**
+ * \brief The coded buffer segment contains valid IV. 
+ *
+ * This flag indicates that the coded buffer segment contains
+ * valid enryption_iv[]. 
+ */
+#define VA_CODED_BUF_STATUS_TEAR_DOWN              0x40000000	
+
+/**
  * \brief Coded buffer segment.
  *
  * #VACodedBufferSegment is an element of a linked list describing
@@ -3193,15 +3214,17 @@ typedef  struct _VACodedBufferSegment  {
     /** \brief Reserved for future use. */
     uint32_t        reserved;
     /** \brief Pointer to the start of the data buffer. */
-    void               *buf;
+    void            *buf;
     /**
      * \brief Pointer to the next #VACodedBufferSegment element,
      * or \c NULL if there is none.
      */
-    void               *next;
-
-    /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_LOW];
+    void            *next;
+    /** 
+      *\brief IV or Counter for encrytion filled by driver.
+      * Only available when status is set with \c VA_CODED_BUF_STATUS_VALID_IV.
+      */ 
+    uint32_t        encryption_iv[4];
 } VACodedBufferSegment;
      
 /**
@@ -4209,6 +4232,7 @@ typedef struct _VAPictureHEVC
 #include <va/va_fei.h>
 #include <va/va_fei_h264.h>
 #include <va/va_vpp.h>
+#include <va/va_cp.h>
 
 /**@}*/
 
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -197,8 +197,8 @@ struct VADriverVTable
 	VAStatus (*vaQuerySurfaceError) (
 		VADriverContextP ctx,
 		VASurfaceID render_target,
-                VAStatus error_status,
-                void **error_info /*out*/
+		VAStatus error_status,
+		void **error_info /*out*/
 	);
 
 	VAStatus (*vaPutSurface) (
@@ -486,8 +486,19 @@ struct VADriverVTable
             void               *descriptor      /* out */
         );
 
+        VAStatus (*vaQueryCenc) (
+            VADriverContextP ctx,
+            VABufferID buffer,
+            VACencStatusBuf *info/* out */
+        );
+
+        VAStatus (*vaEndCenc) (
+            VADriverContextP ctx,
+            VAContextID context
+        );
+
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[57];
+        unsigned long reserved[55];
 };
 
 struct VADriverContext
--- a/va/va_cp.h
+++ b/va/va_cp.h
@@ -1,0 +1,420 @@
+/*
+ * Copyright (c) 2018 Intel Corporation. All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sub license, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ * 
+ * The above copyright notice and this permission notice (including the
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL INTEL AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+* \file va_cp.h
+* \brief The content protection API
+*
+* This file contains the \ref api_cp "Content Protection API".
+*/
+
+#ifndef VA_CP_H
+#define VA_CP_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * \defgroup api_cp Content Protection API
+ *
+ * @{
+ *
+ * The content protection API uses the same paradigm for codec.
+ * - Query for supported encryption attributes through VAConfigAttribEncryption.
+ * - create the specific configuration with the accepted encryption attributes got from the query.
+ * - Set up a codec pipeline;
+ 
+ * The basic operation steps for cenc decode are:
+ * - Query for supported encryption attributes through VAConfigAttribEncryption.
+ * - Create configuration with one of encryption attributes got from the above step.
+ * - Create a video decode.
+ * - Pass the render buffers with the corresponding data to CENC decode, including:
+ *   encrypted bitstream through VAProtectedSliceDataBufferType;
+ *   encryption paramters through VAEncryptionParameters.
+ * - Make the end of CENC call via vaEndCenc
+ * - Get CENC result via vaQueryCenc.
+ * - Pass the render buffers with the corresponding data to decode, including:
+ *   CENC status report number through VACencStatusParameterBufferType.
+ *   DPB information through VAPictureParameterBuffer.
+ * - Make the end of decode call via vaEndPicture
+ */
+
+/** attribute values for VAConfigAttribEncryption */
+#define VA_ENCRYPTION_TYPE_NONE             0x00000000
+#define VA_ENCRYPTION_TYPE_BASIC            0x00000001 
+#define VA_ENCRYPTION_TYPE_CENC_CBC         0x00000002
+#define VA_ENCRYPTION_TYPE_CENC_CTR_LENGTH  0x00000004
+#define VA_ENCRYPTION_TYPE_CENC_CTR         0x00000008
+#define VA_ENCRYPTION_TYPE_CTR_128          0x00000010
+
+/** \brief values for the encryption return status. */
+typedef enum
+{
+    /** \brief Indicate encryption operation is successful.*/ 
+    VA_ENCRYPTION_STATUS_SUCCESSFUL  = 0,
+    /** \brief Indicate encryption operation is incomplete.*/ 
+    VA_ENCRYPTION_STATUS_INCOMPLETE,
+    /** \brief Indicate encryption operation is error.*/ 
+    VA_ENCRYPTION_STATUS_ERROR,
+    /** \brief Indicate encryption result isn't available.*/ 
+    VA_ENCRYPTION_STATUS_UNAVAILABLE
+} VAEncryptionStatus;
+
+/** \brief structure for encrypted segment info. */
+typedef struct _VAEncryptionSegmentInfo
+{
+    /** \brief  The offset relative to the start of the bitstream input of
+     *  the start of the segment*/
+    uint32_t segment_start_offset;
+    /** \brief  The length of the segments in bytes*/
+    uint32_t segment_length;
+    /** \brief  The length in bytes of the remainder of an incomplete block
+     *  from a previous segment*/
+    uint32_t partial_aes_block_size;
+    /** \brief  The length in bytes of the initial clear data */
+    uint32_t init_byte_length; 
+    /** \brief  This will be AES 128 counter for secure decode and secure
+     *  encode when numSegments equals 1 */
+    uint32_t aes_cbc_iv_or_ctr[4];
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t va_reserved[VA_PADDING_MEDIUM];
+} VAEncryptionSegmentInfo;
+
+/** \brief encrytpion parameters*/
+typedef struct _VAEncryptionParameters
+{
+    /** \brief Encryption type, attribute values. */
+    uint32_t encryption_type;
+    /** \brief Hw session id */
+    uint32_t session_id;
+    /** \brief The number of sengments */
+    uint32_t num_segments;
+    /** \brief Pointer of segments */
+    VAEncryptionSegmentInfo *segment_info;
+    /** \brief The status report number for CENC workload */ 
+    uint16_t  status_report_number;
+    /** \brief CENC counter length */
+    uint16_t size_of_length;
+    /** \brief Reserved bytes for future use, must be zero */
+    unsigned long va_reserved[VA_PADDING_MEDIUM];
+} VAEncryptionParameters;
+
+/**
+ * \brief Slice parameter for H.264 cenc decode in baseline, main & high profiles.
+ *
+ * This structure holds information for \c
+ * slice_layer_without_partitioning_rbsp() and nal_unit()of the slice 
+ * as defined by the H.264 specification.
+ *
+ */
+typedef struct _VACencSliceParameterBufferH264 {
+
+    /** \brief Parameters from \c nal_unit() of the slice.*/
+    /**@{*/
+    /** \brief  Same as the H.264 bitstream syntax element. */
+    uint8_t nal_ref_idc;
+    /** \brief Indicate if this is coded slice of an IDR picture. 
+     * Corresponds to IdrPicFlag as the H.264 specification.*/
+    uint8_t idr_pic_flag;
+    /**@}*/
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t slice_type;
+    /** \brief Indicate if this is a field or frame picture.
+     * \c VA_FRAME_PICTURE, \c VA_TOP_FIELD, \c VA_BOTTOM_FIELD*/
+    uint8_t field_frame_flag;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint32_t frame_number;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint32_t idr_pic_id;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t pic_order_cnt_lsb;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t delta_pic_order_cnt_bottom;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t delta_pic_order_cnt[2];
+    /**
+     * \brief decoded reference picture marking. Information for \c 
+     * dec_ref_pic_marking() as defined by the H.264 specification.
+     */
+    /**@{*/
+    union {
+        struct {
+            /** \brief Same as the H.264 bitstream syntax element. */
+            uint32_t no_output_of_prior_pics_flag             : 1;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            uint32_t long_term_reference_flag                 : 1;
+            /** \brief Same as the H.264 bitstream syntax element. */
+            uint32_t adaptive_ref_pic_marking_mode_flag       : 1;
+            /** \brief number of decode reference picture marking. */
+            uint32_t dec_ref_pic_marking_count                : 8;;
+            /** \brief Reserved for future use, must be zero */
+            uint32_t reserved                                 : 21;
+            } bits;
+        uint32_t value;
+    } ref_pic_fields;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t memory_management_control_operation[32];
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t difference_of_pic_nums_minus1[32];
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t long_term_pic_num[32];
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t max_long_term_frame_idx_plus1[32];
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t long_term_frame_idx[32];
+    /**@}*/
+    /** \brief Pointer to the next #VACencSliceParameterBufferH264 element,
+    * or \c NULL if there is none.*/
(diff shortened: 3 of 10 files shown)


va/va_backend.h
```diff
@@ -197,8 +197,14 @@ struct VADriverVTable
 	VAStatus (*vaQuerySurfaceError) (
 		VADriverContextP ctx,
 		VASurfaceID render_target,
-                VAStatus error_status,
-                void **error_info /*out*/
+        VAStatus error_status,
+        void **error_info /*out*/
+	);
```
ceyusa: indentation change on unrelated code.

va/va_backend.h
```diff
@@ -197,8 +197,14 @@ struct VADriverVTable
...
+
+	VAStatus (*vaQueryCenc) (
+        VADriverContextP ctx,
+        VABufferID buffer,
+        VACencStatusBuf *info/*in, out*/
```
ceyusa: Also, I guess, here the indentation is wrong.

va/va_cp.h
```diff
@@ -0,0 +1,160 @@
+/*
```
ceyusa: shouldn't be, this header file, included in va.h ??

fhvwy:
(I assume the intent of this is to block the user from being able to extract any of the pixel data, including when it is compressed in the codec bitstream.  If that's not the case then please do correct me.)

I'm concerned that this is very inflexible.  What is the reasoning for requiring the driver to parse headers and return a specific fixed subset of the information found therein?  Would it not be cleaner to instead return the decrypted headers from the bitstream (without the actual data), and allow the user to parse them appropriately?

This would have the following advantages:
* It would work without any change with new codec profiles (e.g. it would work with the recently-added H.265 RExt profiles, which this definition does not support).
* More metadata would be visible to the user.  The definition above only extracts buffering period and picture timing SEI for H.265 - what about, say, mastering display / content light level data (required for HDR support)?
* The change would be much less invasive - it would reuse most existing paths in VAAPI decoder clients.
* It wouldn't require much change to work on other codecs - the only requirement on the secure component is to be able to find where the header ends and the block layer begins for each slice.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    * Same as the H.264 bitstream syntax element.*/
+    uint8_t idr_flag;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint32_t frame_number;
+    /** \brief Indicate if this is a field or frame picture. */
+    uint8_t field_frame_flag;
+    /** \brief top field or frame POC. */
+    int32_t top_field_order_cnt;
+     /** \brief bottom field POC. */
+    int32_t bottom_field_order_cnt;
```
fhvwy: How are you intending that a driver determines the full POC values?  It probably shouldn't be keeping the internal state necessary to do this.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t va_reserved[VA_PADDING_MEDIUM];
+} VACencSliceParameterBufferH264;
```
fhvwy: Quite a bit of stuff is missing here.  How does the user determine which reference pictures are needed without the ref pic list information?

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    int32_t initial_cpb_removal_delay_offset_vcl[32];
+    /**@}*/
+    /** @name sei picture timing*/    
+    /**@{*/    
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t cpb_removal_delay;    
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t dpb_output_delay;    
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t pic_struct;
```
fhvwy: Missing timestamp fields?

va/va_backend.h
AppleWang001: Accept. Mixed by tab and space.

va/va_backend.h
AppleWang001: Accept

va/va_cp.h
AppleWang001: This header is included in va.h @line4231

va/va.h
```diff
@@ -3190,18 +3199,17 @@ typedef  struct _VACodedBufferSegment  {
     uint32_t        bit_offset;
     /** \brief Status set by the driver. See \c VA_CODED_BUF_STATUS_*. */
     uint32_t        status;
-    /** \brief Reserved for future use. */
-    uint32_t        reserved;
+    /** \brief Indicate if aes_cbc_iv_or_ctr[] is valid. */
+    uint32_t        aes_info_valid;
```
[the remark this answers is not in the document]
AppleWang001: Accepted. Will add a new flag field.

va/va.h
```diff
@@ -3190,18 +3199,17 @@ typedef  struct _VACodedBufferSegment  {
...
      * \brief Pointer to the next #VACodedBufferSegment element,
      * or \c NULL if there is none.
      */
-    void               *next;
-
-    /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_LOW];
+    void            *next;
+    /** \brief aes infor, IV for CBC mode, counter for others*/
+    uint32_t        aes_cbc_iv_or_ctr[4];
```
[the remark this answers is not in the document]
AppleWang001: Accepted.

va/va_backend.h
```diff
@@ -197,8 +197,14 @@ struct VADriverVTable
...
 	);
```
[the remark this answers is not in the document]
AppleWang001: Thanks for noticing this.

va/va_cp.h
```diff
@@ -0,0 +1,160 @@
...
+ * - Pass the render buffers with the corresponding data to CENC decode, including:
+ *   encrypted bitstream through VAProtectedSliceDataBufferType;
+ *   encryption paramters through VAEncryptionParameters.
+ * - Make the end of CENC call via vaEndPicture
+ * - Get CENC result via vaQueryCenc.
+ * - Pass the render buffers with the corresponding data to decode, including:
+ *   CENC status report number through VACencStatusParameterBufferType.
+ *   DPB information through VAPictureParameterBuffer.
+ * - Make the end of decode call via vaEndPicture
+
```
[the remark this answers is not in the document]
AppleWang001: Accepted.

va/va_cp.h
```diff
@@ -0,0 +1,160 @@
...
+#define VA_ENCRYPTION_TYPE_CTR_128          0x00000010
+
+/** \brief values for the encryption return status. */
+typedef enum
+{
+    VA_ENCRYPTION_STATUS_SUCCESSFUL  = 0,
+    VA_ENCRYPTION_STATUS_INCOMPLETE,
+    VA_ENCRYPTION_STATUS_ERROR,
+    VA_ENCRYPTION_STATUS_TEARDOWN,
+    VA_ENCRYPTION_STATUS_UNAVAILABLE
```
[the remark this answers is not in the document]
AppleWang001: Accepted.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+
+/**
+ * \brief Scaling list parameter for H.264 cenc decode
+ *
+ * This structure holds information for \c scaling_list()
+ * as defined by the H.264 specification. 
+ */
+typedef struct _VACencScalingListBufferH264 {
+
+    /** \brief Conrresponds to ScalingList4x4 as the H.264 specification */
```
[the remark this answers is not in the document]
AppleWang001: Accepted.

AppleWang001:
@fhvwy 
The reason why driver parses headers are:
1. Driver parsing is to make sure the bitstream is correct. If bitstream is something wrong or corrupted, driver will detect and report an error.
2. Driver parsing is make sure the decryption is correct. Sometimes key or cp environment issue will cause decrypted bitstream incorrect. Back to step 1 to report an error.
3. if driver doesn't do the header parsing and return the clear header of bitstream, it's possible that someone might fake encrypted bitstream into the header or SEI message to potentially leak content.
4. Since driver should do the header parsing (due to 1/2/3 above), return decrypted header of bitstream will lead a duplicated header parsing from application.
5. Return decrypted header of bitstream might reduce the complexity to crack.  

To increase the flexibility, one idea is to use separate buffers for different parameters. Every parameter has it's buffer type and parameter buffer. Applications can choose  which parameters they wants and provide the buffers with the related buffer type.  Driver should return success status  together with the specified parameter, or invalid status if it can't provide the data. If a new parameter is added, such as a new SEI payload, it's no need to change interfaces and just adding a buffer type and a parameter buffer is fine. What's your thought?

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+        /** \brief Same as the H.264 bitstream syntax element*/
+    uint8_t pic_parameter_set_id;
+    /** \brief Same as the H.264 bitstream syntax element.*/
+    uint8_t seq_parameter_set_id;
+    /** \brief Same as the H.264 bitstream syntax element.*/
+    uint8_t num_slice_groups_minus1;
+     /** \brief Corresponds to the H.264 bitstream syntax element. 
+     *  num_ref_idx_l0_default_active_minus1 + 1 
+     */
+    uint8_t num_ref_idx_l0_active;
```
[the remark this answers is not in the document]
AppleWang001: Will change to num_ref_idx_l0_default_active

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    uint32_t sar_height;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t video_format;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t colour_primaries;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t transfer_characteristics;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t matrix_coefficients;
+    /** \brief Same as the H.264 bitstream syntax element. */
```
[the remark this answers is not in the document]
AppleWang001: Chroma sample location information is for interlaced 4:2:0 format, right? Since most of premium contents are progressive(We didn't see the interlaced premium contents), can chroma sample location be excluded from vui?

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    uint32_t num_units_in_tick;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint32_t time_scale;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint32_t num_reorder_frames;
```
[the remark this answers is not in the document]
AppleWang001: The spec (201003) shows "num_reorder_frames" in vui bitstream restriction information. Let me find a newer one to confirm.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    /** \brief Same as the H.264 bitstream syntax element. 
+     *  value range [-12..12] 
+     */
+    int8_t chroma_qp_index_offset;
+    /** \brief Same as the H.264 bitstream syntax element. 
+     *  value range [-12..12] 
+     */
+    int8_t second_chroma_qp_index_offset;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t pic_scaling_list_present_flag[8];
```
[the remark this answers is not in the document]
AppleWang001: Yes, you are right. Thanks.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t offset_for_non_ref_pic;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t offset_for_top_to_bottom_field;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t num_ref_frames_in_pic_order_cnt_cycle;    
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t offset_for_ref_frame[256];
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t seq_scaling_list_present_flag[8];
```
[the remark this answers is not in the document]
AppleWang001: Yes, you are right. Thanks.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    uint16_t pic_height_in_map_units_minus1;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t bit_depth_luma_minus8;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t bit_depth_chroma_minus8;    
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t offset_for_non_ref_pic;
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t offset_for_top_to_bottom_field;
+    /** \brief Same as the H.264 bitstream syntax element. */
```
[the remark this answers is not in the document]
AppleWang001: Yes, you are right. Thanks a lot.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+    uint8_t num_ref_frames_in_pic_order_cnt_cycle;    
+    /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t offset_for_ref_frame[256];
```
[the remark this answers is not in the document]
AppleWang001: Yes, you are right. Thanks a lot.

va/va_cp_h264.h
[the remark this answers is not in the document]
fhvwy: No, it's used for all 4:2:0 cases.  (When the stream is progressive the top and bottom field values are required to be equal.)

va/va_cp_h264.h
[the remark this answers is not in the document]
fhvwy: Right, the field was renamed in the 2011/06 version of the spec.  I think it's better to use the current name.

fhvwy:
@AppleWang001 
I agree that the protected code needs to parse the bitstream to prevent leaks, but that should not rule out returning the unencrypted headers anyway.

What you are wanting to do here is to return the headers in some serialised form, and I suggest that the relevant part of the unencrypted bitstream is the most convenient serialised form to use.  That is primarily because all VAAPI decoder clients have already implemented everything necessary to handle it - another serialised form such as the structures you define here requires a lot more code in every client which wants to handle it.

Using the unencrypted headers as the serialised form also avoids needing special code in the driver to handle all of the metadata which isn't relevant to the driver at all (pretty much all SEI), including anything defined in future which isn't yet known.

va/va_cp_vp9.h
```diff
@@ -0,0 +1,121 @@
...
+typedef struct _VACencHeaderParameterBufferVP9 {
+    
+    union {
+        struct {
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t profile                              : 2;            
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t show_existing_frame_flag             : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t show_to_show                         : 3;
```
fhvwy: Given the position, this is maybe meant to be `frame_to_show_map_idx`?

va/va_cp_vp9.h
```diff
@@ -0,0 +1,121 @@
...
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t error_resilient_mode                 : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t color_space                          : 2; 
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t subsampling_x                        : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t subsampling_y                        : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t refresh_frame_flags                  : 1;
```
fhvwy: `refresh_frame_flags` is a mask of the reference frame slots to place the current frame in - it should be a separate uint8_t variable.

va/va_cp_vp9.h
```diff
@@ -0,0 +1,121 @@
...
+        uint32_t value;
+    } header_fields; 
+     /** \brief Same as the VP9 bitstream syntax element. */
+    uint16_t frame_width_minus1;
+     /** \brief Same as the VP9 bitstream syntax element. */
+    uint16_t frame_height_minus1;
+    /** \brief Same as the VP9 bitstream syntax element. */
+    uint16_t render_width_minus1;
+    /** \brief Same as the VP9 bitstream syntax element. */
+    uint16_t render_height_minus1;
```
fhvwy: This needs at least the `ref_frame_idx` array and the `intra_only` flag as well I think.

va/va_cp_vp9.h
```diff
@@ -0,0 +1,121 @@
...
+ * of data in VACencStatusBuf
+ *
+ */
+typedef struct _CencStatusReportVp9 {
+    /* \brief encryption status defined by VAEncryptionStatus */
+    uint32_t                       status;
+    /* \brief status report feed back number */
+    uint16_t                       status_report_feedback_number;
+    /** \brief header parameter*/
+    VACencHeaderParameterBufferVP9 header_info;
```
fhvwy: Is it defined that encrypted VP9 streams must always have the superframe index in the clear (so the user decomposes the stream into frames and passes them separately), or must not be encapsulated in superframes?

va/va_cp_vp9.h
```diff
@@ -0,0 +1,121 @@
...
+    uint32_t                       va_reserved[VA_PADDING_MEDIUM];
+}CencStatusReportVp9;
+
+/**@}*/
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* _MEDIA_DDI_CENC_DECODE_VP9_H */
```
fhvwy: Stray comment.

va/va_cp_vp9.h
```diff
@@ -0,0 +1,121 @@
...
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t show_to_show                         : 3;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t frame_type                           : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t show_frame                           : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t error_resilient_mode                 : 1;
+            /** \brief Same as the VP9 bitstream syntax element. */
+            uint32_t color_space                          : 2;
```
fhvwy: `color_space` is three bits.

`color_range` and `ten_or_twelve_bit` should probably be included as well.

AppleWang001:
@fhvwy 
Thanks for the suggestion. We are analyzing the possibility of leaking slice data if returning unencrypted headers.

va/va_cp_vp9.h
AppleWang001: Accepted

va/va_cp_vp9.h
AppleWang001: Accepted

va/va_cp_vp9.h
AppleWang001: When encrypting superframes,  the the superframe header must be clear. Defined in VP Codec ISO Media File Format Binding.

va/va_cp_vp9.h
AppleWang001: Accepted.

va/va_cp_vp9.h
AppleWang001: Accepted.

va/va_cp_vp9.h
AppleWang001: Accepted.

va/va_cp_h264.h
AppleWang001: Remove this struct as SEI will be returned as bitstream.

va/va_cp_h264.h
AppleWang001: ref_pic_marking is returned so that DPB can be handled by user.

va/va_cp_h264.h
AppleWang001: Since driver does the parsing, driver can handle POC calculation. The original idea is to let application does as less as possible. It's fine that application does POC calculation. Will change the struct so that application can handle POC calculation.

va/va_cp_h264.h
```diff
@@ -0,0 +1,481 @@
...
+
+     /** \brief Slice type.
+     *  Range: 0..2, 5..7, i.e. no switching slices.
+     */
+    uint8_t slice_type;
+     /** \brief Same as the H.264 bitstream syntax element. */
+    uint8_t nal_ref_idc;
+    /** \brief Indicate if this is coded slice of an IDR picture. 
+    * Same as the H.264 bitstream syntax element.*/
+    uint8_t idr_flag;
```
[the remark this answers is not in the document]
AppleWang001: Accepted

va/va_cp_h264.h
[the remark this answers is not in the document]
AppleWang001: This struct is removed due to SPS is returned as bitstream.

AppleWang001:
@fhvwy The experiment showed returning slice header as bitstream had the leak potential. That to say, slice data might be stolen if slice data is faked as slice header. SPS/PPS/SEI(some payloads) are fine because Specs have value limitation. Proposed a new struct for VACencStatusBuf to return SPS/PPS/SEI as bistream, and returning slice header as struct or bitstream can be chosen as driver.

va/va.h
```diff
@@ -3171,6 +3176,22 @@ VAStatus vaBufferSetNumElements (
...
+ */
+#define VA_CODED_BUF_STATUS_VALID_IV               0x20000000	
+
+/**
+ * \brief The coded buffer segment contains valid IV. 
+ *
+ * This flag indicates that the coded buffer segment contains
+ * valid enryption_iv[]. 
+ */
+#define VA_CODED_BUF_STATUS_TEAR_DOWN              0x40000000
```
fhvwy: Not the right comment?

va/va_backend.h
```diff
@@ -197,8 +197,8 @@ struct VADriverVTable
 	VAStatus (*vaQuerySurfaceError) (
 		VADriverContextP ctx,
 		VASurfaceID render_target,
-                VAStatus error_status,
-                void **error_info /*out*/
+		VAStatus error_status,
+		void **error_info /*out*/
```
fhvwy: Please don't include unrelated changes.

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+typedef enum
+{
+    /** \brief Indicate encryption operation is successful.*/ 
+    VA_ENCRYPTION_STATUS_SUCCESSFUL  = 0,
+    /** \brief Indicate encryption operation is incomplete.*/ 
+    VA_ENCRYPTION_STATUS_INCOMPLETE,
+    /** \brief Indicate encryption operation is error.*/ 
+    VA_ENCRYPTION_STATUS_ERROR,
+    /** \brief Indicate encryption result isn't available.*/ 
+    VA_ENCRYPTION_STATUS_UNAVAILABLE
```
fhvwy: How do the three failure error codes differ?  It might help to suggest how the API user should respond to each one (retry, show error to end-user, ...).

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+} VAEncryptionStatus;
+
+/** \brief structure for encrypted segment info. */
+typedef struct _VAEncryptionSegmentInfo
+{
+    /** \brief  The offset relative to the start of the bitstream input of
+     *  the start of the segment*/
+    uint32_t segment_start_offset;
```
fhvwy: I guess this must be in bytes?  (Some other offsets are in bits so I think it's worth noting.)

va/va.h
```diff
@@ -3193,15 +3214,17 @@ typedef  struct _VACodedBufferSegment  {
...
-    void               *next;
-
-    /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_LOW];
+    void            *next;
+    /** 
+      *\brief IV or Counter for encrytion filled by driver.
+      * Only available when status is set with \c VA_CODED_BUF_STATUS_VALID_IV.
+      */ 
+    uint32_t        encryption_iv[4];
```
fhvwy: Any particular reason why this is uint32_t[4]?  uint8_t[16] might be clearer, and would avoid any possible endianness issues.

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+    /** \brief  The length of the segments in bytes*/
+    uint32_t segment_length;
+    /** \brief  The length in bytes of the remainder of an incomplete block
+     *  from a previous segment*/
+    uint32_t partial_aes_block_size;
+    /** \brief  The length in bytes of the initial clear data */
+    uint32_t init_byte_length; 
+    /** \brief  This will be AES 128 counter for secure decode and secure
+     *  encode when numSegments equals 1 */
+    uint32_t aes_cbc_iv_or_ctr[4];
```
fhvwy: Same comment as above for the type of this field.

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+    int32_t max_long_term_frame_idx_plus1[32];
+    /** \brief Same as the H.264 bitstream syntax element. */
+    int32_t long_term_frame_idx[32];
+    /**@}*/
+    /** \brief Pointer to the next #VACencSliceParameterBufferH264 element,
+    * or \c NULL if there is none.*/
+    void *next;
+    /** \brief Reserved bytes for future use, must be zero */
+    unsigned long va_reserved[VA_PADDING_MEDIUM];
+} VACencSliceParameterBufferH264;
```
fhvwy: Can you explain how providing the reference pictures to the decoder works here?  With the current structures I'm not sure you have enough information to fill RefPicList0/RefPicList1.

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+    /** \brief Buffer formatted as raw data from bitstream for sequence parameter,
+     *  picture parameter, SEI parameters. Or \c NULL means buf is invalid.*/
+    void               *buf;
+    /** \brief Slice buffer type, see*/
+    VACencSliceBufType slice_buf_type;
+    /** \brief Slice buffer size in bytes. 0 means slice_buf is invalid*/
+    uint32_t           slice_buf_size;
+    /** \brief Slice buffer, parsed slice header information. Or \c NULL
+     *  means slice_buf is invalid.*/
+    void               *slice_buf;
```
fhvwy: I'm not quite sure I understand why there are separate `buf` and `slice_buf` elements here.  Won't you always want exactly one of them?  (Since you get an array of `VACencStatusBuf` from `vaQueryCenc()`, I would expect each one to contain a single bitstream element - either a raw decrypted header or one of the parameter structures above.)

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+ * Block call to get cenc status report. 
+ * After the application sends out cenc workload, it must call vaQueryCenc
+ * to get cenc status report for decrypt result.
+ * upon the return, info will point to an array of \ref VACencStatusBuf structure,
+ * The array is terminated if "status==-1" is detected.
+ */
+VAStatus vaQueryCenc(
+    VADisplay dpy,
+    VABufferID buffer,
+    VACencStatusBuf *info
```
fhvwy: Who allocates the memory for the array here?

va/va_cp.h
```diff
@@ -0,0 +1,420 @@
...
+);
+
+/**
+ * Make the end of rendering for Cenc decode for the encrypted bitstream. 
+ * The server should start all pending rendering operations and decryption 
+ * processing related with  this encrypted bitstream.  */
+VAStatus vaEndCenc (
+    VADisplay dpy,
+    VAContextID context
+);
```
fhvwy: Does this need to be a separate call?  Since you're using the `vaRender*()` calls for the rest of it, it looks like `vaEndPicture()` should be sufficient (though I may well be missing something).

chivakker:
@AppleWang001 can you provide a simple test example with your patch?  I am not opposed to this idea in general but I'd like to see it operational.  That test should be in libva-utils.  I'd like to then comeback here and provide a review.

PR #211, opened 2018-05-21
add RequestCRC flag of hevc picture parameters and crc valuses are requested if set to 1

Jexu:


va/va_dec_hevc.h
```diff
@@ -162,8 +162,10 @@ typedef struct  _VAPictureParameterBufferHEVC
             uint32_t        IdrPicFlag                                  : 1;
             /** current picture has only intra slices */
             uint32_t        IntraPicFlag                                : 1;
+            /** CRC valuses are requested if set to 1 */
+            uint32_t        RequestCRC                                  : 1;
```
fhvwy: Since you're adding this to H.265 standard parts, presumably this is the CRC as defined in §D.3.20?

How will the CRC values for each colour component be retrieved from the decoder?

Jexu: It is defined in that.  @XinfengZhang adds an interface to get the calculated values from decoder in PR https://github.com/intel/libva/pull/213

fhvwy: Add a bit more comment to explain it?

Will anyone ever want the other hash types?  If so, it might be better to add a separate ```uint8_t hash_type``` (as table D-10) and then make this bit be ```request_hash``` (or something along those lines).

PR #213, opened 2018-05-24
add output value interface for the values from driver

XinfengZhang:
sometimes , there are only one single value output
a new buffer type is complex than a api call

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -3625,6 +3625,24 @@ VAStatus vaQuerySurfaceError(
     void **error_info
 );
 
+/**out put value types, used by \vaQueryOutputValue */
+typedef enum
+{
+    VAOutputDecodeCRC = 1
+}VAOutputValueTypes;
+
+/**
+ * this interface is to get some values from driver or HW
+ * after calling vaSyncSurface, the value type is defined with \c VAOutputValueTypes
+ * the output_value should be a pointer allocated by application.
+ * it can be a single value or a structure.
+ */
+VAStatus vaQueryOutputValue(
+    VADisplay dpy,
+    VAContextID context,
+    VAOutputValueTypes output_value_type,
+    void *output_value);
+
 /**
  * Images and Subpictures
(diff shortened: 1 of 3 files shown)


fhvwy:
It might be sensible to have some sort of size information here.  While the CRC hash in H.265 has a fixed size (number of planes * 16 bits, so usually 6 bytes), this API might be useful to use for something which doesn't.

That could either be passed in the function (e.g. an extra argument ```size_t *output_value_size``` set on input to the size of the buffer the user provides and on output to the actual amount written to ```output_value```, with some error if it's not big enough), or there could be some separate query function to find out how big the ```output_value``` buffer needs to be.

va/va_backend.h
```diff
@@ -486,8 +486,15 @@ struct VADriverVTable
             void               *descriptor      /* out */
         );
 
+        VAStatus
+        (*vaQueryOutputValue)(
+            VADriverContextP    ctx,
+            VAContextID         context,          /* in */
```
Jexu: could also add the surfaceID '**VASurfaceID         render_target**' in this interface? it maybe more flexiable for usage.

va/va.h
```diff
@@ -3625,6 +3625,24 @@ VAStatus vaQuerySurfaceError(
...
+
+/**
+ * this interface is to get some values from driver or HW
+ * after calling vaSyncSurface, the value type is defined with \c VAOutputValueTypes
+ * the output_value should be a pointer allocated by application.
+ * it can be a single value or a structure.
+ */
+VAStatus vaQueryOutputValue(
+    VADisplay dpy,
+    VAContextID context,
```
Jexu: could also add the surfaceID 'VASurfaceID render_target' in this interface? it maybe more flexiable for usage.

XinfengZhang: @Jexu, if add VASurfaceID render_target, it seems that it was changed to another function, calculate the input surface CRC?

PR #214, opened 2018-05-25
add hevc subsets parameters structure

Jexu:


va/va_dec_hevc.h
```diff
@@ -561,6 +565,22 @@ typedef struct  _VASliceParameterBufferHEVCExtension
...
+ * order with first slice segment first, second slice segment second, etc... The total entry number
+ * is indicated by parameter TotalNumEntryPointOffsets(no definition) in VAPictureParameterBufferHEVC.
+ * And the first entry position of the entry point offsets for any slice segment is indicated by
+ * parameter entry_offset_to_subset_array in VAPictureParameterBufferHEVC data structure.
+ */
+typedef struct _VASubsetsParameterBufferHevc
+{
+    /** HEVC syntaxt element. */
+    uint32_t entry_point_offset_minus1[440]; 
+} VASubsetsParameterBufferHevc;
```
XinfengZhang: this structure will be a stable one ? will not changed in future, right?

Jexu: @XinfengZhang yeah, will not change.

va/va_dec_hevc.h
```diff
@@ -481,10 +481,14 @@ typedef struct  _VASliceParameterBufferHEVC
...
     uint8_t                 five_minus_max_num_merge_cand;
+    /** HEVC syntax element. */
+    uint8_t                 num_entry_point_offsets;
+    /** HEVC syntax element. */
+    uint8_t                 entry_offset_to_subset_array;
     /**@}*/
 
     /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_LOW];
+    uint32_t                va_reserved[VA_PADDING_LOW - 1];
```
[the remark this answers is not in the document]
Jexu: @XinfengZhang you mean that it needs to reduce 2?

va/va_dec_hevc.h
```diff
@@ -481,6 +481,10 @@ typedef struct  _VASliceParameterBufferHEVC
     int8_t                  ChromaOffsetL1[15][2];
     /** HEVC syntax element. */
     uint8_t                 five_minus_max_num_merge_cand;
+    /** HEVC syntax element. */
+    uint8_t                 num_entry_point_offsets;
+    /** HEVC syntax element. */
+    uint8_t                 entry_offset_to_subset_array;
```
fhvwy: The types of these elements don't look right (uint8_t -> max value 255).

va/va_dec_hevc.h
fhvwy: Could this be defined as an array of uint32_t instead of a structure?  If the buffer is of fixed size like this then it will usually be mostly unused, and may not be big enough for some degenerate streams (with both `tiles_enabled_flag` and `entropy_coding_sync_enabled_flag` the number of entry points can be very large).

va/va_dec_hevc.h
Jexu: @fhvwy yeah, i checked that the type of these elements are uint16_t

va/va_dec_hevc.h
fhvwy: Should be fine, but I'm not sure what you gain by the extra layer of indirection rather than putting the arbitrary-length array inside the buffer. (If you pass a pointer here then the driver will always have to copy the whole array immediately to some internal location, because the lifetime of an array outside the buffer is not constrained by the API when decoding asynchronously.)

Jexu: @fhvwy For some case where both tile and WPP exist, the number of entry points may be larger than 440. However WPP entry points are not needed for HW decoding, so we only need tile entry points that will not more than 440. As you said, if defining the buffer as fixed size, it will usually be mostly unsued, so it maybe better to allocate an array with the element size pointed by app for entry points

PR #220, opened 2018-06-15
add prediction direction caps report

XinfengZhang:
report which frame type is supported by driver

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -761,6 +761,18 @@ typedef enum
      * attribute value \c VAConfigAttribValMaxFrameSize represent max frame size support   
      */
     VAConfigAttribMaxFrameSize           = 38,
+    /** \brief inter frame prediction directrion attribute. Read-only.
+     * this attribute conveys the prediction direction (backward or forword) for specific config
+     * the value could be  VA_PREDICTION_DIRECTION_XXXX. it can be combined with VAConfigAttribEncMaxRefFrames
+     * to describe reference list , and the prediction direction. if this attrib is not present,both direction
+     * should be supported, no restriction.
+     * for example: normal HEVC encoding , maximum reference frame number in reflist 0 and reflist 1 is deduced
+     * by  VAConfigAttribEncMaxRefFrames. so there are typical P frame, B frame,
+     * if VAConfigAttribPredictionDirection is also present. it will stipulate prediction direction in both
+     * reference list. if only one prediction direction present(such as PREVIOUS),all reference frame should be
+     *  previous frame (PoC < current).
+     */
+    VAConfigAttribPredictionDirection   = 39,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
@@ -988,6 +1000,19 @@ typedef union _VAConfigAttribValEncJPEG {
 #define VA_ENC_QUANTIZATION_NONE                        0x00000000
 /** \brief Driver supports trellis quantization */
 #define VA_ENC_QUANTIZATION_TRELLIS_SUPPORTED           0x00000001
+/**@}*/
+
+/** @name Attribute values for VAConfigAttribPredictionDirection */
+/**@{*/
+/** \brief Driver support forward reference frame (inter frame for vpx, P frame for H26x MPEG)
+ * can work with the VAConfigAttribEncMaxRefFrames. for example: low delay B frame of HEVC.
+ * these value can be OR'd together. typical value should be VA_PREDICTION_DIRECTION_PREVIOUS
+ * or VA_PREDICTION_DIRECTION_PREVIOUS | VA_PREDICTION_DIRECTION_FUTURE, theoretically, there
+ * are no stream only include future reference frame.
+ */
+#define VA_PREDICTION_DIRECTION_PREVIOUS                0x00000001
+/** \brief Driver support backward prediction frame/slice */
+#define VA_PREDICTION_DIRECTION_FUTURE                  0x00000002
 /**@}*/
 
 /** @name Attribute values for VAConfigAttribEncIntraRefresh */
--- a/va/va_str.c
+++ b/va/va_str.c
@@ -124,6 +124,7 @@ const char *vaConfigAttribTypeStr(VAConfigAttribType configAttribType)
     TOSTR(VAConfigAttribQPBlockSize);
     TOSTR(VAConfigAttribStats);
     TOSTR(VAConfigAttribMaxFrameSize);
+    TOSTR(VAConfigAttribPredictionDirection);
     case VAConfigAttribTypeMax: break;
     }
     return "<unknown config attribute type>";


mypopydev:
I think  VAConfigAttribEncMaxRefFrames  have implicit support the Frame Type query, why we need another repetitive caps for Frame type query?

XinfengZhang:
VAConfigAttribEncMaxRefFrames can only represent the forward/backward reference number, but cant distinguish the difference between P and low delay B. although low-delay B have two reference list, but both reference list only has forward reference frame.

fhvwy:
I think you need to define more precisely what this actually means for each codec.

For H.264 / H.265, the low-delay B type gives you two lists but disallows bidirectional prediction modes in the macroblocks?  Or would there be a requirement that the POC values must always be increasing?

For VP9, the P type constrains all reference frames to have the same sign bias, while the two B types allow different sign bias but for the low-delay B type bidirectional prediction is not allowed?

For MPEG-2, the P and B types are the usual frame types, and low-delay B is not allowed?

fhvwy:
Sure, if the low-delay type doesn't apply to others then they don't need to be defined there.  It does still need a clear definition of what the frame types mean for the codecs which do support them, though.

ceyusa:
https://bugzilla.gnome.org/show_bug.cgi?id=796984 would depend on this API

XinfengZhang:
@fhvwy  hi Mark, how about current solution. combine the frame type and active reference frame number together to support low delay b?

va/va.h
```diff
@@ -748,6 +748,13 @@ typedef enum
      * support for QP info for buffer #VAEncQpBuffer.
      */
     VAConfigAttribQPBlockSize            = 37,
+    /**
+     * \brief frame type attribute. Read-only.
+     * this attribute conveys the frame type for specific config
+     * the value could be  VA_FRAME_TYPE_XXXX. for VPX, interframe also can be  VAA_FRAME_TYPE_P
+     * by default I frame is supported, no need to report it.
```
fhvwy: For backward compatibility it might be a good idea to write that if this attribute is not present then frame type support can be deduced from `VAConfigAttribEncMaxRefFrames` instead.

va/va.h
```diff
@@ -950,6 +957,19 @@ typedef union _VAConfigAttribValEncJPEG {
...
+/**@{*/
+/** \brief Driver support P frame (inter frame for vpx) 
+ * can work with the VAConfigAttribEncMaxRefFrames, if B frame is supported.
+ * and the max reference frame number in reference list 1 (future) is 0.
+ * means both reference frame list for B frame should only contain previous frames
+ * (POC is smaller than current B frame) it low delay mode */
+#define VA_FRAME_TYPE_P                                 0x00000001
+/** \brief Driver support B frame for MPEG2, H264, HEVC */
+#define VA_FRAME_TYPE_B                                 0x00000002
+/**@}*/
```
fhvwy: For H.26[45] this is really slice types rather than frame types?  (Frame types aren't really a thing in those codecs.)

Perhaps a table defining it precisely would make this clearer.  From what you've said I think what you want is:

| Codec | Always available | Also available with `VA_FRAME_TYPE_P` | Also available with `VA_FRAME_TYPE_B` |
| --- | --- | --- | --- |
| MPEG-2 | `picture_coding_type` = 1 | `picture_coding_type` = 2 | `picture_coding_type` = 3 |
| H.264 | `slice_type` = 2 or 7 | `slice_type` = 0 or 5 | `slice_type` = 1 or 6 |
| H.265 | `slice_type` = 2 | `slice_type` = 0 | `slice_type` = 1 |
| VP8 | `key_frame` = 1 | `key_frame` = 0 | - |
| VP9 | `frame_type` = 0 or `intra_only` = 1 | `frame_type` = 1 and `intra_only` = 0 | - |
| MJPEG | Everything | - | - |

va/va.h
```diff
@@ -748,6 +748,13 @@ typedef enum
      * support for QP info for buffer #VAEncQpBuffer.
      */
     VAConfigAttribQPBlockSize            = 37,
+    /**
+     * \brief frame type attribute. Read-only.
+     * this attribute conveys the frame type for specific config
+     * the value could be  VA_FRAME_TYPE_XXXX. for VPX, interframe also can be  VAA_FRAME_TYPE_P
+     * by default I frame is supported, no need to report it.
+     */
+    VAConfigAttibFrameType              = 38,
```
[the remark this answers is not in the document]
XinfengZhang: done

va/va.h
XinfengZhang: hi @fhvwy , thanks a lot for your comments,  agree with you ,  mostlikely , the P, B is slice type.  but it is a little complex. I prefer to change the attrib to prediction direction support. 
if only support low delay B: reference list0, reference list 1 represented by VAConfigAttribEncMaxRefFrames should be non zero value. and only forward direction is supported.

XinfengZhang:
change the PR . report a caps to indicate whether the forward (POC < current frame) or backward (POC> current frame)  prediction is supported.

XinfengZhang:
if no further comments, I will merge it in 1 week

fhvwy:
I don't understand the intent of this new proposal is.  The POC order has no effect on anything the encoder itself supports - the user is providing the reference frames directly, so it doesn't matter what order the frames are output in (though it could make use of it as a hint, and will need it if writing headers).  It's also unclear what the meaning would be for VP9, which does not have any global ordering number.

I assume this proposal is still wanting to cover the hardware bug referred to elsewhere in which some sorts of frames are not supported.  Would it be possible to explain precisely what the error is there?

XinfengZhang:
hi ,@fhvwy , the issue is, media driver low power HEVC encode just support I frame and low delay B frame. need a capability to export it.  for low delay B frame, the two reference list is same previous frames. no reference frame in future.  the old attribute VAConfigAttribEncMaxRefFrames  just can identify the reference number.  want an interfaces tell application , driver just support low delay B , not normal B.

such like:https://github.com/intel/libva/blob/master/va/va_enc_hevc.h#L462
when application choose hierarchical B, it also can choose low delay B to replace P frame. 
about VP9, 3 reference frames from 8 frames.  it should not care about this attributes. 

ps: I need to check whether media driver can enable normal B frames , will update later

XinfengZhang:
intel ICL low power encode, no future frame in L1. I afraid that some one set the B frame, and set normal reference frames, but encoder failed because driver only support low-delay B.  it's why I want to add this attributes, do you have other ideas to let application know the limitations?

cwang105:
Did not follow the thread. But adding a capability attribute for this (random access B) is reasonable.

Thanks,
Ce

va/va.h
```diff
@@ -761,6 +761,13 @@ typedef enum
      * attribute value \c VAConfigAttribValMaxFrameSize represent max frame size support   
      */
     VAConfigAttribMaxFrameSize           = 38,
+    /** \brief inter frame prediction directrion attribute. Read-only.
+     * this attribute conveys the prediction direction (backward or forword) for specific config
+     * the value could be  VA_PREDICTION_DIRECTION_XXXX. it can be combined with VAConfigAttribEncMaxRefFrames
+     * to describe reference list , and the prediction direction. if this attrib is not present, or value == 0,
+     * frame prediction can be deduced by VAConfigAttribEncMaxRefFrames.
```
xhaihao: How is the frame prediction deduced? It would be better to detail it in the comment. 

You defined VA_PREDICTION_DIRECTION_PREVIOUS (1) and VA_PREDICTION_DIRECTION_FUTURE (2) only, is the value set to 0?

XinfengZhang: if this attrib is not present, or value == 0,
     * frame prediction can be deduced by VAConfigAttribEncMaxRefFrames.

PR #212, opened 2018-05-22
add new BRC mode definition QVBR

XinfengZhang:
QVBR is Quality-Defined Variable Bitrate, comparing to VBR it add quality factor to the paramters

Signed-off-by: XinfengZhang <carl.zhang@intel.com>

PR #224, opened 2018-07-19
Add QVBR bitrate control method.

TangZhiZhen:
Add Quality-based variable bitrate control type.
Signed-off-by: ZhiZhen Tang <zhizhen.tang@intel.com>


XinfengZhang:
it is duplicated with #212 , and the value 0x00000060 does not make sense. please close it. thanks

PR #226, opened 2018-07-28
Check for -fstack-protector

bkuhls:
Not all toolchains provide support for -fstack-protector. This patch
provides a configure check to avoid build errors like

/home/buildroot/buildroot/output/host/opt/ext-toolchain/bin/../lib/gcc/x86_64-buildroot-linux-uclibc/6.4.0/../../../../x86_64-buildroot-linux-uclibc/bin/ld: cannot find -lssp_nonshared
/home/buildroot/buildroot/output/host/opt/ext-toolchain/bin/../lib/gcc/x86_64-buildroot-linux-uclibc/6.4.0/../../../../x86_64-buildroot-linux-uclibc/bin/ld: cannot find -lssp

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -209,6 +209,22 @@ if test "$ac_cv_have_gnuc_visibility_attribute" = "yes"; then
               [Defined to 1 if GCC visibility attribute is supported])
 fi
 
+# Check for -fstack-protector
+ssp_cc=yes
+if test "X$CC-cc" != "X"; then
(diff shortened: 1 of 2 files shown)


bkuhls:
'-fstack-protector' was added by PR #199.

bkuhls:
If the patch is accepted I will send follow-up patches to libva-utils & intel-vaapi-driver because they suffer from the same problem.

PR #228, opened 2018-08-16
Allow import of the DRM PRIME 2 memory type

fhvwy:
This adds support for import using the VADRMPRIMESurfaceDescriptor
structure, which enables use of format modifiers (required for some
compressed surfaces).

This was proposed with the original patches adding vaExportSurfaceHandle() last year, but it wasn't followed up because there wasn't a specific use-case for it at the time.  We have one now in <https://github.com/intel/libva/pull/203>, and an example is added for that case.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1260,7 +1260,12 @@ typedef enum {
     VASurfaceAttribMaxHeight,
     /** \brief Surface memory type expressed in bit fields (int, read/write). */
     VASurfaceAttribMemoryType,
-    /** \brief External buffer descriptor (pointer, write). */
+    /** \brief External buffer descriptor (pointer, write).
+     *
+     * Refer to the documentation for the memory type being created to
+     * determine what descriptor structure to pass here.  If not otherwise
+     * stated, the common VASurfaceAttribExternalBuffers should be used.
+     */
     VASurfaceAttribExternalBufferDescriptor,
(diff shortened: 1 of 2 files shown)


XinfengZhang:
so, in #203, I just need to export the fourcc and surface attributes, so , import will use VADRMPRIMESurfaceDescriptor instead of VASurfaceAttribExternalBuffers when the VASurfaceAttribMemoryType == PRIME_2 , right?

fhvwy:
Rebased, and updated the comment on `VASurfaceAttribExternalBufferDescriptor` to clarify what type needs to be passed when importing external memory.

XinfengZhang:
when VASurfaceAttribMemoryType is VA_SURFACE_ATTRIB_MEM_TYPE_VA VA_SURFACE_ATTRIB_MEM_TYPE_V4L2 VA_SURFACE_ATTRIB_MEM_TYPE_USER_PTR
VASurfaceAttribExternalBufferDescriptor is VASurfaceAttribExternalBuffers , if VASurfaceAttribMemoryType value is VA_SURFACE_ATTRIB_MEM_TYPE_KERNEL_DRM VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2 , VASurfaceAttribExternalBufferDescriptor is VADRMPRIMESurfaceDescriptor

PR #232, opened 2018-09-04
Add max frame size parameters for multi pass pak case in legacy mode

cwang64:
Signed-off-by: Wang, Chengwei C <chengwei.c.wang@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1748,6 +1748,8 @@ typedef enum
     VAEncMiscParameterTypeSkipFrame     = 9,
     /** \brief Buffer type used for region-of-interest (ROI) parameters. */
     VAEncMiscParameterTypeROI           = 10,
+    /** \brief Buffer type used to express a maximum frame size (in bytes) settings for multiple pass. */
+    VAEncMiscParameterTypeMultiPassFrameSize       = 11,
     /** \brief Buffer type used for temporal layer structure */
     VAEncMiscParameterTypeTemporalLayerStructure   = 12,
     /** \brief Buffer type used for dirty region-of-interest (ROI) parameters. */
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -2087,22 +2087,35 @@ typedef struct _VAEncMiscParameterHRD
...
-    /** \brief Maximum size of a frame (in bits). */
+    /** \brief Profile level maximum size of a frame (in bits). */
     uint32_t                max_frame_size;
-
-    /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_LOW];
+    /** \brief delta QP list for every pass, currently AVC encoder can support up to 4 passes*/
+    uint8_t                *multipak_delta_qp;
+    /** \brief frame level max frame size (in bytes) for AVC encoder in multi pak case */
+    uint32_t                multipak_max_frame_size;
```
QuPengfei: two comments:
- is it necessary to define the multipak_max_frame_size for multi pak case? Is it possible to reuse the member max_frame_size? 
- there is different unit size for multipak_max_frame_size(bytes) and max_frame_size(bits). is it OK to use the same unit?

BTW, here  max_frame_size is not the size specified in the spec for Profile/Level. it is configured by user.

cwang64: For the first question, we cannot combine these two type max frame size in one parameter, because the UMD processes these two type max frame size in different ways. 
For the second question, I suggest to change max_frame_size unit from bits to bytes, since AVC FEI and HEVC FEI max frame size parameters unit are bytes as multipak_max_frame_size. But change this unit from bits to bytes will impact the libva interface compatibility.

va/va.h
```diff
@@ -2105,6 +2107,33 @@ typedef struct _VAEncMiscParameterBufferMaxFrameSize {
...
+ * This misc parameter buffer defines the maximum size of a frame (in
+ * bytes) settings for multiple pass. currently only AVC encoder can
+ * support this settings in multiple pass case. If the frame size exceeds
+ * this size, the encoder will do more pak passes to adjust the QP value
+ * to control the frame size.
+ */
+typedef struct _VAEncMiscParameterBufferMultiPassMaxFrameSize {
+    /** \brief Type. Shall be set to #VAEncMiscParameterTypeMultiPassMaxFrameSize. */
+    VAEncMiscParameterType      type;
+    /** \brief Maximum size of a frame (in byte) for AVC encoder in multiple pass case */
```
dlin2: can remove "for AVC encoder in multiple pass case", it's a little redundant compared with the previous structure descriptions.

PR #238, opened 2018-09-19
new interface for query surface alignment

XinfengZhang:
this interface can used for surface importing, other component create surface to meet media driver requirement. 
application need to know the size and alignment requirement when it create a system memory surface

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -2210,3 +2210,22 @@ vaQueryVideoProcPipelineCaps(
     VA_TRACE_RET(dpy, status);
     return status;
 }
+
+VAStatus vaQuerySurfaceAllocation(
+    VADisplay           dpy,
+    uint32_t            format,
+    uint32_t            width,
+    uint32_t            height,
+    VASurfaceAttrib     *attrib_list,
+    unsigned int        num_attribs,
+    VASurfaceAllocationInfo *surfaceinfo
+)
+{
+    VADriverContextP ctx;
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+
+    if (!ctx->vtable->vaQuerySurfaceAllocation)
+        return VA_STATUS_ERROR_UNIMPLEMENTED;
+    return ctx->vtable->vaQuerySurfaceAllocation(ctx, format, width, height, attrib_list, num_attribs, surfaceinfo);
+}
--- a/va/va.h
+++ b/va/va.h
@@ -4225,6 +4225,64 @@ VAStatus vaDeriveImage (
     VAImage *image	/* out */
 );
 
+typedef struct _VASurfaceAllocationInfo{
+    /** \brief pixel format in fourcc. */
+    uint32_t    pixel_format;
+    /* picture width x height */
+    uint32_t	width; 
+    uint32_t	height;
+    /* suggest data_size, it alwsys is minimum requirement*/ 
+    uint32_t	data_size;
+    /* number of planes , can not be greater than 3 */
+    uint32_t	num_planes;
+    /* 
+     * An array indicating the scanline pitch in bytes for each plane.
+     * Each plane may have a different pitch. Maximum 3 planes for planar formats
+     */
+    uint32_t	pitches[4];
+    /* 
+     * An array indicating the byte offset from the beginning of the image data 
+     * to the start of each plane.
+     */
+    uint32_t	offsets[4];
+    /*
+     * An array of ascii characters describing the order of the components within the bytes.
+     * Only entry_bytes characters of the string are used.by default , the order is from pixel format
+     * if there are different layout, can use this value to describe it.
+     */
+    int8_t component_order[4];
+    /* base address alignment, the value should be 0x0, 0x1, 0x2, 0x4, 0x8... only one bit is 1 , other bit is zero*/
+    uint32_t    base_address_align;
+    /* memory size alignment */
+    uint32_t    size_align;
+    /* suggest memory flags, it should be VA_SURFACE_EXTBUF_DESC_XXX */
+    uint32_t    flags;
+    /** \brief Reserved bytes for future use, must be zero */   
+    uint32_t    va_reserved[VA_PADDING_MEDIUM];
+
+}VASurfaceAllocationInfo;
+
+/** brief Queries surface allocation requirements for the userptr/dma buffer import.
+  *
+  * it can be a reference when call vaCreateSurface
+  * with system memory.such as user ptr 
+  *
+  * @param[in] dpy               the VA display
+  * @param[in] width             surface width
+  * @param[in] height            surface height
+  * @param[in] attrib_list       surface attrib list
+  * @param[in] num_attribs       surface attrib number
+  * @param[out] surfaceinfo      surface allocation information */
+VAStatus vaQuerySurfaceAllocation(
+    VADisplay           dpy,
+    uint32_t            format,
+    uint32_t            width,
+    uint32_t            height,
+    VASurfaceAttrib     *attrib_list,
+    unsigned int        num_attribs,
+    VASurfaceAllocationInfo *surfaceinfo
+);
+
 /**
  * Subpictures 
  * Subpicture is a special type of image that can be blended 
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -485,8 +485,18 @@ struct VADriverVTable
             void               *descriptor      /* out */
         );
 
+        VAStatus
+        (*vaQuerySurfaceAllocation)(
+            VADisplay           dpy,
+            uint32_t            format,         /* in */
+            uint32_t            width,          /* in */
+            uint32_t            height,         /* in */
+            VASurfaceAttrib     *attrib_list,   /* in */
+            unsigned int        num_attribs,    /* in */ 
+            VASurfaceAllocationInfo *surfaceinfo /* out */
+        );
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[57];
+        unsigned long reserved[56];
 };
 
 struct VADriverContext


va/va.c
```diff
@@ -2128,3 +2128,20 @@ vaQueryVideoProcPipelineCaps(
...
+    uint32_t  width,
+    uint32_t  height,
+    uint32_t  memflags,
+    VAAppSurfaceResInfo *surfaceinfo
+)
+{
+    VADriverContextP ctx;
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+
```
fhvwy: This needs to check whether the pointer is present, for backward compatibility.

va/va.h
```diff
@@ -4182,6 +4182,37 @@ VAStatus vaDeriveImage (
     VAImage *image	/* out */
 );
 
+typedef struct _VAAppSurfaceResInfo{
+    uint32_t image_width;
```
dvrogozh: This structure declaration looks inconsistent with the rest of the VAAPI. I guess image_height and height are used to specify offset offset between planes, right? But rest of VAAPI, look into VAImage for example, uses offsets[4] for this.

I also think that better if the structure will contain format inside. Comments are needed for some fields, for example for base_addr_align - in which values, bytes, pages, etc.?

I would actually suggest to almost copy declaration of VAImage for this purpose. What you are trying to achieve seems to be almost identical to VAImage structure definition minus allocated data IDs like image_id and buf, plus some alignment hints. It seems something like this:

typedef stuct _VAImageAllocationInfo {
    uint16_t    width;
    uint16_t    height;
    uint32_t    data_size;
    uint32_t    num_planes;     /* can not be greater than 3 */
    uint32_t    pitches[3];
    uint32_t    offsets[3];
    int32_t num_palette_entries;   /* set to zero for non-palette images */
    int32_t entry_bytes;
    int8_t component_order[4];
    uint32_t base_addr_align; /* in bytes, for example =4096 (page alignment) */
    union{
        struct{
        uint32_t islinear  :1;
        uint32_t reserved  :31;
        }bits;
        uint32_t flags;
    };
    uint32_t                va_reserved[VA_PADDING_LOW];
} VAImageAllocationInfo;

va/va.h
```diff
@@ -4182,6 +4182,37 @@ VAStatus vaDeriveImage (
...
+    uint32_t image_height;
+    uint32_t total_size;
+    uint32_t num_planes;
+    uint32_t pitch[4];
```
dvrogozh: why 4 if the rest of VAAPI uses 3?

va/va.h
```diff
@@ -4182,6 +4182,37 @@ VAStatus vaDeriveImage (
...
+ * it can be a reference when call vaCreateSurface
+ * with system memory.such as user ptr */
+VAStatus vaQuerySurfaceAllocation(
+    VADisplay dpy,
+    uint32_t  format,
+    uint32_t  fourcc,
+    uint32_t  width,
+    uint32_t  height,
+    uint32_t  memflags,
+    VAAppSurfaceResInfo *surfaceinfo
```
dvrogozh: structure name looks weird

va/va.h
XinfengZhang: thanks , accept, but remove     int32_t num_palette_entries;   /* set to zero for non-palette images */
    int32_t entry_bytes;
because the palette format is not used now, to avoid risk , I add more reserved byte , if there are requirement , we can add them back

va/va.h
XinfengZhang: VASurfaceAttribExternalBuffers is using 4, 4 element for some planar format? such as A, Y, U, V planar, I am not sure. but anyway, reserved more is flexible than 3.

va/va.h
XinfengZhang: changed

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
     VAImage *image	/* out */
 );
 
+typedef struct _VASurfaceAllocationInfo{
```
dvrogozh: I think this structure should be aligned with VASurfaceAttribExternalBuffers defined here: https://github.com/intel/libva/blob/95eb8cf469367b532b391042fa0e77ca513ac94e/va/va.h#L1379

Optimally allocation info structure should be just used inside VASurfaceAttribExternalBuffers, but likely we will have a chance to do that only with the next major version change if it will ever happen.

Anyhow, following comments follow alignment idea.

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+    /* format fourcc */
+    uint32_t    fourcc;
```
dvrogozh: s/fourcc/pixel_format/

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+    /* picture width x height */
+    uint16_t	width; 
+    uint16_t	height;
```
dvrogozh: s/uint16_t/uint32_t/ for both width and height

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+     * An array indicating the byte offset from the beginning of the image data 
+     * to the start of each plane.
+     */
+    uint32_t	offsets[4];
+    /*
+     * An array of ascii characters describing the order of the components within the bytes.
+     * Only entry_bytes characters of the string are used.by default , the order is from format fourcc
+     * if there are different layout, can use this value to describe it.
+     */
+    int8_t component_order[4];
```
dvrogozh: do we actually need component_order? Isn't it assumed by the color format? Also, VASurfaceAttribExternalBuffers does not have it, so it's a question how user will be able to use component_order info?

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+    /* base address alignment*/
+    uint32_t    base_address_align;
```
dvrogozh: I suggest to specify in which values we expect alignment: "base address alignment in bytes"

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+    /* memory size alignment */
+    uint32_t    size_align;
+    /* suggest memory flags */
+    union{
+        struct{
+        uint32_t islinear :1;
```
dvrogozh: VASurfaceAttribExternalBuffers has this flag:
https://github.com/intel/libva/blob/95eb8cf469367b532b391042fa0e77ca513ac94e/va/va.h#L1407

Should not we reuse it instead of defining something new?

Also, how we intend to use this query structure: to recommend allocation parameters for userptr or we will also cover dmabuf with that? It is not clear to me... Worth explicit comment for whole structure definition.

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+        uint32_t reserved :3;
+        }bits;
+        uint32_t flags;
+    };
+    /** \brief Reserved bytes for future use, must be zero */   
+    uint32_t                va_reserved[VA_PADDING_MEDIUM];
+
+}VASurfaceAllocationInfo;
+
+/** get the real driver surface information
```
dvrogozh: Suggest to rewrite as follows:
/**
 * \brief Queries surface allocation requirements for the userptr buffer import. */

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+ * it can be a reference when call vaCreateSurface
+ * with system memory.such as user ptr */
+VAStatus vaQuerySurfaceAllocation(
+    VADisplay           dpy,
```
dvrogozh: Please, add all parameters to the doxygen info like:
`@param[in] dpy               the VA display`
Mark input/output parameters.

va/va.h
```diff
@@ -4225,6 +4225,62 @@ VAStatus vaDeriveImage (
...
+    uint32_t            format,
+    uint32_t            width,
+    uint32_t            height,
+    VASurfaceAttrib     *attrib_list,
+    unsigned int        num_attribs,
+    VASurfaceAllocationInfo *surfaceinfo
```
dvrogozh: Shouldn't we just return VASurfaceAttribExternalBuffers instead of a new structure?

I also wonder whether `surfaceinfo` is input or output parameter? I am confused because of `islinear`... And if that's input/output parameter then width/height are kind of duplicate which is not good I think.

va/va.h
XinfengZhang: yes, cant add additional member in VASurfaceAttribExternalBuffers now. major version change can break ABI .

va/va.h
XinfengZhang: if we remove it, will specify the order, fro example: YUV444 planar, but maybe some application use different buffer to store each plane, specify it should follow Y, U, V order.

va/va.h
XinfengZhang: it should always be with format b0000000000x0000000... so lets comment it

va/va.h
XinfengZhang: both.

va/va.h
XinfengZhang: also want to use VASurfaceAttribExternalBuffers  to be output , but there are no base_address and page requirement  .

XinfengZhang: or we use such definition typedef struct _VASurfaceAllocationInfo{ VASurfaceAttribExternalBuffers externalbufferdesc, uint32_t    base_address_align; uint32_t    size_align; uint32_t    va_reserved[VA_PADDING_MEDIUM];
}

sreerenjb:
IMO, This API needs to be revisited for including the decode use cases too. If the backend drivers require codec (and gen??) specific tiling and vertical+horizontal pitch alignment for decode target, it has to be communicated to the middleware, especially in the case of the dmabuf use cases. One option could be to pass the vaConfigID as input.

XinfengZhang:
yes, different config should have different request, this PR is too old, let me close it and file another one ， it is better to just add a surfaceattributes instead of a new API function, thanks

PR #243, opened 2018-10-11
add HW assistant copy interface

XinfengZhang:

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -2286,3 +2286,23 @@ vaQueryVideoProcPipelineCaps(
     VA_TRACE_RET(dpy, status);
     return status;
 }
+
+VAStatus
+vaCopy(
+    VADisplay         dpy,
+    VACopyObject      *dst,
+    VACopyObject      *src,
+    VACopyOption      option
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  if(ctx->vtable->vaCopy  == NULL)
+      va_status = VA_STATUS_ERROR_UNIMPLEMENTED;
+  else
+      va_status = ctx->vtable->vaCopy( ctx, dst, src, option);
+  return va_status;
+}
--- a/va/va.h
+++ b/va/va.h
@@ -4889,6 +4889,47 @@ typedef struct _VAPictureHEVC
  */
 #define VA_PICTURE_HEVC_RPS_LT_CURR             0x00000040
 
+typedef enum{
+    VACopyObjectSurface = 0,
+    VACopyObjectBuffer  = 1,
+} VACopyObjectType;
+
+typedef struct _VACopyObject {
+    VACopyObjectType  obj_type;    // type of object.
+    union
+    {
+        VASurfaceID surface_id;
+        VABufferID  buffer_id;
+    } object;
+
+    uint32_t    va_reserved[VA_PADDING_MEDIUM];
+} VACopyObject;
+
+typedef union _VACopyOption{
+    struct {
+        /** \brief va copy synchronization, the value should be /c VA_EXEC_SYNC or /c VA_EXEC_ASYNC */
+        uint32_t va_copy_sync : 2;
+        /** \brief va copy mode, the value should be VA_EXEC_MODE_XXX */
+        uint32_t va_copy_mode : 4;
+        uint32_t reserved     :26;
+    }bits;
+    uint32_t value;
+}VACopyOption;
+
+/** \brief Copies an object.
+ *
+ * Copies specified object (surface or buffer). If non-blocking copy
+ * is requested (VA_COPY_NONBLOCK), then need vaSyncBuffer or vaSyncSurface/vaSyncSurface2
+ * to sync the destination object.
+ *
+ * @param[in] dpy               the VA display
+ * @param[in] dst               Destination object to copy to
+ * @param[in] src               Source object to copy from
+ * @param[in] option            VA copy option
+ * @return VA_STATUS_SUCCESS if successful
+ */
+VAStatus vaCopy(VADisplay dpy, VACopyObject * dst, VACopyObject * src, VACopyOption option);
+
 #include <va/va_dec_hevc.h>
 #include <va/va_dec_jpeg.h>
 #include <va/va_dec_vp8.h>
--- a/va/va.h
+++ b/va/va.h
@@ -358,6 +358,17 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_PADDING_HIGH         16
 #define VA_PADDING_LARGE        32
 
+/** operation options */
+/** synchronization, block call, output should be ready after execution function return*/ 
+#define VA_EXEC_SYNC              0x0
+/** asynchronization,application should call additonal sync operation to access output */
+#define VA_EXEC_ASYNC             0x1
+
+/** operation mode */
+#define VA_EXEC_MODE_DEFAULT      0x0
+#define VA_EXEC_MODE_POWER_SAVING 0x1
+#define VA_EXEC_MODE_PERFORMANCE  0x2
+
 /**
  * Returns a short english description of error_status
  */
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -484,7 +484,6 @@ struct VADriverVTable
             uint32_t            flags,          /* in */
             void               *descriptor      /* out */
         );
-
         VAStatus (*vaSyncSurface2) (
             VADriverContextP ctx,
             VASurfaceID surface,
@@ -497,8 +496,15 @@ struct VADriverVTable
             uint64_t timeout_ns
         );
 
+        VAStatus
+        (*vaCopy)(
+            VADriverContextP    ctx,            /* in */
+            VACopyObject        *dst,           /* in */
+            VACopyObject        *src,           /* in */
+            VACopyOption        option          /* in */
+        );
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[55];
+        unsigned long reserved[54];
 };
 
 struct VADriverContext


va/va.h
```diff
@@ -4492,6 +4492,42 @@ typedef struct _VAPictureHEVC
...
+    VACpSurfaceToSurface  = 1,
+    VACpSurfaceToBuffer   = 2,
+    VACpSurfaceToUserPTR  = 3,
+    VACpBufferToSurface   = 4,
+    VACpBufferToUserPTR   = 5,
+    VACpBufferToBuffer    = 6,
+    VACpUserPTRToSurface  = 7,
+    VACpUserPTRToBuffer   = 8,
+    VACpUserPTRToUserPTR  = 9
+} VAGpuAccelCpDirect;
```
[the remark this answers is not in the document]
XinfengZhang: will change it to definitions which can be OR'd together for the query function.

va/va_backend.h
```diff
@@ -485,9 +485,24 @@ struct VADriverVTable
             uint32_t            flags,          /* in */
             void               *descriptor      /* out */
         );
+        VAStatus
+        (*vaGpuAccelCopy)(
+            VADriverContextP    ctx,            /* in */
+            VAGpuCpObject      *dst,            /* in */
+            VAGpuCpObject      *src,            /* in */
+            VAGpuAccelCpDirect  direct,         /* in */
```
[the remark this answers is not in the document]
XinfengZhang: thanks , will follow your comments

va/va.h
```diff
@@ -4492,6 +4492,42 @@ typedef struct _VAPictureHEVC
...
+
+typedef struct _VAGpuCpObject{
+    union
+    {
+        VASurfaceID surface_id;
+        VABufferID  Buffer_id;
+        uint8_t *   userptr;       //pointer to system memory
+    }obj;
+    uint32_t    size;              // for userptr, it should be same with surface/buffer size
```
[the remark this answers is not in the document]
XinfengZhang: accept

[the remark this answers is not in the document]
XinfengZhang: about each plan, I suppose all format & size of destination and source   should be same

va/va.h
```diff
@@ -4492,6 +4492,42 @@ typedef struct _VAPictureHEVC
...
+
+#define VA_GPU_ACCEL_COPY_BLOCK    0x00000001
+#define VA_GPU_ACCEL_COPY_NONBLOCK 0x00000002
+
+/** GPU accelarate copy
+ * it can be block copy or non block copy
+ * if it is non block copy, a handle is returned for wait funtion*/
+VAStatus vaGpuAccelCopy(VAGpuCpObject * dst, VAGpuCpObject * src, uint32_t option, void **cphandle);
+
+VAStatus vaGpuAccelCopyWait(void * cphandle);
```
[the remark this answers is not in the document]
XinfengZhang: one handle per copy, it is only for non-block call

XinfengZhang:
@fhvwy  hi Mark, I changed it to HW assistant copy.  actually, most cases, it is faster than CPU side, for example, copy tile surface to linear memory.  offload CPU usage is also an option.

va/va.c
```diff
@@ -2128,3 +2128,37 @@ vaQueryVideoProcPipelineCaps(
...
+    VAGpuAccelCpDirect  direct,
+    uint32_t option,
+    void **cphandle
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
```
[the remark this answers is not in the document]
XinfengZhang: accept

va/va.c
```diff
@@ -2128,3 +2128,41 @@ vaQueryVideoProcPipelineCaps(
...
+    VAHWCpObject      *src,
+    uint32_t          option,
+    uintptr_t*        cphandle
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
```
dmitryermilov: Carl, I'd like to propose a more abstract solution, **not to stick with HW copy**.  For video<->system memory copy, the copy (with tiling/detiling) can can be done by SW or HW (EUs or blitter), right? If so, does it look reasonable for you to introduce vaAsstCopy(), the function will accept a flag which will define copy type (SW, EUs, Blitter)?

dvrogozh: I would better go with as simple as `vaCopy()`.

va/va.h
```diff
@@ -4620,6 +4620,46 @@ typedef struct _VAPictureHEVC
...
+ * is requested (VA_COPY_NONBLOCK), sync_handle will contain a handle
+ * to pass to vaCopyWait to wait for completion.
+ *
+ * @param[in] dpy               the VA display
+ * @param[in] src               Source object to copy from
+ * @param[in,out] buf_info      Destination object to copy to
+ * @param[out] sync_handle      Synchronization handle to sync on
+ * @return VA_STATUS_SUCCESS if successful
+ */
+VAStatus vaCopy(VADisplay dpy, VACopyObject * src, VACopyObject * dst, uint32_t flags, uintptr_t* sync_handle);
```
dmitryermilov: A proposal: within dGFX should we consider cross adapter (or at least cross vaDisplay) copy? For now it’s not the case, but might become needed later. But later changes are always hard. 
So probably VACopyObject should include VADisplay for VASurfaceID/VABufferID.

va/va.c
FurongZhang: Hi Carl, May I know what kinds of usage scenarios this vaCopy is for? We have buffer, image, surface definition.  And the input/output parameters are very limited, so what is the target usage for this vaCopy?

dmitryermilov:
Hi @XinfengZhang ,

I still have the following proposals :

1. within dGFX we should consider cross adapter (or at least cross vaDisplay) copy. So probably VACopyObject should include VADisplay for VASurfaceID/VABufferID.
2.  We need an input rectangle and coordinates (a point) at dst surface where to put data.
3. We need to provide an example how to use this interface for video->system and system->video memory copy. Probably it even won't require interface change. E.g. system memory can be represented by vaSurfaceId with userptr.
4. We need to consider supporting cases when planes of surfaces in system memory are allocated by different malloc calls (ffmpeg, HB do it).

Without it, I'm afraid the interface won't be as powerful and useful as it can be. Do the proposals make sense to you?

va/va.h
dmitryermilov: Updated here: https://github.com/intel/libva/pull/243#issuecomment-672289401

XinfengZhang:
> Hi @XinfengZhang ,
> 
> I still have the following proposals :
> 
> 1. within dGFX we should consider cross adapter (or at least cross vaDisplay) copy. So probably VACopyObject should include VADisplay for VASurfaceID/VABufferID.
> 2. We need an input rectangle and coordinates (a point) at dst surface where to put data.
it sound like composition? should we call VPP to handle it?
> 3. We need to provide an example how to use this interface for video->system and system->video memory copy. Probably it even won't require interface change. E.g. system memory can be represented by vaSurfaceId with userptr.
agree, we should add a sample app in https://github.com/intel/libva-utils
> 4. We need to consider supporting cases when planes of surfaces in system memory are allocated by different malloc calls (ffmpeg, HB do it).
there are a lot of restriction of systeam memory allocated by application through malloc. such as the start address should be aligned with page (4k)
> 
> Without it, I'm afraid the interface won't be as powerful and useful as it can be. Do the proposals make sense to you?

dmitryermilov:
> > I still have the following proposals :
> > 
> > 1. within dGFX we should consider cross adapter (or at least cross vaDisplay) copy. So probably VACopyObject should include VADisplay for VASurfaceID/VABufferID.
> > 2. We need an input rectangle and coordinates (a point) at dst surface where to put data.
> 
> it sound like composition? should we call VPP to handle it?

Why VPP? Why composition?  Please see how other similar interfaces are defined:
https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-copysubresourceregion



> > 1. We need to consider supporting cases when planes of surfaces in system memory are allocated by different malloc calls (ffmpeg, HB do it).
> 
> there are a lot of restriction of systeam memory allocated by application through malloc. such as the start address should be aligned with page (4k)

And ? Do you want to say we won't be able all scenarios. It's okay. But we should handle the scenarios which we can handle. Plus anyway, this're discussing interfaces, not implementation. Interacts doesn't care about alignment,

va/va.h
```diff
@@ -4730,6 +4730,56 @@ typedef struct _VAPictureHEVC
...
+
+typedef struct _VACopyObject {
+    VACopyObjectType  obj_type;    // type of object.
+    union
+    {
+        VASurfaceID surface_id;
+        VABufferID  buffer_id;
+    } object;
+
+    uint32_t    va_reserved[VA_PADDING_MEDIA];
```
dmitryermilov, line 4746: VA_PADDING_MEDIA or VA_PADDING_MEDIUM ?

dmitryermilov (approved):

va/va.h
```diff
@@ -4730,6 +4730,56 @@ typedef struct _VAPictureHEVC
...
+#define VA_COPY_SYNC_NONBLOCK 0x1
+
+#define VA_COPY_MODE_BALANCE      0x0
+#define VA_COPY_MODE_PERF         0x1
+#define VA_COPY_MODE_POWER_SAVING 0x2
+
+typedef union _VACopyOption{
+    struct {
+        /** \brief va copy synchronization, the value should be VA_COPY_SYNC_XXX
+         * if it equal VA_COPY_SYNC_NONBLOCK , sync handle should be retrieved.
```
dmitryermilov, line 4759: @XinfengZhang , could you remind me (and better to describe in va.h) what sync handle is and how to retrieved

va/va.h
```diff
@@ -4791,6 +4791,56 @@ typedef struct _VAPictureHEVC
...
+typedef enum{
+    VACopyObjectSurface = 0,
+    VACopyObjectBuffer  = 1,
+} VACopyObjectType;
+
+typedef struct _VACopyObject {
+    VACopyObjectType  obj_type;    // type of object.
+    union
+    {
+        VASurfaceID surface_id;
```
dvrogozh, line 4803: These ID types goes as:
```
typedef unsigned int VAGenericID;
```

which might have different size... Can we add some specific size to this union? Like:
```
    union
    {
        VASurfaceID surface_id;
        VABufferID  buffer_id;
        uint64_t reserved;
    } object;
```

dvrogozh (changes requested):

dmitryermilov (approved):

va/va.h
```diff
@@ -4791,6 +4791,56 @@ typedef struct _VAPictureHEVC
...
+        /** \brief va copy synchronization, the value should be VA_COPY_SYNC_XXX
+         * if it equal VA_COPY_SYNC_NONBLOCK , dst surface should be used for sync operation
+         */
+        uint32_t va_copy_sync : 2;
+        /** \brief va copy mode, the value should be VA_COPY_MODE_XXX */
+        uint32_t va_copy_mode : 4;
+        uint32_t reserved     :26;
+    }bits;
+    uint32_t value;
+}VACopyOption;
```
dvrogozh, line 4828: We've also planned to expose vaMapBuffer2 API w/ flags and maybe w/ copy options... See #377 for the early draft.

Can we, please, consider whether we can apply `VACopyOption` to the vaMapBuffer2 API as well? I worry about va_copy_sync usage... Thought it might still be useful in vase of vaUnmap when we might or might not to wait for the sync...

dmitryermilov: >I worry about va_copy_sync usage... Thought it might still be useful in vase of vaUnmap when we might or might not to wait for the sync... 

@dvrogozh , do you mean asynchronous writing to vaSurface?

dvrogozh: Possible usages:
```
vaMapBuffer2(VA_COPY_SYNC_NONBLOCK); // allocates (maybe) shadow copy, no
                              // copy from original surface due to VA_COPY_SYNC_NONBLOCK (we don't care about the contetn)
memcpy();
unmap(); // schedules copy operation form shadow copy to final surface, we don't wait due to VA_COPY_SYNC_NONBLOCK
```

va/va.c
XinfengZhang: hi Furong, vaCopy is used to faster some copy, use HW to handle the memory copy for better performance, for example, copy memory from local to system, local to local ...

va/va.h
XinfengZhang: thanks , I removed it, we should use vaSyncSurface or vaSyncBuffer to handle it

va/va.h
XinfengZhang: to enlarge the size to 64bit?  which compiler/platform will lead to different size of unsigned int?

dvrogozh: not sure, but I am paranoid:)

va/va.h
dmitryermilov: Yes, it's async writing to vaSurface/vaBuffer. Above, you say "_usages"_. To tell you the truth I don't see other examples. 
But anyway, IMO, it indeed can be useful.

XinfengZhang: it means map is block call , but unmap is un-block call , right?  the interface will be a little weird. need to re-consider it. maybe also a unmap2?

va/va.h
XinfengZhang: maybe we could consider about the "uint32_t"/"uint64_t" "unsigned long"  for future, such as 3.0? , unsigned long will be changed adaptively with the compiler. uint64_t /uint32_t will be dedicate size.

dvrogozh (approved):

PR #245, opened 2018-11-06
drm: Add iHD to driver_name_map

### LINKED ISSUE #244: driver auto-detection doesn't work for iHD under X11
Quoting from 9f2a994643001a03e103aa13268abbd86b49a5c4:

> Sometimes there exist multiple backend vaapi drivers on one system. In such case user hopes to load the specific driver based on the request. Currently it can be specified by using LIBVA_DRIVER_NAME environment. But it is not convenient.

On a system where iHD va-api driver is installed and i965 is not, any va app will crash with `unknown libva error`. It works correctly when the `LIBVA_DRIVER_NAME` is set to `iHD` manually.

First guess was the lookup in `driver_name_map` (va/drm/va_drm_utils.c), but it will actually use `va-DRI2GetDriverName` and end up with `i965` from there instead of `iHD`.

Under wayland, the according wayland_drm code path is used and the detection works when the drm driver mapping is changed.

azhar904:
Add mapping of i915 kernel driver to Intel media driver
vaInitialize succeeds on systems with iHD driver without
having to export LIBVA_DRIVER_NAME=iHD

Fixes #244

Signed-off-by: Azhar Shaikh <azhar.shaikh@intel.com>

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -37,6 +37,7 @@ struct driver_name_map {
 };
 
 static const struct driver_name_map g_driver_name_map[] = {
+    { "i915",       4, "iHD"    }, // Intel Media driver
     { "i915",       4, "i965"   }, // Intel OTC GenX driver
     { "pvrsrvkm",   8, "pvr"    }, // Intel UMG PVR driver
     { "emgd",       4, "emgd"   }, // Intel ECG PVR driver


fhvwy:
In terms of defaults, the vast majority of users only want decode support (for use in media players - vlc, mpv, whatever), and only a minority have hardware which supports the iHD driver.  For decode the drivers operate pretty similarly, while encoding is very much a niche use-case compared to decode.  So, requiring a configuration change (an option in the program to call vaSetDriverName(), or the env var to set it globally) to move from the working-everywhere i965 driver to the faster-for-encode but with more limited hardware support iHD driver isn't unreasonable for now.

Since the code in question in libva just blindly loads the first matching driver, having multiple in the list has no effect and most users would end up failing to load the iHD driver in their media player and likely then fall back to software decode with much lower performance and greater power use.  Therefore this patch is not a good idea in its current form.

On the other hand, we definitely will want to load iHD first on future devices (I assume i965 will never support gen10 fully or anything beyond it), so at least some changes will be needed to support that.  Some possible thoughts on what to do here:
* Query the device and choose iHD if it's a gen10 or greater, otherwise i965.  (The default for gen9 should probably stay as i965 for backward compatibility since users who are already using it could hit problems on a change, especially if iHD isn't installed by their distribution?  Not sure about that part.)
* Look at what drivers are actually available in the driver directory, and load i965 first if present, otherwise iHD.  (This doesn't solve the future-device problem well, but the workaround is very easy because the user can just install one driver matching their hardware.)
* Add support for loading multiple different drivers, and when no override is set try them in turn from the list until one works.  (In this case iHD can probably be tried first.)

chivakker:
@XinfengZhang can you comment on a plan to move from i965 to iHD in the future as the default driver of use? 

iHD become the default should be a result of a good planned strategy which belongs to the maintainer team.  

If I may contribute an idea, I'd prefer to deprecate LIBVA_DRIVER_NAME in favor of only having vaGet/vaSetDriverName and adding a Query option which libva can decide based on the underlying hardware, A preferred driver could be decided inside the driver and return error when for example a user wants to set i965 on gen10 and beyond.

uartie:
> If I may contribute an idea, I'd prefer to deprecate LIBVA_DRIVER_NAME in favor of only having vaGet/vaSetDriverName and adding a Query option which libva can decide based on the underlying hardware, A preferred driver could be decided inside the driver and return error when for example a user wants to set i965 on gen10 and beyond.

Wait, we still need to support the user in having the final say of which driver to load at runtime if they choose.  Currently LIBVA_DRIVER_NAME provides that mechanism.  This is especially useful for developers or automated tests to toggle driver as needed when both are installed.

XinfengZhang:
@chivakker hi Daniel, no detail plan to  move from i965 to iHD. but i965 will not support future platform, only basic support to resolve critical issue. iHD will support from BDW and Gen9+. users can choose any one with overlap platforms

FurongZhang:
I also met this problem. In my case, I want to use "iHD", but "iHD" is not in the list. I am also trying to fix this issue. So now, any plan to check in or fix this issue? Thanks a in advance! @azhar904

XinfengZhang:
merge it, because iHD means future platform support.

ramcq:
@XinfengZhang @azhar904 @chivakker This patch isn't really correct, the VA_DRM_GetDriverName function returns the first g_driver_name_map entry to match the current DRM driver. So, on any "i915" system, only "iHD" VA driver will be tried, and initialising VAAPI will fail if this driver is not available. There is no fallback or check that the driver is available, so having a 2nd struct entry for the "i965" VA driver in the code is inoperative and misleading. The simpler / more clear patch would just - the i965 driver and + the iHD driver.

PR #248, opened 2018-11-10
x11: Remove obsolete B16 & B32 tags in protocol struct definitions

alanc:
They were defined as empty strings on all platforms except for the
long unsupported Cray systems which needed to use bitfields to define
types smaller than 64-bits.

Signed-off-by: Alan Coopersmith <alan.coopersmith@oracle.com>

### DIFF
--- a/va/x11/va_dri2str.h
+++ b/va/x11/va_dri2str.h
@@ -56,171 +56,171 @@
 #define X_DRI2SwapInterval      12
 
 typedef struct {
-    CARD32  attachment B32;
-    CARD32  name B32;
-    CARD32  pitch B32;
-    CARD32  cpp B32;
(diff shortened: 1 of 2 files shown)


fhvwy:
Rather than modifying an old version of the file, wouldn't it be cleaner to import a new version of the upstream file which has this fixed?  (<https://gitlab.freedesktop.org/xorg/proto/xorgproto/blob/master/include/X11/extensions/dri2proto.h>)

XinfengZhang:
hi @alanc, will you copy latest one?

alanc:
Sorry - I didn't realize you were shipping an old copy of the file from X.Org - why not just depend on that file instead of keeping your own copy that you need to resync with upstream?

PR #256, opened 2018-12-07
Add definition for input/output surface flag

Jister:


XinfengZhang:
changes  of input_surface_flag & output_surface_flag  is not backward compatible, it move the "protected" flag from lowest bit to highest bit.   I wonder whether there are user is using the "protected" surface , if no real usage, I am ok to change it. but if there are usage, we cant change it. @xhaihao @fhvwy @chivakker  , could you help to confirm whether there are usage, from i965, FFmpeg, gstreamer view?

XinfengZhang:
if no other concern, I will merge it in 1 week , thanks

XinfengZhang:
@fhvwy @ceyusa  do you have concern about it?

ceyusa:
There is no use of the protection bit in gstreamer-vaapi. Lgtm :)

PR #257, opened 2018-12-07
Add pointer to struct wl_interface implementation for driver to use

xhaihao:
See https://github.com/intel/intel-vaapi-driver/issues/419 for the
information

Note: don't export the interface symbol in case others may reuse it.


jwang11:
LGTM

PR #261, opened 2018-12-21
Add va call for setting and getting context priority

shiqiangyu:
Signed-off-by: Yu Shiqiang <shiqiang.yu@intel.com>


dvrogozh:
How protected interface is related to context priority? And we discuss protected interface in #457, please, don't duplicate the effort.

dvrogozh (changes requested):
ditto

PR #266, opened 2019-01-14
Allow (and use by default) ORIGIN-relative driver paths

rcombs:
This behaves similarly to the `$ORIGIN` feature in the `RPATH` and `RUNPATH` ELF headers. It allows libva and the driver modules to be directory-portable (without having to set the env var at runtime), since libva only needs the path to the library dir relative to its own install dir, rather than an absolute path.

This retains compatibility with existing drivers' build systems by setting the pkgconfig driverdir value, as with the existing code.

As far as I can tell, `dladdr` is available on all targeted systems.

I was initially concerned that using `dladdr` this way might cause problems for static linking, but that's not supported in libva anyway.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -157,10 +157,14 @@ AC_ARG_ENABLE([va-messaging],
 AC_ARG_WITH(drivers-path,
     [AC_HELP_STRING([--with-drivers-path=[[path]]],
                     [drivers path])],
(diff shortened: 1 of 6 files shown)


rcombs:
Anyone have any opinions on this?

rcombs:
Ping; can anyone take a look at this?

PR #267, opened 2019-01-14
va_openDriver: check strdup return value

xuguangxin:
strdup may return NULL when out of memory.
If we pass this NULL to strtok_r,it returns unflushed token from previous
strtok_r call. It's unexpected result.


XinfengZhang:
LGTM

Issue #265, opened 2019-01-11
Improper uint64_t size for VAEncMiscParameterQuantization


fulinjie:
Improper uint64_t size for VAEncMiscParameterQuantization.
If define a struct in FFmpeg,
struct {
        VAEncMiscParameterBuffer misc; //uint32_t
        VAEncMiscParameterQuantization quantization;//uint64_t
} quantization_params;

misc->data will point to empty data instead of quantization because of the memory alignment, and driver could not got the correct value passed down from FFmpeg.

Source code:
https://github.com/intel/libva/blob/382acf177ce18c069e0293408afa34c5875296ff/va/va.h#L2177

fulinjie:
Thanks for the quick response. 
It will be great the have it fixed in libva level if available.

PR #272, opened 2019-01-21
Adjust VAEncMiscParameterQuantization structure to be align with VAEn…

XinfengZhang:
…cMiscParameterBuffer

Fixes:#265

Signed-off-by: intel <carl.zhang@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -2214,17 +2214,18 @@ typedef struct _VAEncMiscParameterQuantization
         struct
         {
 	    /* \brief disable trellis for all frames/fields */
-            uint64_t disable_trellis : 1;
+            uint32_t disable_trellis : 1;
 	    /* \brief enable trellis for I frames/fields */
-            uint64_t enable_trellis_I : 1;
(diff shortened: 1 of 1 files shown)


XinfengZhang:
@lizhong1008 yes, it will in 2.5 if there are no other concern, I am wondering whether it will break ABI

XinfengZhang:
mark it with "help wanted" , this PR potentially will break backward compatible. but it is a fix. previous implementation will have some issue when user use it like #265 , please help to review, if no other concern , I will merge it in 2 weeks.

fulinjie:
Ping?

XinfengZhang:
@Erick555 hi , is libabigail will report issues with this changes? if freedesktop-sdk accept it?

Erick555:
@XinfengZhang I don't know. We're in preparation for a new major release and we won't check ABI until it's finished. At the same time current stable release is more or less frozen, perhaps we won't update libva there anymore. That means we won't complain regardless if this trigger libabigail warning or not :smiley:.

Issue #275, opened 2019-01-25
meson lacks ENABLE_VA_MESSAGING


heftig:
The meson build provides no option to define `ENABLE_VA_MESSAGING`, unlike the autotools one.

PR #276, opened 2019-01-25
build: meson: enables va messaging and visibility

ceyusa:
This patch syncs with the defined compiler flags in autotools: the
enable of VA messaging and the compiler's support to visibility
attribute.

Fixes: #275

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -108,6 +108,15 @@ if get_option('with_wayland') != 'no'
   WITH_WAYLAND = wayland_dep.found()
 endif
 
+va_c_args = []
+if get_option('enable_va_messaging')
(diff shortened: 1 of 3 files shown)


ceyusa:
too late for 2.4.0 ?????

XinfengZhang:
@ceyusa   is it important for 2.4.0?  why?

ceyusa:
> @ceyusa is it important for 2.4.0? why?

It is not :)

Just the release took me by surprise when I was working on this

XinfengZhang:
@ceyusa sorry about it, we plan to release 2.4.0 at last month, but we are late, so ....  :)

PR #279, opened 2019-02-19
libva 2.4.0

htpc-ubuntu-linux:
Signed-off-by: intel <carl.zhang@intel.com>


XinfengZhang:
@htpc-ubuntu-linux  , prepare to bump mater the version to 2.5.0.pre1. is it ok?

XinfengZhang:
such like #283

PR #285, opened 2019-03-13
va: update vp9 header

ankuljai:
Signed-off-by: Ankul Jain <ankul.k.jain@intel.com>


sreerenjb:
What do you mean by this patch ? :) 
Whatever be the intention, it is not allowed to break the API like this. Please close the bug.

PR #293, opened 2019-04-09
configure: don't build glx if VA/X11 isn't built

xhaihao:


xhaihao:
Previous glx is still built even if glx is disabled in the command line 
./autogen.sh --enable-x11=no --enable-glx=no

PR #295, opened 2019-04-17
Remove unnecessary linux/videodev2.h include

rcombs:
Nothing in this header is actually used for anything, and this simplifies building for FreeBSD.

### DIFF
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -30,7 +30,6 @@
 #define _VA_BACKEND_H_
 
 #include <va/va.h>
-#include <linux/videodev2.h>
 
 typedef struct VADriverContext *VADriverContextP;
 typedef struct VADisplayContext *VADisplayContextP;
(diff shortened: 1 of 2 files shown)


XinfengZhang:
@rcombs sorry, I still have a question about time.h to be sys/time.h. how about android and other os?

rcombs:
`gettimeofday` is in `<sys/time.h>` regardless of platform (check its man page). `linux/videodev2.h` happens to `#include <sys/time.h>`, which is why there wasn't an error there before.

XinfengZhang:
if so, sys/time.h should already cover time.h, right?

rcombs:
`<sys/time.h>` doesn't include `<time.h>`, or vice versa.

XinfengZhang:
@rcombs  thanks, from the va_trace.c, there are only gettimeofday, why we still include <time.h>

rcombs:
Ah, sorry, I misunderstood. Yes, there's no reason for that file to include `<time.h>`; re-pushed with that removed.

rcombs:
Anything blocking this from merging?

PR #302, opened 2019-05-22
Added 'make rpm' target

troyev:
Hello,

I added an option to run "make rpm" after the build and generate an rpm file, which would be handy for us CentOS/RHEL users. Please take a look and let me know what you think.

Thanks,
Troy

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -45,7 +45,7 @@ rpm: build/$(PACKAGE).spec
 	./va/.rpmbuild/SRPMS \
 	./va/.rpmbuild/RPMS/noarch \
 	./va/.rpmbuild/tmp
-	make dist-gzip
+	tar cf $(PACKAGE)-$(VERSION).tar.gz . --transform 's,^,$(PACKAGE)-$(VERSION)/,'
 	mv $(PACKAGE)-$(VERSION).tar.gz ./va/.rpmbuild/SOURCES
 	rpmbuild --define "_topdir $$PWD/va/.rpmbuild" \
 	 --define "_tmppath $$PWD/va/.rpmbuild/tmp" \
--- a/Makefile.am
+++ b/Makefile.am
@@ -45,15 +45,13 @@ rpm: build/$(PACKAGE).spec
 	./va/.rpmbuild/SRPMS \
 	./va/.rpmbuild/RPMS/noarch \
 	./va/.rpmbuild/tmp
-	git archive --format=tar.gz --prefix=$(PACKAGE)-$(VERSION)/ -o ./$(PACKAGE)-$(VERSION).tar.gz HEAD
-	cp -f $(PACKAGE)-$(VERSION).tar.gz ./va/.rpmbuild/SOURCES
+	make dist-gzip
+	mv $(PACKAGE)-$(VERSION).tar.gz ./va/.rpmbuild/SOURCES
 	rpmbuild --define "_topdir $$PWD/va/.rpmbuild" \
 	 --define "_tmppath $$PWD/va/.rpmbuild/tmp" \
 	 --define "_sourcefile $(PACKAGE)-$(VERSION).tar.gz" \
 	 --define "_version $(VERSION)" \
-	 --define "_sourcefile $(PACKAGE)-$(VERSION).tar.gz" \
 	 -ba $(top_srcdir)/build/$(PACKAGE).spec
 	@echo "============================================================"
 	@echo "RPMs located in ./va/.rpmbuild/RPMS and ./va/.rpmbuild/SRPMS"
 endif
-
--- a/Makefile.am
+++ b/Makefile.am
@@ -35,3 +35,25 @@ MAINTAINERCLEANFILES = \
 	aclocal.m4 compile config.guess config.sub \
 	configure depcomp install-sh ltmain.sh     \
 	Makefile.in missing
+
+if HAVE_RPMBUILD
+rpm: build/$(PACKAGE).spec
+	rm -rf ./va/.rpmbuild
+	mkdir -p ./va/.rpmbuild/BUILD \
+	./va/.rpmbuild/SPECS \
+	./va/.rpmbuild/SOURCES \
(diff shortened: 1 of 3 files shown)


build/libva.spec
```diff
@@ -0,0 +1,39 @@
...
+make install DESTDIR=%{buildroot}
+#mkdir -p %{buildroot}/usr/bin/
+#install -m 755 hello-world.sh %{buildroot}/usr/bin/hello-world.sh
+
+%files
+/usr/local/lib/*.so*
+%files devel
+/usr/local/include/va/*.h
+/usr/local/lib/pkgconfig/*.pc
+/usr/local/lib/*.la
```
yangnin: Why do u use local here?

Makefile.am
```diff
@@ -35,3 +35,25 @@ MAINTAINERCLEANFILES = \
...
+if HAVE_RPMBUILD
+rpm: build/$(PACKAGE).spec
+	rm -rf ./va/.rpmbuild
+	mkdir -p ./va/.rpmbuild/BUILD \
+	./va/.rpmbuild/SPECS \
+	./va/.rpmbuild/SOURCES \
+	./va/.rpmbuild/SRPMS \
+	./va/.rpmbuild/RPMS/noarch \
+	./va/.rpmbuild/tmp
+	git archive --format=tar.gz --prefix=$(PACKAGE)-$(VERSION)/ -o ./$(PACKAGE)-$(VERSION).tar.gz HEAD
```
yangnin: if git is not installed, Does it still work?

build/libva.spec
troyev: I changed it to be more generic

Makefile.am
troyev: Ok, changing to use "make dist-gzip" instead of git

troyev: "make dist-gzip" looks like it fails twice, then succeeds. Seems to be generating files related to wayland. Looking into how to make it work on the first try

troyev: Ok, I switched to using tar. Do you think it is safe to assume that tar is available?

build/libva.spec
```diff
@@ -0,0 +1,37 @@
...
+%description
+Intel libva
+
+%prep
+
+%setup
+
+%build
+./configure --prefix %{_prefix} --libdir %{_libdir}
+make -j12
```
yangnin: it 's better not directly use a number here. You could let script to detect the cpu core numbers

troyev: Ok, updated to "make -j$(nproc)"

XinfengZhang:
@troyev I tried the patch:
make rpm
rm -rf ./va/.rpmbuild
mkdir -p ./va/.rpmbuild/BUILD \
./va/.rpmbuild/SPECS \
./va/.rpmbuild/SOURCES \
./va/.rpmbuild/SRPMS \
./va/.rpmbuild/RPMS/noarch \
./va/.rpmbuild/tmp
tar cf libva-2.5.0.pre1.tar.gz . --transform 's,^,libva-2.5.0.pre1/,'
tar: ./libva-2.5.0.pre1.tar.gz: file is the archive; not dumped
mv libva-2.5.0.pre1.tar.gz ./va/.rpmbuild/SOURCES
rpmbuild --define "_topdir $PWD/va/.rpmbuild" \
 --define "_tmppath $PWD/va/.rpmbuild/tmp" \
 --define "_sourcefile libva-2.5.0.pre1.tar.gz" \
 --define "_version 2.5.0.pre1" \
 -ba ./build/libva.spec
/bin/sh: rpmbuild: command not found
Makefile:875: recipe for target 'rpm' failed
make: *** [rpm] Error 127

ps: should explicit add "HAVE_RPMBUILD=true" right? is it possible to enable it just like make -dist-bzip2?

troyev:
@XinfengZhang 

Thanks for giving it a try.
The HAVE_RPMBUILD variable should be set automatically by configure.ac depending on whether you have "rpmbuild" installed on your machine. Can you install rpmbuild,  and re-run ./autogen.sh; make rpm? On CentOS/RedHat I do a "sudo yum install rpm-build" to install it. I imagine Ubuntu has a similar package.


configure.ac:
```
AC_CHECK_PROG(HAVE_RPMBUILD, rpmbuild, true, false)
AM_CONDITIONAL(HAVE_RPMBUILD,$HAVE_RPMBUILD)
```

XinfengZhang:
@XinfengZhang it works, I merged it, but I forgot one question, why you put it in va folder instead in top folder?

troyev:
@XinfengZhang 
Thanks! I put the output in the va folder because I saw that is where the shared libraries are created. When I run make I see va/.libs/libva.so. I was just trying to follow the pattern of where other files are created by putting it in va/.rpmbuild. you can move it around if you want.

dvrogozh:
Unfortunately I did not see this PR before. Why this feature is really needed? Distros create their own packaging spec files and tend to ignore whatever exists in the repo. Moreover, this packaging support will start to conflict with whatever is in the distros. For redhat family, i.e. fedora, centos, redhat, why you can't reuse fedora libva packaging: https://src.fedoraproject.org/rpms/libva/tree/master?

XinfengZhang:
@dvrogozh , @troyev  seems there are some concerns . so , could we force push it then reopen the PR to continue the discussion ?

troyev:
@XinfengZhang @dvrogozh 

Ok, thanks for the link. Feel free to revert the changes and/or close the request. I will look into using the fedora spec file as an alternative.

The reason I made the PR for this repo and MediaSDK was to be consistent with the other two Intel libraries "media-driver" and "gmmlib" which already have the ability to make an rpm from inside the repo. However, I can understand wanting to eliminate redundant packaging code.

dvrogozh:
@troyev : it would be really helpful if you will take a look into fedora packaging and feedback here whether you see some issues, what did and what did not work for you.

XinfengZhang:
@troyev , I revert it, if you want to change it or reopen the PR, please file new one, of course, we can continue discussion under this PR

PR #315, opened 2019-06-17
Add comments for colour primaries and transfer characteristics.

FurongZhang:
For colour_primaries and transfer_characteristics, recommend app to define in enum or macro definition according to Spec.
In the comments, we just list the typical value for the reference.

Signed-off-by: furongzh <furong.zhang@intel.com>


XinfengZhang:
LGTM

zhoujd:
LGTM

PR #317, opened 2019-06-18
Add av1 decode interfaces

ZefuLI:
add av1 decoding api

### DIFF
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -1,0 +1,656 @@
+/*
+ * Copyright (c) 2019 Intel Corporation. All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sub license, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL INTEL AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * \file va_dec_av1.h
+ * \brief The AV1 decoding API
+ *
+ * This file contains the \ref api_dec_av1 "AV1 decoding API".
+ */
+
+#ifndef VA_DEC_AV1_H
+#define VA_DEC_AV1_H
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * \brief AV1 Decoding Picture Parameter Buffer Structure
+ *
+ * This structure conveys picture level parameters.
+ * App should send a surface with this data structure down to VAAPI once
+ * per frame.
+ *
+ */
+
+/** \brief Segmentation Information
+  */
+typedef struct _VASegmentationStructAV1 {
+    union {
+        struct {
+             /** Indicates whether segmentation map related syntax elements 
+             *  are present or not for current frame. If equal to 0, 
+             *  the segmentation map related syntax elements are 
+             *  not present for the current frame and the control flags of 
+             *  segmentation map related tables feature_data[][], and 
+             *  feature_mask[] are not valid and shall be ignored by accelerator.
+             */
+            uint32_t         enabled                                     : 1;
+            /** Value 1 indicates that the segmentation map are updated 
+             *  during the decoding of this frame. 
+             *  Value 0 means that the segmentation map from the previous 
+             *  frame is used.
+             */
+            uint32_t         update_map                                  : 1;
+            /** Value 1 indicates that the updates to the segmentation map 
+             *  are coded relative to the existing segmentation map. 
+             *  Value 0 indicates that the new segmentation map is coded 
+             *  without reference to the existing segmentation map.
+             */
+            uint32_t         temporal_update                             : 1;
+            /** Value 1 indicates that new parameters are about to be 
+             *  specified for each segment.
+             *  Value 0 indicates that the segmentation parameters 
+             *  should keep their existing values.
+             */
+            uint32_t         update_data                                 : 1;
+            
+            /** \brief Reserved bytes for future use, must be zero */
+            uint32_t         reserved                                    : 28;
+        } bits;
+        uint32_t             value;
+    } segment_info_fields;
+
+    /** \brief Segmentation parameters for current frame.
+     *  feature_data[segment_id][feature_id] 
+     *  where segment_id has value range [0..7] indicating the segment id.
+     *  and feature_id is defined as
+        typedef enum {
+            SEG_LVL_ALT_Q,       // Use alternate Quantizer ....
+            SEG_LVL_ALT_LF_Y_V,  // Use alternate loop filter value on y plane vertical
+            SEG_LVL_ALT_LF_Y_H,  // Use alternate loop filter value on y plane horizontal
+            SEG_LVL_ALT_LF_U,    // Use alternate loop filter value on u plane
+            SEG_LVL_ALT_LF_V,    // Use alternate loop filter value on v plane
+            SEG_LVL_REF_FRAME,   // Optional Segment reference frame
+            SEG_LVL_SKIP,        // Optional Segment (0,0) + skip mode
+            SEG_LVL_GLOBALMV,
+            SEG_LVL_MAX
+        } SEG_LVL_FEATURES;
+     *  feature_data[][] is equivalent to variable FeatureData[][] in spec, 
+     *  which is after clip3() operation. 
+     *  Clip3(x, y, z) = (z < x)? x : ((z > y)? y : z);
+     *  The limit is defined in Segmentation_Feature_Max[ SEG_LVL_MAX ] = {
+     *  255, MAX_LOOP_FILTER, MAX_LOOP_FILTER, MAX_LOOP_FILTER, MAX_LOOP_FILTER, 7, 0, 0 }
+     */
+    int16_t                 feature_data[8][8];
+
+    /** \brief indicates if a feature is enabled or not. 
+     *  Each bit field itself is the feature_id. Index is segment_id.
+     *  feature_mask[segment_id] & (1 << feature_id) equal to 1 specify that the feature of  
+     *  feature_id for segment of segment_id is enabled, otherwise disabled.
+     */
+    uint8_t                 feature_mask[8];
+
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                va_reserved[VA_PADDING_LOW];
+
+} VASegmentationStructAV1;
+
+/** \brief Film Grain Information
+  */
+typedef struct _VAFilmGrainStructAV1 {
+    union {
+        struct {
+            /** \brief Specify whether or not film grain is applied on current frame.
+             *  If set to 0, all the rest parameters should be set to zero
+             *  and ignored.
+             */
+            uint32_t        apply_grain                                 : 1;
+            uint32_t        chroma_scaling_from_luma                    : 1;
+            uint32_t        grain_scaling_minus_8                       : 2;
+            uint32_t        ar_coeff_lag                                : 2;
+            uint32_t        ar_coeff_shift_minus_6                      : 2;
+            uint32_t        grain_scale_shift                           : 2;
+            uint32_t        overlap_flag                                : 1;
+            uint32_t        clip_to_restricted_range                    : 1;
+            /** \brief Reserved bytes for future use, must be zero */
+            uint32_t        reserved                                    : 20;
+        } bits;
+        uint32_t            value;
+    } film_grain_info_fields;
+
+    uint16_t                grain_seed;
+    /*  value range [0..14] */
+    uint8_t                 num_y_points;
+    uint8_t                 point_y_value[14];
+    uint8_t                 point_y_scaling[14];
+    /*  value range [0..10] */
+    uint8_t                 num_cb_points;
+    uint8_t                 point_cb_value[10];
+    uint8_t                 point_cb_scaling[10];
+    /*  value range [0..10] */
+    uint8_t                 num_cr_points;
+    uint8_t                 point_cr_value[10];
+    uint8_t                 point_cr_scaling[10];
+    /*  value range [-128..127] */
+    int8_t                  ar_coeffs_y[24];
+    int8_t                  ar_coeffs_cb[25];
+    int8_t                  ar_coeffs_cr[25];
+    uint8_t                 cb_mult;
+    uint8_t                 cb_luma_mult;
+    uint16_t                cb_offset;
+    uint8_t                 cr_mult;
+    uint8_t                 cr_luma_mult;
+    uint16_t                cr_offset;
+
+    /** \brief Reserved bytes for future use, must be zero */
+    uint32_t                va_reserved[VA_PADDING_LOW];
+
+} VAFilmGrainStructAV1;
+
+typedef enum {
+    /** identity transformation, 0-parameter */
+    VAAV1Identity           = 0,
+    /** translational motion, 2-parameter */
+    VAAV1Translation        = 1,
+    /** simplified affine with rotation + zoom only, 4-parameter */
+    VAAV1Rotzoom            = 2,
+    /** affine, 6-parameter */
+    VAAV1Affine             = 3,
+    VAAV1Trans_Types,
+} VAAV1TransformationType;
+
+typedef struct _VAWarpedMotionParamsAV1{
+
+    /** \brief Specify the type of warped motion */ 
+    VAAV1TransformationType  wmtype;
+
+    /** \brief Specify warp motion parameters
+     *  wm.wmmat[] corresponds to gm_params[][] in spec. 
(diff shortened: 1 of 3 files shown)


va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+     */
+    uint8_t                 tile_cols;
+    uint8_t                 tile_rows;
+    uint16_t                width_in_sbs_minus_1[63];
+    uint16_t                height_in_sbs_minus_1[63];
+
+    /** \brief number of tiles in large scale tile list
+     *  Valid only when large_scale_tiles == 1.
+     */
+    uint16_t                tile_count_minus_1;
```
xhaihao: Is this field set to 'number of tiles' - 1? If yes, please change the  comment

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+    uint8_t                 order_hint;
+
+    VASegmentationStructAV1 seg_info;
+    VAFilmGrainStructAV1    film_grain_info;
+
+    /** \brief tile structure
+     *  When uniform_tile_spacing_flag == 1, the value of tile_cols and
+     *  tile_rows need to be power of 2, and width_in_sbs_minus_1[] and
+     *  height_in_sbs_minus_1[] should be ignored, which will be generated
+     *  by driver based on tile_cols and tile_rows.
```
xhaihao: What's sbs? could you add some comment?

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+            uint32_t        allow_high_precision_mv                     : 1;
+            uint32_t        is_motion_mode_switchable                   : 1;
+            uint32_t        use_ref_frame_mvs                           : 1;
+            /* disable_frame_end_update_cdf is coded as refresh_frame_context. */
+            uint32_t        disable_frame_end_update_cdf                : 1;
+            uint32_t        uniform_tile_spacing_flag                   : 1;
+            uint32_t        allow_warped_motion                         : 1;
+            /** \brief indicate if current frame in large scale tile mode */
+            uint32_t        large_scale_tile                            : 1;
+            uint32_t        reserved                                    : 15;
```
xhaihao: Please add comment to make sure reserved bits must be 0

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+ * sent once per tile.
+ *
+ * Bit stream data buffer of VASliceDataBufferType is used
+ * to send the bitstream.
+ *
+ * Please note that host decoder is responsible to parse out the
+ * per tile information. And the bit stream in sent to driver in per
+ * tile granularity.
+ */
+typedef struct _VABitStreamParameterBufferAV1
```
xhaihao: How about renaming BitStream with Slice because VASliceDataBufferType is also used to send the bit stream data?

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+/** \brief Segmentation Information
+  */
+typedef struct _VASegmentationStructAV1 {
+    union {
+        struct {
+            uint8_t         enabled                                     : 1;
+            uint8_t         update_map                                  : 1;
+            uint8_t         temporal_update                             : 1;
+            uint8_t         update_data                                 : 1;
+            uint8_t         reserved                                    : 4;
```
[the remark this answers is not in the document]
ZefuLI: sure, I will add comment for all reserved bits.

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+        } bits;
+        uint8_t             value;
```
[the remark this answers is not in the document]
ZefuLI: will change it, thanks

va/va_dec_av1.h
ZefuLI: yes, it should be number of tiles minus 1.

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+    /** simplified affine with rotation + zoom only, 4-parameter */
+    VAROTZOOM     = 2,
+    /** affine, 6-parameter */
+    VAAFFINE      = 3,
+    VATRANS_TYPES,
+} VATransformationType;
+
+typedef struct _VAWarpedMotionParamsAV1{
+    VATransformationType    wmtype;
+    int32_t                 wmmat[8];
```
xuguangxin: It's no wmmat in spec. So how I fill these fields? What is the mean of 8?
If possible, please keep the name in spec.
If not possible, it's better to have some document point to spec.

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+              *  0:     KEY_FRAME;
+              *  1:     INTER_FRAME;
+              *  2:     INTRA_ONLY_FRAME;
+              *  3:     SWITCH_FRAME
+              *  For SWITCH_FRAME, application shall set error_resilient_mode = 1,
+              *  refresh_frame_flags, etc. appropriately. And driver will convert it
+              *  to INTER_FRAME.
+              */
+            uint32_t        frame_type                                  : 2;
+            uint32_t        show_frame                                  : 1;
```
xuguangxin: why you need show_frame and showable_frame? The frame reordering and output should handle by up layer.

va/va_dec_av1.h
xuguangxin: Why we need this. is it equals to tile_cols * tile_rows  - 1?

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+     *  value range [-63..63].
+     */
+    int8_t                  mode_deltas[2];
+
+    /** \brief quantization
+     */
+    /** \brief Y AC index
+     *  value range [0..255]
+     */
+    uint16_t                base_qindex;
```
xuguangxin: why we nedd uint16_t if we know it in [0.. 255]

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+            uint32_t        log2_delta_lf_res                           : 2;
+
+            /** \brief CONFIG_LOOPFILTER_LEVEL
+             */
+            uint32_t        delta_lf_multi                              : 1;
+
+            /** \brief read_tx_mode
+             *  value range [0..6]
+             */
+            uint32_t        tx_mode                                     : 2;
```
xuguangxin: how we hold 6 in 2 bits. Is it [0..2]?

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+
+            /** \brief read_frame_reference_mode
+             */
+            uint32_t        reference_mode                              : 2;
```
xuguangxin: it's no read_frame_refence_mode in "Version 1.0.0 with Errata 1"
could you help check?

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+     */
+    uint32_t                bit_stream_data_offset;
+    /**
+     * see VA_SLICE_DATA_FLAG_XXX definitions
+     */
+    uint32_t                bit_stream_data_flag;
+
+    uint16_t                tile_row;
+    uint16_t                tile_column;
+    uint16_t                start_tile_idx;
```
xuguangxin: whey we need start_tile_idx, it's possible calc if from tile_fow tile_column?

va/va_dec_av1.h
cwang105: super blocks

va/va_dec_av1.h
ZefuLI: will follow your comment.

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+            uint8_t         update_map                                  : 1;
+            uint8_t         temporal_update                             : 1;
+            uint8_t         update_data                                 : 1;
+            uint8_t         reserved                                    : 4;
+        } bits;
+        uint8_t             value;
+    } segment_info_fields;
+
+    int16_t                 feature_data[8][8];
+    uint8_t                 feature_mask[8];
```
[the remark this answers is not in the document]
ZefuLI: Accept, I will use Macros to replace absolute value

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+    uint8_t                 point_cr_value[10];
+    uint8_t                 point_cr_scaling[10];
+    /*  value range [-128..127] */
+    int8_t                  ar_coeffs_y[24];
+    int8_t                  ar_coeffs_cb[25];
+    int8_t                  ar_coeffs_cr[25];
+    uint8_t                 cb_mult;
+    uint8_t                 cb_luma_mult;
+    /*  value range [0..512] */
+    uint8_t                 cb_offset;
```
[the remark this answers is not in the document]
ZefuLI: Thanks for pointing out it, will fix it

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+    uint8_t                 cr_mult;
+    uint8_t                 cr_luma_mult;
+    /*  value range [0..512] */
+    uint8_t                 cr_offset;
```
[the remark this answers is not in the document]
ZefuLI: same as above.

va/va_dec_av1.h
ZefuLI: yes, I will fix it

va/va_dec_av1.h
ZefuLI: This is one tile list semantic for large scale specifically, it is equal to the number of tile_list_entry in the list minus 1.

va/va_dec_av1.h
ZefuLI: I will fix it, thanks

va/va_dec_av1.h
ZefuLI: I checked aom reference code, and found read_frame_reference_mode is there. however in Spec, it is named as  frame_reference_mode.

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+    /** identity transformation, 0-parameter */
+    VAIDENTITY    = 0,
+    /** translational motion 2-parameter */
+    VATRANSLATION = 1,
+    /** simplified affine with rotation + zoom only, 4-parameter */
+    VAROTZOOM     = 2,
+    /** affine, 6-parameter */
+    VAAFFINE      = 3,
+    VATRANS_TYPES,
+} VATransformationType;
```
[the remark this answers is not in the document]
ZefuLI: Accept.

va/va_dec_av1.h
ZefuLI: It is the index of the first tile in the tile group of which the current tile belongs to. The index is in raster scan order of all tiles in frame. For the first tile group, this value should be 0. This parameter can be used to uniquely identify the tile group.

va/va_dec_av1.h
```diff
@@ -0,0 +1,546 @@
...
+            uint32_t        ar_coeff_shift_minus_6                      : 2;
+            uint32_t        grain_scale_shift                           : 2;
+            uint32_t        overlap_flag                                : 1;
+            uint32_t        clip_to_restricted_range                    : 1;
+            uint32_t        reserved                                    : 20;
+        } bits;
+        uint32_t            value;
+    } film_grain_info_fields;
+
+    uint16_t                random_seed;
```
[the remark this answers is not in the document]
ZefuLI: random_seed is the name of aom reference code.

va/va_dec_av1.h
ZefuLI: yes, VASliceDataBufferType is used to send the bit stream data, which is same as other codecs

va/va_dec_av1.h
xuguangxin: ok, thanks for explain.

va/va_dec_av1.h
xuguangxin: This makes me more confused.  
Is reference_mode equals to reference_select in frame_reference_mode?Apparently, it's not. because  reference_select  has only 1 bit, but reference_mode has 2 bits.
Is it the same thing as comp_mode?  not likely, because comp_mode only has two value SINGLE_REFERENCE, COMPOUND_REFERENCE. 1 bits is enough for them.

IMO, we'd better stick name and document to spec. aom just one implement for spec. the name may change after spec finalized. and it may add an unnecessary definition for optimization/code simplify purpose. 
If we stick to spec, the one who implement hw codec only need to check the spec. not the aom code.

va/va_dec_av1.h
xuguangxin: @ZefuLI, any feedback for this?

thanks

va/va_dec_av1.h
xuguangxin: @ZefuLI, any feedback for this?

va/va_dec_av1.h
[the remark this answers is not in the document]
xuguangxin: thanks.
Some document for feature_mask is important too.  Else, nobody knows feature_mask[0] = 1 will enable the feature_data[0][0] or feature_data[0][7]

va/va_dec_av1.h
xuguangxin: could you explain more about this? Is equals to tg_start and tg_end?
If yes, why we need tile_row and tile_column?

va/va_dec_av1.h
ZefuLI: All right, we will stick name to spec, actually in aom code, reference mode has below types.
typedef enum {
  SINGLE_REFERENCE = 0,
  COMPOUND_REFERENCE = 1,
  REFERENCE_MODE_SELECT = 2,
  REFERENCE_MODES = 3,
} REFERENCE_MODE;
so we use 2 bits, however, we will swicth to spec defination <reference_select> and use 1 bit.

va/va_dec_av1.h
ZefuLI: they are tg_start and tg_end, and will use these names.

va/va_dec_av1.h
ZefuLI: I will remove show_frame and showable_frame.

va/va_dec_av1.h
ZefuLI: it should be [0..3].

va/va_dec_av1.h
ZefuLI: wm[7].wmmat[8] is corresponding to gm_params[ ref ][ j ] ( for ref = LAST_FRAME..ALTREF_FRAME, for j = 0..5.)  in spec, for wmmat[8], valid index value is [0..5], we use 8 here is to keep consistent with libaom.
Meanwhile, I will add commment to point to spec.

va/va_dec_av1.h
xuguangxin: The SINGLE_REFERENCE/COMPOUND_REFERENCE is for macroblock level(in read_ref_frames). But libva's user only handles stream in slice level(in frame_reference_mode). this why the caller can only give you a bool value(reference_select)

va/va_dec_av1.h
xuguangxin: we only have following tx mode, right?
0 ONLY_4X4
1 TX_MODE_LARGEST
2 TX_MODE_SELECT

va/va_dec_av1.h
xuguangxin: thanks

va/va_dec_av1.h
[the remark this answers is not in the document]
xuguangxin: yeah , if you do want use reference code name. please point it to spec.
thanks

va/va_dec_av1.h
ZefuLI: tile_row/tile_column is tile row/column index, which is frame level , we need it since AV1 is tile based decoding, we use tile index info to verify which the current tile it is.

va/va_dec_av1.h
ZefuLI: you are right,
Value range:
// frame transform mode
typedef enum {
  ONLY_4X4 = 0,     // only 4x4 transform used
  TX_MODE_LARGEST,  // transform size is the largest possible for pu size
  TX_MODE_SELECT,   // transform specified for each block
  TX_MODES,
} TX_MODE;

va/va_dec_av1.h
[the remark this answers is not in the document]
ZefuLI: > How about using uint32_t ?

sorry, it seems that 8 bits are enough for this value?

rosetta-jpn:
Sorry for interrupting.
When is the estimated time of checking in this patch?

sreerenjb:
As Haihao already mentioned, I would like to see the use of VASliceParameterBufferXXX instead of VABitStreamParameterBufferXXX and also 
%s/bit_stream_data_size/slice_data_size
%s/bit_stream_data_offset/slice_data_offset
%s/bit_stream_data_flag/slice_data_flag
Which will make life easier for the middleware implementations .Especially GStreamer using common interface for all codecs.

ZefuLI:
> As Haihao already mentioned, I would like to see the use of VASliceParameterBufferXXX instead of VABitStreamParameterBufferXXX and also
> %s/bit_stream_data_size/slice_data_size
> %s/bit_stream_data_offset/slice_data_offset
> %s/bit_stream_data_flag/slice_data_flag
> Which will make life easier for the middleware implementations .Especially GStreamer using common interface for all codecs.

Hi, you can see we are using VASliceParameterBufferAV1 in my latest update.

va/va_dec_av1.h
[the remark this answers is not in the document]
XinfengZhang: @ZefuLI  why not change it?

va/va_dec_av1.h
```diff
@@ -0,0 +1,653 @@
...
+    /*  value range [-128..127] */
+    int8_t                  ar_coeffs_y[24];
+    int8_t                  ar_coeffs_cb[25];
+    int8_t                  ar_coeffs_cr[25];
+    uint8_t                 cb_mult;
+    uint8_t                 cb_luma_mult;
+    uint16_t                cb_offset;
+    uint8_t                 cr_mult;
+    uint16_t                cr_luma_mult;
+    uint8_t                 cr_offset;
```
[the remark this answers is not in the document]
ZefuLI: @XinfengZhang 
The latest version I uploaded used uint8 uint8 uint16, like,
    uint8_t                 cb_mult;
    uint8_t                 cb_luma_mult;
    uint16_t               cb_offset;
    uint8_t                 cr_mult;
    uint8_t                 cr_luma_mult;
    uint16_t               cr_offset;

HeJunyan:
Need to add va_dec_av1.h to va/meson.build file's libva_headers.  We still miss that header file in system include dir when apply this patch.

uartie:
This breaks builds that include installed va/va.h:  `va/va.h:4630:10: fatal error: va/va_dec_av1.h: No such file or directory`.  Did you forget to add an install target for `va_dec_av1.h`?

HeJunyan:
yes, this patch forget adding va_dec_av1.h to va/meson.build, please help

uartie:
>yes, this patch forget adding va_dec_av1.h to va/meson.build, please help

Actually, the header is added to meson.build in this patch, afaict.  But It was not added to Makefile.am (we still use automake in CI).  I fixed the automake issue in #383.  Does the meson build not work for you?

PR #323, opened 2019-07-18
va/va_trace: Dump VP9 parameters for profile 1~3

xhaihao:


HeJunyan:
I have a try and it works OK for me.

PR #324, opened 2019-07-20
add multiple frame capability report

XinfengZhang:

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -773,6 +773,10 @@ typedef enum
      *  previous frame (PoC < current).
      */
     VAConfigAttribPredictionDirection   = 39,
+    /** \brief combined submission of multiple frames from different streams, it is optimization for different HW
+     * implementation, multiple frames encode/decode can improve HW concurrency
+     */
+    VAConfigAttribMultipleFrame         = 40,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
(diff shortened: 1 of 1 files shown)


va/va.h
```diff
@@ -773,6 +773,10 @@ typedef enum
      *  previous frame (PoC < current).
      */
     VAConfigAttribPredictionDirection   = 39,
+    /** \brief combined submission of mutiple frames from different stream,  it is optimization for different HW
```
dmitryermilov: multiple

dmitryermilov: streams

va/va.h
```diff
@@ -773,6 +773,10 @@ typedef enum
...
+     * implementation, mutiple frames encode/decode can improve HW concurrency
```
dmitryermilov: multiple

va/va.h
```diff
@@ -1111,6 +1115,21 @@ typedef union _VAConfigAttribValEncRateControlExt {
     uint32_t value;
 } VAConfigAttribValEncRateControlExt;
 
+/** \brief Attribute value for VAConfigAttribMultipleFrame*/
+typedef union _VAConfigAttribValMultipleFrame {
+    struct {
+        /** \brief max num of concurrent frames from different stream */
+        max_num_concurrent_frames      : 8;
+        /** \brief indicate whether MF support mixed codecs */
+        mixed_codecs                   : 1;
```
dmitryermilov: I believe it would be great if the new interface can say which combination of codec is supported.

va/va.h
```diff
@@ -1111,6 +1115,21 @@ typedef union _VAConfigAttribValEncRateControlExt {
...
+        /** \brief indicate whether all stream must support same quality level */
+        mixed_quality_level            : 1;
```
dmitryermilov: As as above.

va/va.h
XinfengZhang: see, but 32 bit is hard to describe it. I dont want to declare some Macro such as CODEC_H264 CODEC_HEVC  which can be OR'd together, because there are vaProfileXXXX already.

va/va.h
XinfengZhang: describe it , need 7 bits at least, and cant be common, I dont know how much quality levels is supported by other vendors

va/va.h
dmitryermilov: Then, IMO, it's useless because application won't be able to build a proper logic using  bool 'mixed_codecs'. In such case it's better to simply remove it.

XinfengZhang: thanks, agree, will remove it

Maggieie: hi Carl, 
last time we discussed to remove mixed_codecs for now, right? could you remove it?

va/va.h
Maggieie: we support any combination of mixed quality level. so will it be ok to just use one flag?

Issue #327, opened 2019-08-13
vatrace leaks memory


dvrogozh:
Used latest master:
* https://github.com/intel/libva/commit/f094de317b4f498b57d47d8b0823a2a24cb125e1
* https://github.com/intel/libva-utils/commit/81b5929f2b71d0b657e6ef605a522edc3bcd3e80

Enable trace with:
```
# cat /etc/libva.conf
LIBVA_TRACE=/home/tmp/file
```

Verify via vainfo on memory leaks:
```
[dvrogozh@dvrscl build]$ valgrind --tool=memcheck --leak-check=full vainfo
==4785== Memcheck, a memory error detector
==4785== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==4785== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==4785== Command: vainfo
==4785== 
libva info: Open new log file /home/dvrogozh/trace/file.140245.thd-0x000012b1 for the thread 0x000012b1
libva info: LIBVA_TRACE is on, save log into /home/dvrogozh/trace/file.140245.thd-0x000012b1
libva info: VA-API version 1.6.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'iHD'
libva info: Trying to open /usr/lib64/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_6
libva info: va_openDriver() returns 0
vainfo: VA-API version: 1.6 (libva 2.6.0.pre1)
vainfo: Driver version: Intel iHD driver - 1.0.0
vainfo: Supported profile and entrypoints
      VAProfileNone                   : VAEntrypointVideoProc
      VAProfileNone                   : VAEntrypointStats
      VAProfileMPEG2Simple            : VAEntrypointVLD
      VAProfileMPEG2Simple            : VAEntrypointEncSlice
      VAProfileMPEG2Main              : VAEntrypointVLD
      VAProfileMPEG2Main              : VAEntrypointEncSlice
      VAProfileH264Main               : VAEntrypointVLD
      VAProfileH264Main               : VAEntrypointEncSlice
      VAProfileH264Main               : VAEntrypointFEI
      VAProfileH264Main               : VAEntrypointEncSliceLP
      VAProfileH264High               : VAEntrypointVLD
      VAProfileH264High               : VAEntrypointEncSlice
      VAProfileH264High               : VAEntrypointFEI
      VAProfileH264High               : VAEntrypointEncSliceLP
[...]

PR #328, opened 2019-08-13
trace: fix memory leak on closing the trace

dvrogozh:
Fixes: #327

This removes redundunt check for the number of log files. The whole
'struct va_trace' is being calloc-d and we can just check that
pointers to log files are valid instead of relying on number of
active log files.


dvrogozh:
@XinfengZhang : can you, please, help review?

XinfengZhang:
LGTM, num never ++, and the log_file[] is circle.

PR #345, opened 2019-10-24
Pass offset and size of pred_weight_table

Tianhaol:
These two parameters are gotten from packed slice header, are used
to write back pred_weight_tabe into packed slice header
of the final bit stream.


XinfengZhang:
this change means: backend driver will calculate the weight table and send it back to packed header , then insert packed header to bitstream.  it will assume that application know it, application should insert blank weight table into it and tell driver the table start point and size. need extra interface to let application know this.

PR #347, opened 2019-11-20
Update va.h for multi-threaded usages

jonathanbian:
Added description for multi-threaded usages of the API

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -130,6 +130,93 @@ extern "C" {
  * 	- \ref api_fei
  * 	- \ref api_fei_h264
  * 	- \ref api_fei_hevc
+ * 
+ * \section threading Multithreading Guide
+ * All VAAPI functions implemented in libva are thread-safe. For any VAAPI
+ * function that requires the implementation of a backend (e.g. hardware driver),
+ * the backend must ensure that its implementation is also thread-safe. If the
+ * backend implementation of a VAAPI function is not thread-safe then this should
+ * be considered as a bug against the backend implementation. 
+ *
+ * It is assumed that none of the VAAPI functions will be called from signal 
+ * handlers.
+ *
+ * Thread-safety in this context means that when VAAPI is being called by multiple
+ * concurrent threads, it will not crash or hang the OS, and VAAPI internal
+ * data structures will not be corrupted. When multiple threads are operating on
+ * the same VAAPI objects, it is the application's responsibility to synchronize
+ * these operations in order to generate the expected results. For example, using
+ * a single VAContext from multiple threads may generate unexpected results.
+ *
+ * Following pseudo code illustrates a multithreaded transcoding scenario, where
+ * one thread is handling the decoding operation and another thread is handling 
+ * the encoding operation, while synchronizing the use of a common pool of
+ * surfaces.
+ *
+ * // Initialization
+ * dpy = vaGetDisplayDRM(fd);
+ * vaInitialize(dpy, ...); 
+ *
+ * // Create surfaces required for decoding and subsequence encoding
+ * vaCreateSurfaces(dpy, VA_RT_FORMAT_YUV420, width, height, &surfaces[0], ...);
+ *
+ * // Set up a queue for the surfaces shared between decode and encode threads
+ * surface_queue = queue_create();
+ *
+ * // Create decode_thread
+ * pthread_create(&decode_thread, NULL, decode, ...);
+ *
+ * // Create encode_thread
+ * pthread_create(&encode_thread, NULL, encode, ...);
+ *
+ * // Decode thread function
+ * decode() {
+ *   // Find the decode entrypoint for H.264 
+ *   vaQueryConfigEntrypoints(dpy, h264_profile, entrypoints, ...);
+ *
+ *   // Create a config for H.264 decode
+ *   vaCreateConfig(dpy, h264_profile, VAEntrypointVLD, ...);
+ *
+ *   // Create a context for decode
+ *   vaCreateContext(dpy, config, width, height, VA_PROGRESSIVE, surfaces, 
+ *     num_surfaces, &decode_context); 
+ *
+ *   // Decode frames in the bitstream
+ *   for (;;) {
+ *     // Parse one frame and decode 
+ *     vaBeginPicture(dpy, decode_context, surfaces[surface_index]); 
+ *     vaRenderPicture(dpy, decode_context, buf, ...);
+ *     vaEndPicture(dpy, decode_context);
+ *     // Poll the decoding status and enqueue the surface in display order after 
+ *     // decoding is complete
+ *     vaQuerySurfaceStatus();
+ *     enqueue(surface_queue, surface_index);
+ *   }
+ * }
+ *
+ * // Encode thread function
+ * encode() {
+ *   // Find the encode entrypoint for HEVC
+ *   vaQueryConfigEntrypoints(dpy, hevc_profile, entrypoints, ...);
+ *
+ *   // Create a config for HEVC encode
+ *   vaCreateConfig(dpy, hevc_profile, VAEntrypointEncSlice, ...);
+ *
+ *   // Create a context for encode
+ *   vaCreateContext(dpy, config, width, height, VA_PROGRESSIVE, surfaces,
+ *     num_surfaces, &encode_context); 
+ *
+ *   // Encode frames produced by the decoder
+ *   for (;;) {
+ *     // Dequeue the surface enqueued by the decoder    
+ *     surface_index = dequeue(surface_queue);
+ *     // Encode using this surface as the source
+ *     vaBeginPicture(dpy, encode_context, surfaces[surface_index]);
+ *     vaRenderPicture(dpy, encode_context, buf, ...);
+ *     vaEndPicture(dpy, encode_context);
+ *   }
+ * }
+ *
  */
 
 /**


va/va.h
```diff
@@ -129,6 +129,96 @@ extern "C" {
...
+ * responsibility of the driver to guarantee thread-safty of the backend
+ * functions. It is assumed that none of the VAAPI functions will be called from
+ * signal handlers.
+ *
+ * Thread-safty in this context means that when VAAPI is being called by multiple
+ * concurrent threads, it will not crash or hang the OS, and VAAPI internal
+ * states will not be corrupted. When multiple threads are operating on the same
+ * VAAPI objects, it it the application's responsibility to synchronize these
+ * operations in order to generate the expected results.
+ *
```
yellowdoge: These two paragraphs are confusing, because it says that `VAAPI functions are thread-safe` but then `it is the responsibility of the driver to guarantee thread-safty [sic]`, so: please rewrite the sections to reflect what you mean, e.g.: `VA-API internal objects are thread safe, but VA-API does not have nor impose or provide any guarantees of thread-safety on the underlying driver or backend, nor on any application-managed VA-API objects, including in particular VAContext `.

jonathanbian: Thanks for the comments and patch will be updated.

va/va.h
```diff
@@ -129,6 +129,99 @@ extern "C" {
  * 	- \ref api_fei
  * 	- \ref api_fei_h264
  * 	- \ref api_fei_hevc
+ * 
+ * \section threading Multithreading Guide
+ * All VAAPI functions implemented in libva should be thread-safe. However, 
+ * VAAPI and libva do not impose thread-safety requirements on the driver or
+ * the backend implementation, and it is the responsibility of the driver or the
+ * backend implementation to ensure thread-saftey of VAAPI functions. In other
```
yellowdoge: s/saftey/safety/

va/va.h
```diff
@@ -129,6 +129,99 @@ extern "C" {
...
+ * concurrent threads, it will not crash or hang the OS, and VAAPI internal
+ * states will not be corrupted. When multiple threads are operating on the same
+ * VAAPI objects, it is the application's responsibility to synchronize these
+ * operations in order to generate the expected results.
+ *
+ * It is worth noting that VAContext is not meant to be shared among different 
+ * threads. i.e if a VAContext object is shared among two or more threads, its 
+ * internal state may get corrupted which may lead to unexpected results for 
+ * operations using this VAContext. It is expected that each thread shall create
+ * and maintain its own VAContext object for all operations.
```
dvrogozh: > VAContext is not meant to be shared among different threads

I believe this is a too unclear statement. "Is not meant (or intended)" can be read as "could be and this is not prohibited according to the API". That's how I read it at least.

If we will transform that to the strict requirement, then I believe it is too strong requirement. I don't actually see why single VAContext can't be used from few threads. For example, it can be created in one thread, tasks submitted in another thread and it can be destroyed in yet another thread. What's wrong in such usage and why this can't be done as soon as operations are serialized?

I would suggested to be more specific and clear. Maybe we should better clarify whether it is application's responsibility to serialize vaBeginFrame-vaEndFrame sequencies? I.e. you can't have 2 concurrent vaBeginFrame-vaEndPicture sequences at a time. Having a single sequence serialized, but executed in different threads is likely OK.

rosetta-jpn:
I am confused about the purpose of this patch.
Looks like the patch clarifies that libva (i.e VA API level) does NOT guarantee thread-safety. It depends on backend and application implementation.
What I expected discussing with some intel persons is to clarify the backend must guarantee thread-safety, thus an application will not have to serialize VA-API calls if the application is implemented correctly.

Otherwise, gstreamer-vaapi's implementation can be regarded as wrong because it doesn't do anything for VA-API call serialization.

va/va.h
jonathanbian: Updated based on the feedbacks

va/va.h
```diff
@@ -129,6 +129,98 @@ extern "C" {
  * 	- \ref api_fei
  * 	- \ref api_fei_h264
  * 	- \ref api_fei_hevc
+ * 
+ * \section threading Multithreading Guide
+ * All VAAPI functions implemented in libva should be thread-safe, and the
```
yellowdoge: Please be careful with the wording. I'd recommend sticking to what https://www.ietf.org/rfc/rfc2119.txt suggests, and in particular the use of MUST and SHOULD. That said, this paragraph is misleading, because it starts saying "A is guaranteed" then "B should be guaranteed" but then says that if "B is not verified" then "A is not verified either".

Recommended writing:
> All VAAPI functions implemented in libva are thread-safe insofar as the driver and backend
> implementation are thread-safe. Note that libva cannot impose thread-safety guarantees
> on neither the driver nor the backend, so it is ultimately the responsability of the latter to
> ensure said thread-safety.

Please fix "`thread-saftey`" with "`thread-safety`" (ideally also introduce a ref-link
to e.g. https://en.wikipedia.org/wiki/Thread_safety).

va/va.h
```diff
@@ -129,6 +129,98 @@ extern "C" {
...
+ * It is assumed that none of the VAAPI functions will be called from signal 
+ * handlers.
+ *
+ * Thread-safety in this context means that when VAAPI is being called by multiple
+ * concurrent threads, it will not crash or hang the OS, and VAAPI internal
+ * states will not be corrupted. When multiple threads are operating on the same
+ * VAAPI objects, it is the application's responsibility to synchronize these
+ * operations in order to generate the expected results. This is especially true
+ * when a VAContext object is shared among different threads. Due to the fact that
+ * the internal states of the VAContext object can be modified through API calls
```
yellowdoge: I'm confused here. A couple of lines above you say:
> and VAAPI internal states will not be corrupted

but here I read
> the internal states of the VAContext object can be modified

and, presumably, corrupted.  Could you please elaborate on what do you
mean with "`internal states`" in l.145-146 and why VAContext is not one 
of those?

va/va.h
jonathanbian: Thanks for the suggested changes and it's been incorporated into the latest version

va/va.h
jonathanbian: Changed "states" to "data structures" which may be a better term. Also removed the block on VAContext as it can just be treated as other VAAPI objects under multi-threaded usage, i.e. app needs to handle the synchronization with operations using VAContext to ensure the correct results.

sreerenjb:
@jonathanbian should we make it as a mandatory requirement for the va-backend driver to keep thread safety ? Otherwise the applications still have to think whether a particular driver is providing thread-safety or not.

jonathanbian:
@sreerenjb my thinking is that as there is no formal certification/conformation process for the backend driver, making it a mandatory requirement won't mean much and the application still needs to verify whether the backend driver is thread-safe or not.

sreerenjb:
@jonathanbian  the point of making this mandatory is that the applications don't have to take care of locks themselves rather they can point out the place where driver is misbehaving. Once it is pointed out, it is up to the driver devs to fix the  issue. Otherwise each middleware implementation has to write their own locking which is going to bring another level of bugs. I rather prefer to have the proper locking in one place for all middleware implementations which is the backend driver.

dvrogozh:
Yes, without making it mandatory on VAAPI level as spec conformance requirement, this PR does not make sense.

jonathanbian:
@sreerenjb @dvrogozh Updated with more strict requirements on the backend

ceyusa:
> Otherwise, gstreamer-vaapi's implementation can be regarded as wrong because it doesn't do anything for VA-API call serialization.

must of the VA-API calls (at least in gstvaapidisplay and gstvaapisurface) are mutexed: https://gitlab.freedesktop.org/gstreamer/gstreamer-vaapi/blob/master/gst-libs/gst/vaapi/gstvaapidisplay.c#L923

va/va.h
```diff
@@ -129,6 +129,92 @@ extern "C" {
...
+ *
+ *   // Decode frames in the bitstream
+ *   for (;;) {
+ *     // Parse one frame and decode 
+ *     vaBeginPicture(dpy, decode_context, surfaces[surface_index]); 
+ *     vaRenderPicture(dpy, decode_context, buf, ...);
+ *     vaEndPicture(dpy, decode_context);
+ *     // Poll the decoding status and enqueue the surface in display order after 
+ *     // decoding is complete
+ *     vaQuerySurfaceStatus();
```
XinfengZhang, line 190: or vaSyncSurface?

jonathanbian: vaSyncSurface() will block and vaQuerySurfaceStatus() may provide better performance. Either is fine in this context which is to demonstrate multi-threading usage.

daijh:
Not sure if it is the right place to ask.

The calling sequence described is quit common usage on libva, that DECODE, ENCODE and VPP work on different threads.
Even Intel-Media-SDK uses libva in this way, without global lock.

Want to double confirm if Intel-Media-Driver va-backend is 
- thread-safety, OR
- context-safety (no global lock required on independent va-contexts, perhaps still need lock on multiple threads access the same va-context?) 

Is the thread-safety or context-safety guaranteed by Intel-Media-Driver's design?

XinfengZhang:
@daijh sorry for missing your comments, yes, context-safety guaranteed by  https://github.com/intel/media-driver design

XinfengZhang (approved):

XinfengZhang (approved):

dvrogozh (approved):

PR #350, opened 2019-12-09
va: Allow clean build on OpenBSD 6.6

iarebatman:
These allow a clean, functional build on OpenBSD 6.6.
OpenBSD doesn't provide the same mechanism to get the
current thread id. Move existing TID references into
a new function get_pid_t and implement specifics there.

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -398,10 +398,10 @@ static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
         }
 
         va_infoMessage(dpy, "Trying to open %s\n", driver_path);
-#ifndef ANDROID
-        handle = dlopen( driver_path, RTLD_NOW | RTLD_GLOBAL | RTLD_NODELETE );
-#else
+#if (!defined(RTLD_NODELETE) || defined(ANDROID))
         handle = dlopen( driver_path, RTLD_NOW| RTLD_GLOBAL);
+#else
+        handle = dlopen( driver_path, RTLD_NOW | RTLD_GLOBAL | RTLD_NODELETE );
 #endif
         if (!handle) {
             /* Don't give errors for non-existing files */
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -24,6 +24,11 @@
  */
 
 #define _GNU_SOURCE 1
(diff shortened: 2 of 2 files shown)


iarebatman:
> Was that previously done by some custom patch during BSD packaging? If yes, can you, please, provide a link to the packaging scripts?

It is not currently in the OpenBSD ports system.  I'm currently putting a new port together and this
is my own patch.

dvrogozh:
@iarebatman : thank you for clarification. I believe that while libva is not in OpenBSD, it presents in FreeBSD and they are dealing with basically the same issue. Here is their patch: https://svnweb.freebsd.org/ports/head/multimedia/libva/files/patch-va_va__trace.c?view=markup. Moreover, looking into it they are doing pretty the same as you do in this PR and it seems they thought about OpenBSD support:
https://svnweb.freebsd.org/ports/head/multimedia/libva/files/patch-va_va__trace.c?view=markup#l36. Can you, please, evaluate whether this change will work for you?

Anyhow, I believe we need to take FreeBSD effort into account and simplify everyone's live providing support for variety of operating systems. As of now FreeBSD patch seems to be more generic. I suggest to put it under review here. Would you mind to create corresponding PR or update this one?

iarebatman:
> @iarebatman : thank you for clarification. I believe that while libva is not in OpenBSD, it presents in FreeBSD and they are dealing with basically the same issue. Here is their patch: https://svnweb.freebsd.org/ports/head/multimedia/libva/files/patch-va_va__trace.c?view=markup. Moreover, looking into it they are doing pretty the same as you do in this PR and it seems they thought about OpenBSD support:
> https://svnweb.freebsd.org/ports/head/multimedia/libva/files/patch-va_va__trace.c?view=markup#l36. Can you, please, evaluate whether this change will work for you?
> 
> Anyhow, I believe we need to take FreeBSD effort into account and simplify everyone's live providing support for variety of operating systems. As of now FreeBSD patch seems to be more generic. I suggest to put it under review here. Would you mind to create corresponding PR or update this one?

It does look like their patches would work.  I will close this PR and create a new one.

PR #356, opened 2019-12-30
adjust call sequence to ensure authenticate operation is executed

### LINKED ISSUE #355: Cannot force i965 driver on X11
Regressed by 6650b390d1c21c. Works fine on KMS console and Wayland but fails on X11. Forcing `iHD` (via `LIBVA_DRIVER_NAME`) and `i965` auto-detection (via `iHD` absence) works fine.
```
$ LIBVA_DRIVER_NAME=i965 vainfo
libva info: VA-API version 1.6.0
libva info: User environment variable requested driver 'i965'
libva info: Trying to open /usr/local/lib/dri/i965_drv_video.so
libva info: Found init function __vaDriverInit_1_6
libva error: /usr/local/lib/dri/i965_drv_video.so init failed
libva info: va_openDriver() returns -1
vaInitialize failed with error code -1 (unknown libva error),exit
```
According to `VA_INTEL_DEBUG=1` init function fails due to [DRI2 check](https://github.com/intel/intel-vaapi-driver/blob/329975c63123610fc750241654a3bd18add75beb/src/intel_driver.c#L110-L113).

XinfengZhang:
fixes #355

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -367,11 +367,23 @@ static VAStatus va_getDriverNameByIndex(VADisplay dpy, char **driver_name, int c
     VADisplayContextP pDisplayContext = (VADisplayContextP)dpy;
     const char *driver_name_env = NULL;
     VADriverContextP ctx;
+    VAStatus status = VA_STATUS_SUCCESS;
 
     ctx = CTX(dpy);
+    if (pDisplayContext->vaGetDriverNameByIndex) {
+        /*if vaGetDriverNameByIndex is implemented*/
+        status = pDisplayContext->vaGetDriverNameByIndex(pDisplayContext, driver_name, candidate_index);
+    } else {
+        if (candidate_index == 0)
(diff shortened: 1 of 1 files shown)


va/va.c
```diff
@@ -380,19 +392,14 @@ static VAStatus va_getDriverNameByIndex(VADisplay dpy, char **driver_name, int c
         va_infoMessage(dpy, "User requested driver '%s'\n", *driver_name);
         return VA_STATUS_SUCCESS;
     } else if (driver_name_env && (geteuid() == getuid())) {
+        if(!(*driver_name))
```
pierre-labastie: I'm almost sure this is wrong, unless something very clever is involved here: you seem to free `*driver_name` when it is 0...

XinfengZhang: sorry about it, I made a mistake, thanks you

PR #364, opened 2020-01-20
add definition to enforce both reflist not empty

XinfengZhang:
it is a suppliment of https://github.com/intel/libva/pull/220
to distinguish P frame and low delay B frame

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1019,6 +1019,8 @@ typedef union _VAConfigAttribValEncJPEG {
 #define VA_PREDICTION_DIRECTION_PREVIOUS                0x00000001
 /** \brief Driver support backward prediction frame/slice */
 #define VA_PREDICTION_DIRECTION_FUTURE                  0x00000002
+/** \brief Dirver require both reference list must be not empty for inter frame */
+#define VA_PREDICTION_DIRECTION_BI_NOT_EMPTY            0x00000004
 /**@}*/
 
 /** @name Attribute values for VAConfigAttribEncIntraRefresh */


HeJunyan:
OK for me, and @xhaihao ?
please implement to return correct value for query.

HeJunyan:
According to new status, I think this info may not be enough.
On ICELake, for VMEPAK, both low-delay-b and non low-delay-b is supported, while only low-delay-b for VDENC.

But on TGL,  only low-delay-b for VMEPAK and  only low-delay-b for VDENC.

This flag means the non low-delay-b is necessary for some entrypoint, then with this flag set, we must set low-delay-b enabled. But for ICELake's VMEPAK case, if this flag is set,  we then can not choose non low-delay-b mode. And if not set, we do not know whether the low-delay-b mode is available.

XinfengZhang:
@HeJunyan , hi , I am not sure whether I understand your concern, it just used to distinguish whether driver support P frame, if P frame is not support , it should be 1. or it should be 0

HeJunyan:
So, if this flag set, means only non-P support, and if not set,  we can choose P and non-P, one of them?

HeJunyan:
For example, on ICELake,  VDENC does not support P, so this flags is set, and I willl always set low-delay-b enabled, just ignore the user  setting of low-delay-b.
And still on ICELake, VMEPAK supports both P and non-P, so this flag will not be set(because we can support P),  and I can choose to enabled/disabled  low-delay-b based on user setting.
But on SKL, for VMEPAK, only support P, not support non-P, so this flag also will not be set, and I still enabled/disabled  low-delay-b based on user setting. But if user set low-delay-b to true, we will fail.

I am worried about when we really do not support  low-delay-b, but user force to use it, just like the last case above, the GPU will hang, and really bad user experience.

XinfengZhang:
@HeJunyan 
For example, on ICL, VDENC does not support P, so this flags is set, you should use low-delay B to replace P frame.
And still on ICL, you can choose to enabled/disabled low-delay-b based on user setting.
But on SKL, for VMEPAK, only support P, not support non-P, so this flag also will not be set， but there are other values (VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD) and attributes-VAConfigAttribEncMaxRefFrames, so you always know which frame type is supported, if user set un-supported one, you could check it.

HeJunyan:
`
But on SKL, for VMEPAK, only support P, not support non-P, so this flag also will not be set， but there are other values (VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD) and attributes-VAConfigAttribEncMaxRefFrames, so you always know which frame type is supported, if user set un-supported one, you could check it.
`
No, because you have B frame, VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD are always be set in this case, unless you just use IP mode.

HeJunyan:
`
But on SKL, for VMEPAK, only support P, not support non-P, so this flag also will not be set， but there are other values (VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD) and attributes-VAConfigAttribEncMaxRefFrames, so you always know which frame type is supported, if user set un-supported one, you could check it.
`
No, because you have B frame, VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD are always be set in this case, unless you just use IP mode.

HeJunyan:
`
But on SKL, for VMEPAK, only support P, not support non-P, so this flag also will not be set， but there are other values (VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD) and attributes-VAConfigAttribEncMaxRefFrames, so you always know which frame type is supported, if user set un-supported one, you could check it.
`
No, because you have B frame, VA_PREDICTION_DIRECTION_PREVIOUS VA_PREDICTION_DIRECTION_BACKWARD are always be set in this case, unless you just use IP mode.

HeJunyan:
Synced with @xhaihao ,  OK now, no problems. Please help to merge it. thanks.

HeJunyan:
I note that on ICElake VDEnc, VAConfigAttribEncMaxRefFrames query return ref0 is 3 and ref1 is 0. So, B frame is not supported. But the low-delay-b just converts P to B. So according to my understand, we should support B and then can support low-delay-b, am I right? 
If I am wrong, please correct me. If not, this should be a bug?

HeJunyan:
Any update for this MR?

XinfengZhang:
@xhaihao ,could you help to approve it?

HeJunyan:
@XinfengZhang please help to merge it. And help to implement it in the media driver, especially for VME mode. So far, VAConfigAttribPredictionDirection query for VME mode just return error.

XinfengZhang:
@HeJunyan , there are already a PR with media driver https://github.com/intel/media-driver/pull/858

PR #367, opened 2020-02-06
add comments for VAConfigAttribQPBlockSize to clarify the value

XinfengZhang:


cwang105 (approved):
agree with the change

PR #369, opened 2020-02-10
Return error when no va devices available.

LeptonWu:
Signed-off-by: Lepton Wu <ytht.net@gmail.com>

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -54,7 +54,7 @@ VA_DRM_GetNumCandidates(VADriverContextP ctx, int * num_candidates)
 {
     struct drm_state * const drm_state = ctx->drm_state;
     drmVersionPtr drm_version;
-    int num_of_candidate = 0;
+    int count = 0;
     const struct driver_name_map *m = NULL;
     if (!drm_state || drm_state->fd < 0)
         return VA_STATUS_ERROR_INVALID_DISPLAY;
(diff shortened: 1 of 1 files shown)


LeptonWu:
FYI, this is for fixing a regression introduced by https://github.com/intel/libva/commit/6650b39,

The current code without this fix  crashes on system without supported hardware (for example, inside an vm).

va/drm/va_drm_utils.c
```diff
@@ -69,7 +69,7 @@ VA_DRM_GetNumCandidates(VADriverContextP ctx, int * num_candidates)
     }
     drmFreeVersion(drm_version);
     *num_candidates = num_of_candidate;
-    return VA_STATUS_SUCCESS;
+    return num_of_candidate ? VA_STATUS_SUCCESS : VA_STATUS_ERROR_UNKNOWN;
```
XinfengZhang: *num_of_candidate?

dvrogozh: no:). num_of_candidate is local variable declared int. num_candidates is function argument declared int*. I would suggest to use simpler and more different names to avoid misreading:).

XinfengZhang: sorry，a mistake

LeptonWu: So should I roll back to original version?

LeptonWu: OK, now I changed num_of_candidate to count in my cl.
Please take another look, thanks!

LeptonWu:
Ping, any comments? Thanks

XinfengZhang:
LGTM, will merge it

XinfengZhang:
@LeptonWu please help to re-base it with latest code

LeptonWu:
Thanks, can we get this merge now?

PR #372, opened 2020-02-21
Add field disable_rir_overlap to VAEncMiscParameterRIR

pgribov:


va/va.h
```diff
@@ -2121,6 +2121,8 @@ typedef struct _VAEncMiscParameterRIR
             uint32_t enable_rir_column : 1;
 	    /* \brief enable RIR in row */
             uint32_t enable_rir_row : 1;
+	    /* \brief disable RIR overlapping */
+            uint32_t disable_rir_overlap : 1;
 	    uint32_t reserved : 30;
```
DenWolf: Please, decrease the number of reserved bits (should be 29 for your case)

DenWolf (changes requested):

va/va.h
pgribov: fixed

PR #377, opened 2020-02-24
add new interface vaMapBuffer2 for map operation optimization

XinfengZhang:
add new vaMapBuffer2
backend driver need some usage hint to optimize the operation

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -1430,14 +1430,43 @@ VAStatus vaMapBuffer(
 )
 {
     VADriverContextP ctx;
-    VAStatus va_status;
+    VAStatus va_status = VA_STATUS_SUCCESS;
 
     CHECK_DISPLAY(dpy);
     ctx = CTX(dpy);
 
-    va_status = ctx->vtable->vaMapBuffer(ctx, buf_id, pbuf);
+    if (ctx->vtable->vaMapBuffer2) {
+        va_status = ctx->vtable->vaMapBuffer2(ctx, buf_id, pbuf, VA_MAPBUFFER_FLAG_DEFAULT);
+    } else if (ctx->vtable->vaMapBuffer) {
+        va_status = ctx->vtable->vaMapBuffer(ctx, buf_id, pbuf);
+    }
 
-    VA_TRACE_ALL(va_TraceMapBuffer, dpy, buf_id, pbuf);
+    VA_TRACE_ALL(va_TraceMapBuffer, dpy, buf_id, pbuf, VA_MAPBUFFER_FLAG_DEFAULT);
+    VA_TRACE_RET(dpy, va_status);
+
+    return va_status;
+}
+
+VAStatus vaMapBuffer2(
+    VADisplay dpy,
+    VABufferID buf_id,  /* in */
+    void **pbuf,    /* out */
+    uint32_t flags      /*in */
+)
+{
+    VADriverContextP ctx;
+    VAStatus va_status = VA_STATUS_SUCCESS;
+
+    CHECK_DISPLAY(dpy);
+    ctx = CTX(dpy);
+
+    if (ctx->vtable->vaMapBuffer2) {
+        va_status = ctx->vtable->vaMapBuffer2(ctx, buf_id, pbuf, flags);
+    } else if (ctx->vtable->vaMapBuffer) {
+        va_status = ctx->vtable->vaMapBuffer(ctx, buf_id, pbuf);
+    }
+
+    VA_TRACE_ALL(va_TraceMapBuffer, dpy, buf_id, pbuf, flags);
     VA_TRACE_RET(dpy, va_status);
 
     return va_status;
--- a/va/va.h
+++ b/va/va.h
@@ -3895,6 +3895,28 @@ VAStatus vaMapBuffer(
 );
 
 /**
+ * Map data store of the buffer into the client's address space
+ * this interface could be used to convey the operation hint
+ * backend driver could use these hint to optimize the implementations
+ */
+
+/** \brief VA_MAPBUFFER_FLAG_DEFAULT is used when there are no flag specified
+ * same as VA_MAPBUFFER_FLAG_READ | VA_MAPBUFFER_FLAG_WRITE.
+ */
+#define VA_MAPBUFFER_FLAG_DEFAULT 0
+/** \brief application will read the surface after map */
+#define VA_MAPBUFFER_FLAG_READ    1
+/** \brief application will write the surface after map */
+#define VA_MAPBUFFER_FLAG_WRITE   2
+
+VAStatus vaMapBuffer2(
+    VADisplay dpy,
+    VABufferID buf_id,  /* in */
+    void **pbuf,        /* out */
+    uint32_t flags      /* in */
+);
+
+/**
  * After client making changes to a mapped data store, it needs to
  * "Unmap" it to let the server know that the data is ready to be
  * consumed by the server
--- a/va/va_backend.h
+++ b/va/va_backend.h
@@ -504,8 +504,16 @@ struct VADriverVTable {
         VACopyObject        *src,           /* in */
         VACopyOption        option          /* in */
     );
+
+    VAStatus(*vaMapBuffer2)(
+        VADriverContextP ctx,
+        VABufferID buf_id,                  /* in */
+        void **pbuf,                        /* out */
+        uint32_t flags                      /* in */
+    );
     /** \brief Reserved bytes for future use, must be zero */
-    unsigned long reserved[54];
+    unsigned long reserved[53];
+
 };
 
 struct VADriverContext {
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -1740,7 +1740,8 @@ static void va_TraceCodedBufferIVFHeader(struct trace_context *trace_ctx, void *
 void va_TraceMapBuffer(
     VADisplay dpy,
     VABufferID buf_id,    /* in */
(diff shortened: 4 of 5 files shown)


va/va.c
```diff
@@ -1432,14 +1432,48 @@ VAStatus vaMapBuffer (
...
   ctx = CTX(dpy);
   
   VA_FOOL_FUNC(va_FoolMapBuffer, dpy, buf_id, pbuf);
+  if (ctx->vtable->vaMapBuffer)
+  { 
+      va_status = ctx->vtable->vaMapBuffer( ctx, buf_id, pbuf );
+  }
+  else if (ctx->vtable->vaMapBuffer2)
+  {
+      va_status = ctx->vtable->vaMapBuffer2( ctx, buf_id, pbuf,VAMapBufferFlagDefault );
```
dvrogozh: Maybe we need to start trying from vaMapBuffer2?

va/va.h
```diff
@@ -3437,6 +3437,25 @@ VAStatus vaMapBuffer (
...
 );
 
+/**
+ * Map data store of the buffer into the client's address space
+ * this interface could be used to convey the operation hint
+ * backend driver could use these hint to optimize the implementations
+ */
+typedef enum
+{
+    VAMapBufferFlagDefault = 0,
```
dvrogozh: I think we need to add a comment what default means (i.e. read|write). Also, I think we need to add a comment that these enum flags are being supported for bitwise operations.

va/va.c
```diff
@@ -1432,14 +1432,48 @@ VAStatus vaMapBuffer (
...
+  
+  VA_FOOL_FUNC(va_FoolMapBuffer, dpy, buf_id, pbuf);
+  if(ctx->vtable->vaMapBuffer2)
+  { 
+      va_status = ctx->vtable->vaMapBuffer2(ctx, buf_id, pbuf, flags);
+  }
+  else if(ctx->vtable->vaMapBuffer)
+  {
+      va_status = ctx->vtable->vaMapBuffer(ctx, buf_id, pbuf);
+  }
```
dvrogozh: We need va_TraceMapBuffer2 and support for new READ|WRITE flags...

va/va.c
```diff
@@ -1432,16 +1432,50 @@ VAStatus vaMapBuffer (
...
   VA_FOOL_FUNC(va_FoolMapBuffer, dpy, buf_id, pbuf);
+  if (ctx->vtable->vaMapBuffer2)
+  {
+      va_status = ctx->vtable->vaMapBuffer2( ctx, buf_id, pbuf,VAMapBufferFlagDefault );
+  }
+  else if (ctx->vtable->vaMapBuffer)
+  { 
+      va_status = ctx->vtable->vaMapBuffer( ctx, buf_id, pbuf );
+  }
+  VA_TRACE_ALL(va_TraceMapBuffer, dpy, buf_id, pbuf, 0);
```
dvrogozh: VAMapBufferFlagDefault

va/va.h
dvrogozh: please, resolve

va/va.h
```diff
@@ -3437,6 +3437,28 @@ VAStatus vaMapBuffer (
...
+/**
+ * Map data store of the buffer into the client's address space
+ * this interface could be used to convey the operation hint
+ * backend driver could use these hint to optimize the implementations
+ */
+typedef enum
+{
+    /** \brief VAMapBufferFlagDefault is used when there are no flag specified*/
+    VAMapBufferFlagDefault = 0,
+    /** \brief application will read the surface after map */
```
dmitryermilov: Can the flags be used on OR ? It's not clear from text description.

va/va.h
dvrogozh: I suggest this text:
```
/** \brief VAMapBufferFlagDefault is used when there are no flag specified
 * Same as VAMapBufferFlagRead | VAMapBufferFlagWrite.
 */

```

va/va.h
dvrogozh: +1. Not clear whether this `VAMapBufferFlagRead | VAMapBufferFlagWrite` can be done. For example:
```
/** \brief Bitwise flag: application will read the surface after map */
```

and same below for write:
```
/** \brief Bitwise flag: application will write the surface after map */
```

va/va.h
XinfengZhang: they are enum values , not macro, it is why I dont use  VAMapBufferFlagRead | VAMapBufferFlagWrite, or next extended flag must set value to 0x4?

dvrogozh: > next extended flag must set value to 0x4?

Maybe we will just use `#define VAMapBufferFlagDefault` 0 and so on then?

IlyaSergeyev:
@XinfengZhang What the current status? This change can significantly improve encoding performance.

va/va.h
```diff
@@ -3440,6 +3440,28 @@ VAStatus vaMapBuffer (
...
+/**
+ * Map data store of the buffer into the client's address space
+ * this interface could be used to convey the operation hint
+ * backend driver could use these hint to optimize the implementations
+ */
+
+/** \brief VA_MAPBUFFER_FLAG_DEFAULT is used when there are no flag specified
+ * same as VA_MAPBUFFER_FLAG_READ | VA_MAPBUFFER_FLAG_WRITE.
+ */
+#define VA_MAPBUFFER_FLAG_DEFAULT 0
```
dmitryermilov, line 3452: @dvrogozh , @XinfengZhang , can we do drop VA_MAPBUFFER_FLAG_DEFAULT because VA_MAPBUFFER_FLAG_DEFAULT in fact == VA_MAPBUFFER_FLAG_READ | VA_MAPBUFFER_FLAG_WRITE

va/va.c
```diff
@@ -1432,18 +1432,52 @@ VAStatus vaMapBuffer (
...
+  VAStatus va_status = VA_STATUS_SUCCESS;
   
   CHECK_DISPLAY(dpy);
   ctx = CTX(dpy);
-  
+
   VA_FOOL_FUNC(va_FoolMapBuffer, dpy, buf_id, pbuf);
-  
-  va_status = ctx->vtable->vaMapBuffer( ctx, buf_id, pbuf );
+  if (ctx->vtable->vaMapBuffer2)
```
HeJunyan, line 1441: It's really a good optimization because of the flags can save the un-needed copy.
The misc usage of vtable->vaMapBuffer and vtable->vaMapBuffer2 within vaMapBuffer vaMapBuffer2 APIs may be verbose. For vaMapBuffer2 API, if driver does not provide the virtual function point, we can return VA_STATUS_ERROR_UNIMPLEMENTED directly. The app user should consider the fallback cases.

XinfengZhang: in the future, I suppose, somedriver will just implement vaMapBuffer2, to be compatible with old app, need vaMapBuffer also be supported. this why I add similar fallback logic in vaMapBuffer and vaMapBuffer2

nowrep:
What's the status of this? It would be great to have it merged.
Currently the issue in Mesa is that mapping all buffers read+write leads to bad performance on dGPUs.

dvrogozh:
If I recall correctly, it was not merged because of lack of implementation on the (intel) driver side. However, considering that new driver appeared (vaon12) and obviously it has interest in this feature, I think we need to go forward with it. PR needs to be rebased and conflicts which appeared while it was hanging around resolved (@XinfengZhang : can you, please, do that? or I can pick it up). And PR needs re-review.

@nowrep : you are looking for this PR in regards of mesa's vaon12?

nowrep:
That sounds good, thanks!

My interest for this PR is radeonsi but it will benefit all Mesa drivers, so that also includes vaon12.

dvrogozh:
Is there any discussion on mesa side regarding this? Maybe some issue filed?

nowrep:
This PR came up in discussion of this MR: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25597

Context: Mesa always mapped the buffer as write-only. Up until recently vaDeriveImage was not implemented for multi-planar formats on radeonsi, so applications would fallback to vaPut/GetImage. That's not the case anymore and it is now an issue when applications try to read from the mapped memory. The simple fix was to map the buffer as read+write, but it regressed performance when used for writing only so the change was reverted.

va/va.h
dvrogozh: I think that having explicitly said what's the behavior in case of naive default value (=0) is good. I would leave DEFAULT declared.

dvrogozh (approved):
still looks good

dvrogozh:
@nowrep, @sivileri : we are ready to merge in this libva change as soon as some driver(s) will raise a PR on their side to use this new VAAPI feature. Can you, please, help to prepare a change on mesa side and get it tried out?

nowrep:
Mesa implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25620

dvrogozh:
Thank you @nowrep for quick follow up. Quoting perf example from https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25620#note_2119488:

"Example with patched ffmpeg to use vaMapBuffer2:
```
ffmpeg -vaapi_device /dev/dri/renderD128 -f lavfi -i testsrc=size=1280x720 -vf format=nv12,hwupload -c:v h264_vaapi -async_depth 1 -f null -

Before (read+write): 540fps
After  (write only): 620fps
```

It's around 10-15% faster depending on the resolution. At 2048x2048 it's 160fps before and 180fps after."

dvrogozh:
I don't see a reason to hold it on any further. Merging in.

PR #378, opened 2020-02-25
[va_enc_vp9] Updated API for VP9 encoding

eshiryae:
- Added uncompressed_header_byte_offset, seg_id_block_size and
segmentation_update_data fields to VAEncPictureParameterBufferVP9.
- Added VAConfigAttribNumScalablePipes and VAConfigAttribSegmentFeatureSupport

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -794,6 +794,19 @@ typedef enum
      *  attribute value is \c VAConfigAttribValContextPriority
      */
     VAConfigAttribContextPriority       = 41,
+
+    /** \brief indicates the maximum supported number of tile columns minus 1.
+     * When app sets the tile column number equaling to OptMaxTileColumnsMinus1 +1, the performance could be
+     * optimized due to parallel processing. But app may also choose other tile column numbers which are not
+     * subject to this restriction. There is no such restriction for tile rows.
+     */
+    VAConfigAttribOptMaxTileColumnsMinus1 = 42,
+    /** \brief VP9 encoding features attribute.  Read-only.
+     *
+     * This attribute describes the supported block sizes and features of
+     * an VP9 encoder config.
+     */
+    VAConfigAttribEncVP9Features        = 43,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
@@ -1178,6 +1191,38 @@ typedef union _VAConfigAttribValContextPriority{
  * profile/entrypoint pair, then set the value to the following 
  */
 #define VA_ATTRIB_NOT_SUPPORTED 0x80000000
+
+/** \brief Attribute value for VAConfigAttribEncVP9Features.*/
+typedef union _VAConfigAttribValEncVP9Features {
+    struct {
+        /** \brief indicates if encoder supports application forced segmentation.
+         * If so, application should provide segmentation map and segmentation parameters.
+         * Encoder will honor it without altering.max num of concurrent frames from different stream
+         */
+        uint32_t forced_segmentation_support    : 1;
+        /** \brief indicates if encoder will internally determine the segmentation map when
+         * segmentation_enabled in VAEncPictureParameterBufferVP9 is enabled and segment map surface
+         * is not provided by app.
+         */
+        uint32_t auto_segmentation_support      : 1;
+        /** \brief indicates supported segmentation features:
+         * 0b0001:  Seg QP index delta is supported;
+         * 0b0010:  Seg loop filter level delta is supported;
+         * 0b0100:  reference is supported;
+         * 0b1000:  skip is supported;
+         * Values can be combined.
+         */
+        uint32_t segment_feature_support        : 4;
+        /** \brief indicates supported encode acceleration functions:
+         * 0b0001:  Enc+Pak;
+         * 0b0010:  Hybrid Pak function;
+         * 0b0100:  Enc only function;
+        uint32_t coding_function_type           : 3;
+        /** \brief reserved bit for future, must be zero */
+        uint32_t reserved                       : 23;
+    } bits;
+    uint32_t value;
+} VAConfigAttribValEncVP9Features;
 
 /** Get maximum number of profiles supported by the implementation */
 int vaMaxNumProfiles (
--- a/va/va_enc_vp9.h
+++ b/va/va_enc_vp9.h
@@ -154,8 +154,11 @@ typedef struct  _VAEncPictureParameterBufferVP9
     /** VP9 encoder may support dynamic scaling function.
      *  If enabled (enable_dynamic_scaling is set), application may request
      *  GPU encodes picture with a different resolution from the raw source.
-     *  GPU should handle the scaling process of source and
-     *  all reference frames.
+     *  Applications are expected to handle the scaling explicitly:
+     *  application should scale the raw source and send to driver what
+     *  would be encoded via frame_width_src/frame_height_src. Backend driver
+     *  would handle the scaling process of reference frames. In this case
+     *  frame_width_dst/frame_height_dst should be set to VA_INVALID_SURFACE.
      */
     /* raw source frame width in pixels */
     uint32_t    frame_width_src;
@@ -183,7 +186,7 @@ typedef struct  _VAEncPictureParameterBufferVP9
      */
     VASurfaceID reference_frames[8];
 
-	  /* buffer to store coded data */
+    /* buffer to store coded data */
     VABufferID  coded_buf;
 
     union {
@@ -202,7 +205,7 @@ typedef struct  _VAEncPictureParameterBufferVP9
             uint32_t    ref_frame_ctrl_l1              : 3;
 
             /** \brief Last Reference Frame index
-             *  Specifies the index to RefFrameList[] which points to the LAST
+             *  Specifies the index to reference_frames[] which points to the LAST
              *  reference frame. It corresponds to active_ref_idx[0] in VP9 code.
              */
             uint32_t    ref_last_idx                   : 3;
@@ -213,7 +216,7 @@ typedef struct  _VAEncPictureParameterBufferVP9
             uint32_t    ref_last_sign_bias             : 1;
 
             /** \brief GOLDEN Reference Frame index
-             *  Specifies the index to RefFrameList[] which points to the Golden
+             *  Specifies the index to reference_frames[] which points to the Golden
              *  reference frame. It corresponds to active_ref_idx[1] in VP9 code.
              */
             uint32_t    ref_gf_idx                     : 3;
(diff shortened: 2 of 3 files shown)


va/va.h
```diff
@@ -777,6 +777,18 @@ typedef enum
      * implementation, multiple frames encode/decode can improve HW concurrency
      */
     VAConfigAttribMultipleFrame         = 40,
+    /** \brief indicates the number of scalable pipes minus 1 that the underline encoder supports.
+     * Application may use this number to decide the number of tile columns if scalable pipes are used.
+     */
+    VAConfigAttribNumScalablePipes      = 41,
```
cwang105: it does not need to put something like "Minus1" in the parameter name?

eshiryae: Done

eshiryae:
Hi @cwang105 @XinfengZhang @onabiull could you please do a code-review?

cwang105 (approved):

cwang105:
approved

onabiull (approved):
LGTM

dvrogozh (approved):

dvrogozh:
@eshiryae : can you, please, rebase pulling in changes from the latest master?

eshiryae:
Commit was rebased.

dvrogozh (approved):

va/va_enc_vp9.h
```diff
@@ -151,20 +151,20 @@ typedef struct  _VAEncSequenceParameterBufferVP9
...
 typedef struct  _VAEncPictureParameterBufferVP9
 {
-    /** VP9 encoder may support dynamic scaling function.
+    /** VP9 encoder supports dynamic scaling function.
      *  If enabled (enable_dynamic_scaling is set), application may request
      *  GPU encodes picture with a different resolution from the raw source.
-     *  GPU should handle the scaling process of source and
-     *  all reference frames.
+     *  However application should scale the raw source, and send to driver
+     *  what would be encoded. GPU would handle the scaling process of reference frames.
```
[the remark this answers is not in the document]
sreerenjb: Good point. You never know who is using VA-API. Remember, it is not just used by Intel drivers. We don't support a feature in the driver doesn't mean that we can remove the corresponding API from the spec. But the proposed patch is only trying to "deprecate" the field. Ideally, we are not supposed to even deprecate unless there is a genuine reason which should affect all vendors. How about putting a comment for recommended usage. Something like "applications are expected to handle the scaling explicitly"?

eshiryae:
Changed description of scaling process.
@XinfengZhang @cwang105 @sreerenjb could you please have a look?

va/va_enc_vp9.h
```diff
@@ -151,11 +151,14 @@ typedef struct  _VAEncSequenceParameterBufferVP9
  */
 typedef struct  _VAEncPictureParameterBufferVP9
 {
-    /** VP9 encoder may support dynamic scaling function.
+    /** VP9 encoder supports dynamic scaling function.
      *  If enabled (enable_dynamic_scaling is set), application may request
```
sreerenjb, line 155: I'm confused.
Are we supposed to set separate frame_width_src/dst & frame_height_src/dst only when enabling Bitrate control?
I don't think that assumption is true.
For eg: Spatial encode with CQP mode can just feed width_src/dst & height_src/dst where there is no VAEncMiscParameterRateControl. It used to work(partially) with i965 driver.

eshiryae: No, it's not depend on rate control mode. This only allows per frame dynamic resolution change and driver will scale reference frames on the fly if dynamic scaling is enabled. Otherwise resolution can be changed with reset and adding key frame.

cwang105:
Since source frame scaling is supposed to be handled by host encoder, we may consolidate width_src/dst & height_src/dst to be just width and height. The _src and _dst design was for driver to scale the source.

Thanks,
Ce

va/va.h
```diff
@@ -780,6 +780,18 @@ typedef enum
      * implementation, multiple frames encode/decode can improve HW concurrency
      */
     VAConfigAttribMultipleFrame         = 40,
+    /** \brief indicates the number of scalable pipes minus 1 that the underline encoder supports.
```
XinfengZhang, line 783: from view of application , I could not understand salable pipes, could you change to another word or explain what's it

eshiryae: @XinfengZhang could you please have a look at updated description?

eshiryae:
> Since source frame scaling is supposed to be handled by host encoder, we may consolidate width_src/dst & height_src/dst to be just width and height. The _src and _dst design was for driver to scale the source. Thanks, Ce 

@XinfengZhang @sreerenjb Do you agree with this comment from @cwang105 ?

va/va.h
```diff
@@ -1152,6 +1164,38 @@ typedef union _VAConfigAttribValMultipleFrame {
...
+        /** \brief indicates supported encode acceleration functions:
+         * 0b0001:  Enc+Pak;
+         * 0b0010:  Hybrid Pak function;
+         * 0b0100:  Enc only function;
+        uint32_t coding_function_type           : 3;
+        /** \brief reserved bit for future, must be zero */
+        uint32_t reserved                       : 23;
+    } bits;
+    uint32_t value;
+} VAConfigAttribEncVP9Features;
```
XinfengZhang, line 1197: could we add a value to indicate the 422 or 444 support? it is not enough to use VAVP9Profile1

va/va_enc_vp9.h
```diff
@@ -154,8 +154,11 @@ typedef struct  _VAEncPictureParameterBufferVP9
     /** VP9 encoder may support dynamic scaling function.
      *  If enabled (enable_dynamic_scaling is set), application may request
      *  GPU encodes picture with a different resolution from the raw source.
-     *  GPU should handle the scaling process of source and
-     *  all reference frames.
+     *  Applications are expected to handle the scaling explicitly:
+     *  application should scale the raw source and send to driver what
+     *  would be encoded via frame_width_src/frame_height_src. GPU would
```
XinfengZhang, line 159: please replace GPU to "HW" or "backend driver"

eshiryae: Should I replace GPU everywhere or just here?

va/va.h
eshiryae: Now 422/444 support is set with VAConfigAttribRTFormat value (if it set as VA_RT_FORMAT_YUV422/444).
Do we need additional field to specify this?

va/va.h
dmitryermilov: I agree with Carl that it's not clear what scalable pipes are. Better to explain it at first or use another term

va/va.h
```diff
@@ -781,6 +781,18 @@ typedef enum
      * implementation, multiple frames encode/decode can improve HW concurrency
      */
     VAConfigAttribMultipleFrame         = 40,
+    /** \brief indicates the number of scalable pipes minus 1 that the underline encoder supports.
+     * Application may use this number to decide the number of tile columns if scalable pipes are used:
+     * in case of scalable pipeline (number of scalable pipes > 1), number of tile columns should be 
+     * less or equal to (number of scalable pipes minus 1) + 1.
```
dmitryermilov, line 787: "number of tile columns should be ".. what about tile rows?

va/va_enc_vp9.h
dmitryermilov: @eshiryae , in current PR

va/va.h
cwang105: You may change the parameter name to something like VAConfigAttribOptMaxTileColumnsMinus1. 
And add some comment like:
When app sets the tile column number equaling to OptMaxTileColumnsMinus1 +1, the performance could be optimized due to parallel processing. But app may also choose other tile column numbers which are not subject to this restriction. There is no such restriction for tile rows.

va/va.h
```diff
@@ -794,6 +794,19 @@ typedef enum
      *  attribute value is \c VAConfigAttribValContextPriority
      */
     VAConfigAttribContextPriority       = 41,
+
+    /** \brief indicates the maximum supported number of tile columns minus 1.
+     * When app sets the tile column number equaling to OptMaxTileColumnsMinus1 +1, the performance could be
```
XinfengZhang, line 799: @eshiryae , if the tile columns number equaling to (OptMaxTileColumnsMinus1 + 1) . is there also performance benefit?

eshiryae: yes, in that case the performance could be optimized due to parallel processing

va/va_enc_vp9.h
```diff
@@ -353,7 +356,20 @@ typedef struct  _VAEncPictureParameterBufferVP9
...
+             * 0: 16x16 block size, default value;
+             * 1: 32x32 block size;
+             * 2: 64x64 block size;
+             * 3: 8x8 block size.
+             */
+            uint32_t    seg_id_block_size              : 2;
+
+            /** \brief Indicate that per segment parameters are changed.
+             */
+            uint32_t    segmentation_update_data       : 1;
```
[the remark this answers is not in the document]
eshiryae: needs to align with Windows. Also seg map can be updated but seg data not updated. and as for Linux segmentation_update_data is always set to 1 on driver side if segmentation enabled what is not correct, because segmentation_update_data should be set to 1 only if seg parameters changed. In that case there is a difference between header parameters in output streams encoded on Windows and Linux.

va/va_enc_vp9.h
```diff
@@ -510,8 +530,17 @@ typedef struct  _VAEncPictureParameterBufferVP9
...
     uint32_t    skip_frames_size;
 
+    /** \brief This parameter specifies the uncompressed header starting byte offset.
+     *  VP9 uncompressed header is generated by application. Application may also
+     *  inserts extra bytes in front of the uncompressed header for example IVF
+     *  container header. This parameter specifies the uncompressed header starting
+     *  byte offset in the compressed buffer that app prepares for driver to insert
+     *  into final bit stream.
+     */
+    uint32_t    uncompressed_header_byte_offset;
```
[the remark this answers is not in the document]
XinfengZhang: if application has no IVF frame header inserted into the packed header, it is zero, if the IVF header already been inserted into the packed header, the offset should be a constant 12. right? is it possible that uncompressed_header_byte_offset != 0 , != 12? another question is , which usage case set it to be non-zero value? GST & FFmpeg use sw-mux to add IVF into the bitstream.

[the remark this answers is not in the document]
XinfengZhang: of course, if there are IVF file header with first frame, the value maybe is 44?

va/va_enc_vp9.h
```diff
@@ -154,8 +154,11 @@ typedef struct  _VAEncPictureParameterBufferVP9
     /** VP9 encoder may support dynamic scaling function.
      *  If enabled (enable_dynamic_scaling is set), application may request
      *  GPU encodes picture with a different resolution from the raw source.
-     *  GPU should handle the scaling process of source and
-     *  all reference frames.
+     *  Applications are expected to handle the scaling explicitly:
+     *  application should scale the raw source and send to driver what
+     *  would be encoded via frame_width_src/frame_height_src. Backend driver
+     *  would handle the scaling process of reference frames. In this case
+     *  frame_width_dst/frame_height_dst should be set to VA_INVALID_SURFACE.
```
XinfengZhang, line 161: should be set to zero or equal frame_width_src/frame_width_dst , right?

PR #379, opened 2020-02-26
add new config attribute to report codec specific trivial features

XinfengZhang:
Signed-off-by: Carl Zhang <carl.zhang@intel.com>

PR #385, opened 2020-03-05
Add config attribute to advertise HEVC encoder features

fhvwy:
I see there is an incomplete version of pretty much the same thing already at https://github.com/intel/libva/pull/379 .  To compare with the most obvious change between them, I don't think there is any point in trying to make the config attribute look codec-generic because they have to be different for each codec anyway (like the existing JPEG one) and we aren't going to run out of index values any time soon.  There will likely need to be further attributes for HEVC RExt and SCC if they are added, since the additional flags there will not fit in the current attribute.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -371,6 +371,30 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_EXEC_MODE_POWER_SAVING 0x1
 #define VA_EXEC_MODE_PERFORMANCE  0x2
 
+/* Values used to describe device features. */
+/** The feature is not supported by the device.
+ *
+ * Any corresponding feature flag must not be set.
+ */
+#define VA_FEATURE_NOT_SUPPORTED  0
+/** The feature is supported by the device.
+ *
+ * The user may decide whether or not to use this feature.
+ *
+ * Note that support for a feature only indicates that the hardware
+ * is able to use it; whether it is actually a positive change to
+ * enable it in a given situation will depend on other factors
+ * including the input provided by the user.
+ */
+#define VA_FEATURE_SUPPORTED      1
+/** The feature is required by the device.
+ *
+ * The device does not support not enabling this feature, so any
+ * corresponding feature flag must be set and any additional
+ * configuration needed by the feature must be supplied.
+ */
+#define VA_FEATURE_REQUIRED       2
+
 /**
  * Returns a short english description of error_status
  */
@@ -956,6 +980,26 @@ typedef enum
      */
     VAConfigAttribProtectedContentUsage = 49,
 
+    /** \brief HEVC/H.265 encoding features.  Read-only.
+     *
+     * This attribute describes the supported features of an
+     * HEVC/H.265 encoder configuration.  The value returned uses the
+     * VAConfigAttribValEncHEVCFeatures type.
+     *
+     * If this attribute is supported by a driver then it must also
+     * support the VAConfigAttribEncHEVCBlockSizes attribute.
+     */
+    VAConfigAttribEncHEVCFeatures       = 50,
+    /** \brief HEVC/H.265 encoding block sizes.  Read-only.
+     *
+     * This attribute describes the supported coding tree and transform
+     * block sizes of an HEVC/H.265 encoder configuration.  The value
+     * returned uses the VAConfigAttribValEncHEVCBlockSizes type.
+     *
+     * If this attribute is supported by a driver then it must also
+     * support the VAConfigAttribEncHEVCFeatures attribute.
+     */
+    VAConfigAttribEncHEVCBlockSizes     = 51,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
--- a/va/va_enc_hevc.h
+++ b/va/va_enc_hevc.h
@@ -45,6 +45,193 @@ extern "C" {
  * @{
  */
 
+/** Attribute value for VAConfigAttribEncHEVCFeatures.
+ *
+ * This attribute decribes the supported features of an HEVC/H.265
+ * encoder configuration.
+ *
+ * All of the field values in this attribute are VA_FEATURE_* values,
+ * indicating support for the corresponding feature.
+ */
+typedef union VAConfigAttribValEncHEVCFeatures {
+    struct {
+        /** Separate colour planes.
+         *
+         * Allows setting separate_colour_plane_flag in the SPS.
+         */
+        uint32_t separate_colour_planes     : 2;
+        /** Scaling lists.
+         *
+         * Allows scaling_list() elements to be present in both the SPS
+         * and the PPS.  The decoded form of the scaling lists must also
+         * be supplied in a VAQMatrixBufferHEVC buffer when scaling lists
+         * are enabled.
+         */
+        uint32_t scaling_lists              : 2;
+        /** Asymmetric motion partitions.
+         *
+         * Allows setting amp_enabled_flag in the SPS.
+         */
+        uint32_t amp                        : 2;
+        /** Sample adaptive offset filter.
+         *
+         * Allows setting slice_sao_luma_flag and slice_sao_chroma_flag
+         * in slice headers.
+         */
+        uint32_t sao                        : 2;
+        /** PCM sample blocks.
+         *
+         * Allows setting pcm_enabled_flag in the SPS.  When enabled
+         * PCM parameters must be supplied with the sequence parameters,
+         * including block sizes which may be further constrained as
+         * noted in the VAConfigAttribEncHEVCBlockSizes attribute.
+         */
+        uint32_t pcm                        : 2;
+        /** Temporal motion vector Prediction.
+         *
+         * Allows setting slice_temporal_mvp_enabled_flag in slice
+         * headers.
+         */
+        uint32_t temporal_mvp               : 2;
+        /** Strong intra smoothing.
+         *
+         * Allows setting strong_intra_smoothing_enabled_flag in the SPS.
+         */
+        uint32_t strong_intra_smoothing     : 2;
+        /** Dependent slices.
+         *
+         * Allows setting dependent_slice_segment_flag in slice headers.
+         */
+        uint32_t dependent_slices           : 2;
+        /** Sign data hiding.
+         *
+         * Allows setting sign_data_hiding_enable_flag in the PPS.
+         */
+        uint32_t sign_data_hiding           : 2;
+        /** Constrained intra prediction.
+         *
+         * Allows setting constrained_intra_pred_flag in the PPS.
+         */
+        uint32_t constrained_intra_pred     : 2;
+        /** Transform skipping.
+         *
+         * Allows setting transform_skip_enabled_flag in the PPS.
+         */
+        uint32_t transform_skip             : 2;
+        /** QP delta within coding units.
+         *
+         * Allows setting cu_qp_delta_enabled_flag in the PPS.  When
+         * enabled, diff_cu_qp_delta_depth must be set to zero.
+         */
+        uint32_t cu_qp_delta                : 2;
+        /** Weighted prediction.
+         *
+         * Allows setting weighted_pred_flag and weighted_bipred_flag in
+         * the PPS.  The pred_weight_table() data must be supplied with
+         * every slice header when weighted prediction is enabled.
+         */
+        uint32_t weighted_prediction        : 2;
+        /** Transform and quantisation bypass.
+         *
+         * Allows setting transquant_bypass_enabled_flag in the PPS.
+         */
+        uint32_t transquant_bypass          : 2;
+        /** Deblocking filter disable.
+         *
+         * Allows setting slice_deblocking_filter_disabled_flag.
+         */
+        uint32_t deblocking_filter_disable  : 2;
+        /* Reserved for future use. */
+        uint32_t reserved                   : 2;
+    } bits;
+    uint32_t value;
+} VAConfigAttribValEncHEVCFeatures;
+
+/** Attribute value for VAConfigAttribEncHEVCBlockSizes.
+ *
+ * This attribute describes the supported coding tree and transform block
+ * sizes of an HEVC/H.265 encoder configuration
+ */
+typedef union VAConfigAttribValEncHEVCBlockSizes {
+    struct {
+        /** Largest supported size of coding tree blocks.
+         *
+         * CtbLog2SizeY must not be larger than this.
+         */
+        uint32_t log2_max_coding_tree_block_size_minus3    : 2;
+        /** Smallest supported size of coding tree blocks.
+         *
+         * CtbLog2SizeY must not be smaller than this.
+         *
+         * This may be the same as the maximum size, indicating that only
+         * one CTB size is supported.
+         */
+        uint32_t log2_min_coding_tree_block_size_minus3    : 2;
+
+        /** Smallest supported size of luma coding blocks.
+         *
+         * MinCbLog2SizeY must not be smaller than this.
+         */
+        uint32_t log2_min_luma_coding_block_size_minus3    : 2;
+
+        /** Largest supported size of luma transform blocks.
+         *
+         * MaxTbLog2SizeY must not be larger than this.
+         */
+        uint32_t log2_max_luma_transform_block_size_minus2 : 2;
+        /** Smallest supported size of luma transform blocks.
+         *
+         * MinTbLog2SizeY must not be smaller than this.
+         */
+        uint32_t log2_min_luma_transform_block_size_minus2 : 2;
+
+        /** Largest supported transform hierarchy depth in inter
+         *  coding units.
+         *
+         * max_transform_hierarchy_depth_inter must not be larger
+         * than this.
+         */
+        uint32_t max_max_transform_hierarchy_depth_inter   : 2;
+        /** Smallest supported transform hierarchy depth in inter
+         *  coding units.
+         *
+         * max_transform_hierarchy_depth_inter must not be smaller
+         * than this.
+         */
+        uint32_t min_max_transform_hierarchy_depth_inter   : 2;
+
+        /** Largest supported transform hierarchy depth in intra
+         *  coding units.
+         *
+         * max_transform_hierarchy_depth_intra must not be larger
+         * than this.
+         */
+        uint32_t max_max_transform_hierarchy_depth_intra   : 2;
+        /** Smallest supported transform hierarchy depth in intra
+         *  coding units.
+         *
+         * max_transform_hierarchy_depth_intra must not be smaller
+         * than this.
+         */
+        uint32_t min_max_transform_hierarchy_depth_intra   : 2;
+
+        /** Largest supported size of PCM coding blocks.
+         *
+         *  Log2MaxIpcmCbSizeY must not be larger than this.
+         */
+        uint32_t log2_max_pcm_coding_block_size_minus3     : 2;
+        /** Smallest supported size of PCM coding blocks.
+         *
+         *  Log2MinIpcmCbSizeY must not be smaller than this.
+         */
+        uint32_t log2_min_pcm_coding_block_size_minus3     : 2;
+
+        /** Reserved for future use. */
+        uint32_t reserved                                  : 10;
+    } bits;
+    uint32_t value;
+} VAConfigAttribValEncHEVCBlockSizes;
+
 /**
  * @name Picture flags
  *
--- a/va/va_str.c
+++ b/va/va_str.c
@@ -141,6 +141,8 @@ const char *vaConfigAttribTypeStr(VAConfigAttribType configAttribType)
     TOSTR(VAConfigAttribProtectedContentCipherMode);
     TOSTR(VAConfigAttribProtectedContentCipherSampleType);
     TOSTR(VAConfigAttribProtectedContentUsage);
+    TOSTR(VAConfigAttribEncHEVCFeatures);
+    TOSTR(VAConfigAttribEncHEVCBlockSizes);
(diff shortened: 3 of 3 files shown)


fhvwy:
See https://github.com/intel/media-driver/pull/866 and the set at https://lists.ffmpeg.org/pipermail/ffmpeg-devel/2020-March/257991.html for use of this.

XinfengZhang:
@xuguangxin @fulinjie please check interface whether it meet your requirement

fulinjie:
It makes sense, and got several questions raised in:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20200305002528.11418-3-sw@jkqxz.net/

va/va.h
```diff
@@ -1141,6 +1147,64 @@ typedef union _VAConfigAttribValMultipleFrame {
...
+        uint32_t log2_min_luma_transform_block_size_minus2 : 2;
+
+        /** Maximum supported transform hierarchy depth for inter blocks. */
+        uint32_t max_transform_hierarchy_depth_inter       : 2;
+        /** Maximum supported transform hierarchy depth for intra blocks. */
+        uint32_t max_transform_hierarchy_depth_intra       : 2;
+
+        /** Minimum supported block size containing a QP delta. */
+        uint32_t min_cu_qp_delta_block_size_minus3         : 2;
+    } bits;
```
XinfengZhang: 27 bit used, should we add "uint32_t reserved :5" ?

XinfengZhang: @fhvwy  hi Mark , could you help to add reserved bits and rebase the PR? thanks

fhvwy:
(Only back to this recently, apologies for the long delay.)
Here is a new version.
The main change is that feature support is now tristate - the driver can indicate that a feature is either not supported, optionally supported, or required.  This was prompted by some Intel hardware requiring `cu_qp_delta_enabled_flag` to be set, even in CQP mode when that isn't wanted, and I think it does make sense in general as other features may have similar constraints in other hardware.
Since all the feature fields are now two bits, that change necessitated splitting the attribute in two - the block sizes are now all in a separate attribute (though I've documented that a driver must support both if it supports either).
Documentation is more precise this time, too - it would be helpful if someone familiar with the block sizes could check that the bounds it offers are correctly defined and do the right thing, since that's the trickiest part of this.

fhvwy:
For FFmpeg implementing the user-side of this, see https://github.com/fhvwy/FFmpeg/tree/h265query

va/va_enc_hevc.h
```diff
@@ -45,6 +45,200 @@ extern "C" {
...
+ * This attribute describes the supported coding tree and transform block
+ * sizes of an HEVC/H.265 encoder configuration
+ */
+typedef union VAConfigAttribValEncHEVCBlockSizes {
+    struct {
+        /** Largest supported size of coding tree blocks.
+         *
+         * CtbLog2SizeY must not be larger than this.
+         */
+        uint32_t log2_max_coding_tree_block_size_minus3    : 2;
```
XinfengZhang, line 162: ask a question: if some device support LTU = 64, but not support LTU=32. how to report this attributes?
how to distinguish it from : support LTU=64 , also support LTU=32. both ok.

fhvwy: Only 64x64 luma CTBs supported:
`log2_max_coding_tree_block_size_minus3 = 3`
`log2_min_coding_tree_block_size_minus3 = 3`
Both 64x64 and 32x32 luma CTBs supported:
`log2_max_coding_tree_block_size_minus3 = 3`
`log2_min_coding_tree_block_size_minus3 = 2`

XinfengZhang:
LGTM, will prepare a correlated PR with https://github.com/intel/media-driver to use it, then will merge it.

XinfengZhang (approved):

va/va_enc_hevc.h
```diff
@@ -45,6 +45,200 @@ extern "C" {
...
+         * and the PPS.  The decoded form of the scaling lists must also
+         * be supplied in a VAQMatrixBufferHEVC buffer when scaling lists
+         * are enabled.
+         */
+        uint32_t scaling_lists              : 2;
+        /** Asymmetric motion partitions.
+         *
+         * Allows setting amp_enabled_flag in the SPS.
+         */
+        uint32_t amp                        : 2;
```
Tianhaol, line 75: If amp is always supported by driver, how to set this field?

va/va_enc_hevc.h
```diff
@@ -45,6 +45,200 @@ extern "C" {
...
+ * All of the field values in this attribute are VA_FEATURE_* values,
+ * indicating support for the corresponding feature.
+ */
+typedef union VAConfigAttribValEncHEVCFeatures {
+    struct {
+        /** Separate colour planes.
+         *
+         * Allows setting separate_colour_plane_flag in the SPS.
+         */
+        uint32_t separate_colour_planes     : 2;
```
Tianhaol, line 62: If from driver side, this value should be 0, cannot be set to 1,  how to set this field?  The meaning of 0 and 1 are identical to Recommendation ITU-T H.265

va/va_enc_hevc.h
```diff
@@ -45,6 +45,200 @@ extern "C" {
...
+         *
+         * Allows setting amp_enabled_flag in the SPS.
+         */
+        uint32_t amp                        : 2;
+        /** Sample adaptive offset filter.
+         *
+         * Allows setting slice_sao_luma_flag and slice_sao_chroma_flag
+         * in slice headers.
+         */
+        uint32_t sao                        : 2;
```
Tianhaol, line 81: Driver allows these two slice_sao_luma_flag and slice_sao_chroma_flag to be disabled/enabled respectively. How to set this field, which bit represent luma/chroma?

va/va_enc_hevc.h
```diff
@@ -45,6 +45,200 @@ extern "C" {
...
+         * allowed by the VAConfigAttribEncHEVCBlockSizes attribute.
+         */
+        uint32_t cu_qp_delta                : 2;
+        /** Weighted prediction.
+         *
+         * Allows setting weighted_pred_flag and weighted_bipred_flag in
+         * the PPS.  The pred_weight_table() data must be supplied with
+         * every slice header when weighted prediction is enabled.
+         */
+        uint32_t weighted_prediction        : 2;
```
Tianhaol, line 134: Same with SAO. Driver also allows weighted_pred_flag and weighted_bipred_flag to be disabled/enabled respectively. How to set this field?

va/va_enc_hevc.h
fhvwy: If the driver requires amp to always be enabled then it should return VA_FEATURE_REQUIRED in this field.

va/va_enc_hevc.h
fhvwy: Yes.  `VA_FEATURE_NOT_SUPPORTED` from the driver indicates that the separate colour planes are not usable, so `separate_colour_planes` must then always be set to zero by the user.

va/va_enc_hevc.h
fhvwy: Are you thinking of hardware which supports SAO only on one type of plane but not the other?  I was thinking that no such hardware would exist, but if you think it is possible then we could add separate `sao_luma` and `sao_chroma` fields to distinguish the cases.

va/va_enc_hevc.h
fhvwy: So the same case but for hardware which only has partial weighted prediction support?  Again, I was imagining that no such hardware would exist, but we could separate them if you do have such hardware.  (Hardware which lacks support for encoding P or B frames wouldn't care about the value of the other flag, because it would never be used.)

va/va_enc_hevc.h
fhvwy: Rereading that, I'm now not sure that that was your question.  To clarify my intended meaning:

If `VAConfigAttribValEncHEVCFeatures.sao == VA_FEATURE_NOT_SUPPORTED`, then the user must set both `slice_sao_luma_flag` and `slice_sao_chroma_flag` to zero.

If `VAConfigAttribValEncHEVCFeatures.sao == VA_FEATURE_SUPPORTED`, then the user can pick whatever values they want for `slice_sao_luma_flag` and `slice_sao_chroma_flag`, and they need not be the same.

If `VAConfigAttribValEncHEVCFeatures.sao == VA_FEATURE_REQUIRED`, then the user must set both `slice_sao_luma_flag` and `slice_sao_chroma_flag` to one.

Do you have/expect hardware for which that would be a problem?

va/va_enc_hevc.h
XinfengZhang: I have a question about weighted_prediction, media-driver support two kind of weighted-prediction: 1. explicit weighted prediction, application should provide pred_weight_table, and media-driver do weighted prediction basing on the pred_weight_table from applicaiton, so a pre-process needed (such as, fade in/out detection) before encoding  2. media driver based weighted prediction (or implicit weight prediction), media-driver will analyze the statistic datas and will provide a pred_weigt_table internally, to help the video quality. it happens inside media-driver during encoding process. 
my question is: how to report this kind of capability?

va/va_enc_hevc.h
```diff
@@ -45,6 +45,200 @@ extern "C" {
...
+         * than this.
+         */
+        uint32_t min_max_transform_hierarchy_depth_intra   : 2;
+
+        /** Size of the smallest supported coding unit which can
+         *  carry a QP delta.
+         *
+         * Log2MinCuQpDeltaSize must not be smaller than this.
+         */
+        uint32_t log2_min_cu_qp_delta_size_minus3          : 2;
```
Tianhaol, line 224: From HEVC SPEC,  Log2MinCuQpDeltaSize = CtbLog2SizeY − diff_cu_qp_delta_depth,  if diff_cu_qp_delta_depth can be assigned to 0 or 3 for different LCU,  how to set log2_min_cu_qp_delta_size_minus3?

va/va_enc_hevc.h
fhvwy: @XinfengZhang  I think the setting for driver-produced prediction weights should be signalled by its own attribute rather than embedded here, since it's not actually related to the codec implementation - it would exist in H.264/H.266 as well.

(There are other similar cases where either a driver-internal process or some kind of preprocessing step could be used to produce values for codec processes - e.g. the segmentation map in VP8/VP9/AV1, scaling lists in H.264/H.265/H.266, global motion in AV1.  It would be sensible for these to all be treated in a consistent way.)

va/va_enc_hevc.h
fhvwy: @Tianhaol  The intent here is that the limit of a single hardware implementation is going to be the level of the smallest coding unit for which it can usefully vary qp.  If that is, for example, at the 16x16 block size then we set `log2_min_cu_qp_delta_size_minus3 = 1`, and that then implies that the max value of `diff_cu_qp_delta_depth` would be 2 for 64x64 CTBs or 1 for 32x32 CTBs.

Does that fit with how you use it?  If you have hardware which has significantly different implementations at different CTU sizes and therefore doesn't match that setup then we may need to come up with something else.

XinfengZhang: I have no such concern with current HW. but , if some HW support both 64x64 CTBs and 32x32CTBs, and with different CTBs, there are different values , for 64x64CTBs, diff_cu_qp_delta_depth = 3, log2_min_cu_qp_delta_size_minus3 = 0,
for 32x32CTBs , log2_min_cu_qp_delta_size_minus3 = 2.  of course, we are discussing potential risk. is it extendable ?

XinfengZhang:
@fhvwy @Tianhaol  implement the interface with https://github.com/intel/media-driver/pull/1121, could you help to take a look and review it?

va/va_enc_hevc.h
Tianhaol: If diff_cu_qp_delta_depth only can be assigned to two values, 3 or 0, not in the range of 0 to 3, how to set this field log2_min_cu_qp_delta_size_minus3?

XinfengZhang: @fhvwy  hi Mark, another question: if log2_min_cu_qp_delta_size_minus3 = 0, it will mean diff_cu_qp_delta_depth  = 3, does application could set a value < 3? if HW just support one setting, such like diff_cu_qp_delta_depth  must equal to 3. how to report it?

fhvwy: @Tianhaol @XinfengZhang  I think you're both talking about the same case, which would be for hardware with 64x64 CTBs where the inclusion of `delta_qp()` is either once at the beginning of each CTU or it has to be included in every CU.

I was not expecting a case like that, and the range-based definition I've given does not support it.  Given that it's not going to work as a range then I think we need some sort of mask setup for whether the depth is allowed to be the given value for each CTB size.

So, perhaps a new config attribute `VAConfigAttribValEncHEVCQPDeltaDepthSizes` defining that mask?  If CTB size NxN can support a QP delta depth of D then we set bit `(log2(N) - 3) * 4 + D` to 1 (and those bits are zero otherwise).  In your example case we would set the value `0b1001_0000_0000_0000 = 0x9000` to indicate that two allowed possibilities are for 64x64 CTBs with a depth of 0 or 3.  Would that make sense?

XinfengZhang: @fhvwy how about to add a "log2_max_cu_qp_delta_size_minus3"  which could be used with "log2_min_cu_qp_delta_size_minus3" to describe specific range, such as 3-3, 2-3, 2-2, 0-0 etc.

fhvwy: @XinfengZhang  Unfortunately that doesn't support the case of 0 or 3 but not anything in between.

I did a bit more investigation into hardware support here:

Intel Gen9 (Skylake): https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-skl-vol10-hevc.pdf
Page 122 has the requirement that `diff_cu_qp_delta_depth = 0`.  Since this was the first VAAPI H.265 encode implementation, this value is what ended up hard-coded by users (and which we are now seeking to remove) - e.g. http://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavcodec/vaapi_encode_h265.c;h=3e0af17ea543e9d6f8095c1e28f4cbf47158be1f;hb=HEAD#l546 .

Intel Gen9.5 (Kaby Lake, Coffee Lake, Comet Lake): https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-kbl-vol10-hevc.pdf
Page 215 has the requirement that `diff_cu_qp_delta_depth = 0`, matching the Gen9 behaviour.

Intel Gen11 (Ice Lake): https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-icllp-vol11-mediaengines_0.pdf
Page 109 offers that `diff_cu_qp_delta_depth` must be equal to either `0` or `(lcu_size - min_cu_size)`.  `lcu_size` means `CtbSizeY` and `min_cu_size` means `MinCbSizeY`, so this is exactly the case we described above - it will either have a single `delta_qp()` at the start of the CTU (with `diff_cu_qp_delta_depth = 0`) or have `delta_qp()` in every CU (with `diff_cu_qp_delta_depth = CtbSizeY - MinCbSizeY`), but no case in between.  This is the problematic case which means a range won't work, leading to the bit mask suggestion in my previous comment.

I can't find any documentation for Intel Gen12 (Tiger Lake / others) or for any of the AMD encode blocks (VCE/VCN).

XinfengZhang: thanks @fhvwy , yes, Gen12 should be similiar with Gen11 about this setting. I suggest to remove this field from this structure in this PR, and discuss it another new PR. or. rename it to represent cu_qp_delta_enabled_flag
for this case, from media_driver side, we want to suggest diff_cu_qp_delta_depth as 3 for better quality (maybe it will waste some bits for some bitstream, because of additional bits for cu_qp_delta_abs , cu_qp_delta_sign_flag?) .

fhvwy: @XinfengZhang  Ok, that's fair.  I think removing the field is the cleanest option for now, so I've done that and stated against the `cu_qp_delta` feature that `diff_cu_qp_delta_depth` must be set to zero in this version (if it gets revisited later with a new attribute then that comment will need to be updated to point to the new attribute).

XinfengZhang: thanks @fhvwy , about this comment change of cu_qp_delta [https://github.com/intel/libva/pull/385/files#diff-6fddfe52dd51a4b8baecfd42b1895e34f8d41f4e7ba69c4298f3802490b81292R124 ] , it means that application must set  diff_cu_qp_delta_depth to be 0 when cu_qp_delta = VA_FEATURE_SUPPORTED or VA_FEATURE_REQUIRED, right?  TBH, it was not my intension, because it cause 2 potential problems: 
1. backward compatible issue in future, even we add one new attribute to represent diff_cu_qp_delta_depth in future, application should always set it to 0.    
2. it is workable for gen9 and gen95, but suppose not workable for VDEnc path of gen11 and gen12
from this point of view, keeping log2_min_cu_qp_delta_size_minus3   and adding log2_max_cu_qp_delta_size_minus3 is better. 
or, keeping log2_min_cu_qp_delta_size_minus3, and change the comments of cu_qp_delta:
/* Allows setting cu_qp_delta_enabled_flag in the PPS.  if it is VA_FEATURE_SUPPORTED diff_cu_qp_delta_depth  range is [ 0 ~ log2 CtbSizeY - 3 - log2_min_cu_qp_delta_size_minus3] , if it is VA_FEATURE_REQUIRED, diff_cu_qp_delta_depth should be the value of   (log2 CtbSizeY - 3 - log2_min_cu_qp_delta_size_minus3) , 

sorry about my previous comments, I did not have a deeper thinking about it.

fhvwy: @XinfengZhang  My intent is that we should have the other attribute, but I'm trying to avoid getting bogged down on exactly what form it needs to take here when I don't have all of the information about what it needs to look like.  If `diff_cu_qp_delta_depth = 0` works with all existing hardware then I think it's good enough for these attributes, since that is the current state anyway.

To answer 1, that comment to always use zero would be changed once the new attribute is introduced to say that you should look at the new attribute, or use zero if it is not present.  This is only a problem if some future hardware is introduced which doesn't support zero at all, but that is broken anyway with the current definition of the API (where people have generally hardcoded Gen9 values).

For 2, based on the documentation Gen11 still supports zero, and if Gen12 is similar then it does too?  (Please correct me if this is wrong.)  So that works now, and the future attribute would then be able to offer the other possibilities as well (for the Gen11 example, the choice between 0 and 3 when using 64x64 CTUs).  Putting an Intel-only special case into this attribute when we are going to replace it soon after anyway does not seem like a good step to me.

fhvwy: @XinfengZhang  No it doesn't.  If `cu_qp_delta_enabled_flag` is not set then `delta_qp()` never codes anything regardless of the level it happens at and the QP is fixed - see 7.3.8.14.  The inference you quote from 7.4.3.3.1 only exists because it is convenient in 8.6.1 to have the `Log2MinCuQpDeltaSize` for quantisation groups which isn't the whole slice.

The reason not to add partial fields now is that they would be deprecated when we add a proper attribute later, and I don't think it is a good idea to be maintaining an extra incomplete source of the same information.  If you want to include that information in this PR, what do you think of the suggestion above of a `VAConfigAttribValEncHEVCQPDeltaDepthSizes` attribute?

XinfengZhang: yes, thanks, if cu_qp_delta_enabled_flag  == 1 and diff_cu_qp_delta_depth  == 0, it means there are only one syntax unit  delta_qp() for maximum TU when maximum TU is selected.  if cu_qp_delta_enabled_flag = 0  , no delta_qp() present.
 
add new VAConfigAttribValEncHEVCQPDeltaDepthSizes , there will be enough bits to describe this capability. but  as I mentioned, I am not very sure whether diff_cu_qp_delta_depth = 0 works on Gen12 VDEnc, even there are similar description with G11. need some test to verify it. @Tianhaol , please help to check.

about "but that is broken anyway with the current definition of the API (where people have generally hardcoded Gen9 values" , it is true for GST & FFmpeg.  because we have no such capability report, GST & FFmpeg may set it to 0 by default , but I suppose MSDK  set it to 3 by default for G12 VDEnc, if we add new attributes, MSDK also need to query and handle it. so it will be set to 0 instead.   suppose PSNR /SSIM will be impact , and impact validation cases.

about "proper attribute" , you mean that "min/max" is not enough for some cases, so after adding these, still need "proper attribute" then deprecate "min/max". what's the case? want to support different setting for different CTB size?

fhvwy: If software written for one driver rather than against generic VAAPI is already using special knowledge about the behaviour of the driver it runs on then I don't see why it couldn't continue to do that.  (Correct me if I'm wrong, but I assume there is no intent that the Intel Media SDK would run with, say, the Mesa VAAPI driver in future.)

Min/max is not sufficient for the 0/3 case described by @Tianhaol and in the Ice Lake documentation, so if we want to support that we need an attribute with more fine-grained information.

XinfengZhang: hi @fhvwy , @Tianhaol  has do some test, we try to use 0, 1, 2, 3 . it works. so maybe we could still keep previous design?log2_min_cu_qp_delta_size_minus3

XinfengZhang:
@fhvwy, if no other concern, I will merge the PR firstly, then I will file another PR to change the comments from "
" * Allows setting cu_qp_delta_enabled_flag in the PPS.  When        
 * enabled, diff_cu_qp_delta_depth must be set to zero."
 to 
"* Allows setting cu_qp_delta_enabled_flag in the PPS"

if there are new attribute to describe the restriction of delta_depth. it could be a supplementary.

is it ok?

PR #386, opened 2020-03-05
change the format of va_dec_av1.h to unix style

XinfengZhang:
and delete trailing white space

### DIFF
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -174,14 +174,15 @@ typedef struct _VAFilmGrainStructAV1 {
 
 typedef enum {
     /** identity transformation, 0-parameter */
-    VAAV1Identity           = 0,
+    VAAV1TransformationIdentity           = 0,
     /** translational motion, 2-parameter */
-    VAAV1Translation        = 1,
+    VAAV1TransformationTranslation        = 1,
     /** simplified affine with rotation + zoom only, 4-parameter */
-    VAAV1Rotzoom            = 2,
+    VAAV1TransformationRotzoom            = 2,
     /** affine, 6-parameter */
-    VAAV1Affine             = 3,
-    VAAV1Trans_Types,
+    VAAV1TransformationAffine             = 3,
+    /** transformation count */
+    VAAV1TransformationCount
 } VAAV1TransformationType;
 
 typedef struct _VAWarpedMotionParamsAV1{
@@ -613,15 +614,15 @@ typedef struct _VASliceParameterBufferAV1
      *  It uses the name slice_data_size to be consistent with other codec,
      *  but actually means tile_data_size.
      */
-    uint32_t                Slice_data_size;
+    uint32_t                slice_data_size;
     /**
      * offset to the first byte of the data buffer.
      */
-    uint32_t                Slice_data_offset;
+    uint32_t                slice_data_offset;
     /**
      * see VA_SLICE_DATA_FLAG_XXX definitions
      */
-    uint32_t                Slice_data_flag;
+    uint32_t                slice_data_flag;
 
     uint16_t                tile_row;
     uint16_t                tile_column;
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -233,6 +233,9 @@ typedef struct  _VADecPictureParameterBufferAV1
      */
     uint8_t                 bit_depth_idx;
 
+    /** \brief corresponds to AV1 spec variable of the same name. */
+    uint8_t                 matrix_coefficients;
+
     union {
         struct {
             uint32_t        still_picture                               : 1;
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -52,33 +52,33 @@ extern "C" {
 typedef struct _VASegmentationStructAV1 {
     union {
         struct {
-             /** Indicates whether segmentation map related syntax elements 
-             *  are present or not for current frame. If equal to 0, 
-             *  the segmentation map related syntax elements are 
-             *  not present for the current frame and the control flags of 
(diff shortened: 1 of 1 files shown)


va/va_dec_av1.h
```diff
@@ -1,656 +1,656 @@
...
- */
-typedef struct _VASliceParameterBufferAV1
-{
-    /**@{*/
-    /** \brief The byte count of current tile in the bitstream buffer,
-     *  starting from first byte of the buffer.
-     *  It uses the name slice_data_size to be consistent with other codec,
-     *  but actually means tile_data_size.
-     */
-    uint32_t                Slice_data_size;
```
uartie: I realize this isn't part of your change, but using uppercase on this member/field name (and others below) don't follow the rest of the vaapi coding style, afaict.  Can you address this, too?  Sorry I did not get to review the original patchset.

uartie: If you think it's allowed, then I'm ok with it.

XinfengZhang: @uartie I changed it, please help to check it

XinfengZhang: is there other violation? the header file just be merged about 10 hour, so we still have chance to change the variable name, OR, we will have backward compatible issue.

va/va_dec_av1.h
```diff
@@ -1,656 +1,656 @@
...
+typedef enum {
+    /** identity transformation, 0-parameter */
+    VAAV1Identity           = 0,
+    /** translational motion, 2-parameter */
+    VAAV1Translation        = 1,
+    /** simplified affine with rotation + zoom only, 4-parameter */
+    VAAV1Rotzoom            = 2,
+    /** affine, 6-parameter */
+    VAAV1Affine             = 3,
+    VAAV1Trans_Types,
```
uartie: mixed style

uartie: How about using similar naming and style as, for example, VAProcColorBalanceType enum.
```C
typedef enum {
  VAAV1TransformIdentity = 0,
  VAAV1TransformTranslation,
  VAAV1TransformRotZoom,
  VAAV1TransformAffine,
  VAAV1TransformCount
}  VAAV1TransformTypes;
```

uartie: or maybe change all `VAAV1Transform` to `VATransformAV1`, would be a little easier to read?

uartie:
LGTM

IlyaSergeyev:
This change named "change the format of va_dec_av1.h to unix style", but contain names change and data structure change?

va/va_dec_av1.h
```diff
@@ -1,656 +1,660 @@
...
+    /** \brief bit depth index
+     *  value range [0..2]
+     *  0 - bit depth 8;
+     *  1 - bit depth 10;
+     *  2 - bit depth 12;
+     */
+    uint8_t                 bit_depth_idx;
+
+    /** \brief corresponds to AV1 spec variable of the same name. */
+    uint8_t                 matrix_coefficients;
```
dmitryermilov: Hi @XinfengZhang , @uartie , am I right that this is now a backward compatible change? I mean adding "uint8_t matrix_coefficients" to the middle of the structure? Will we have to rebuild all libva based applications(ffmpeg, msdk, etc)?

uartie: This header file was just merged yesterday.  So I doubt applications are using them already.  Either way, it hasn't been included in any release, as of yet.

XinfengZhang: it will not break compatible , but the variable name change will lead to compatible issue, but it just merged 15 hours ago, so we have chance to change it

mgonchar: @uartie @XinfengZhang this insertion will not break compatibility only if compiler applies reasonable padding policy. Much safer is to insert new fields to the end of structs or reserve space in the middle with "reserved" fields, which can be replaced by some meaningful value in future releases. But anyway you cannot guarantee binary compatibility of structures layout without fixing it with special compiler pragmas. Even update of compiler without code change may result in changed structures binary layout. You can [look at our MSDK approach](https://github.com/Intel-Media-SDK/MediaSDK/pull/1426) which guarantees desired binary layout of our API structs. The only way to check that binary layout stays the same is fixing offset of each struct's field to precise value in ```static_assert``` like we did [here](https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/shared/src/mfx_static_assert_structs.cpp)

dvrogozh: Folks, I understand that this was not released, but, please, let's try to avoid such things in a future. I think it's much better and developer friendly to have master branch always supported in a backward compatible style. Otherwise developers may start to see random failures just because they forgot to update libva on their system! I really would like to limit backward incompatible changes in master branch to the absolutely required show stopper things. Everything else means that we did not spend enough time doing code review.

uartie: Yes, it's unfortunate that the initial code review for this public API was not thorough enough before it got merged.  I don't think anyone is trying to intentionally make a habit out of doing postmortem changes like this when that happens.  And, considering if things left unsaid, I think everyone already understands the consequences of doing such things.

The [time] window between the initial API merge and these changes was very small, so the call was made to allow it based on that.  I doubt there was any harm done other than slightly inconveniencing or annoying a few people.

Either way, this proves that it is very important to thoroughly review new public API's before they are merged.  The more eyes, the better.  And **coding standards really should be put in place and enforced** to help with this matter.  Otherwise, we just end up with sloppy and inconsistent interfaces that only discourage people from using them.

If the initial commit did not break my compilation due to missing install header then, unfortunately, I probably wouldn't have noticed some of the other problems I reported (i.e. code style and DOS format).

PR #387, opened 2020-03-05
Add new interface vaGetDeviceInfo

AntonGrishin:
For get specific information about device from the backend driver

Signed-off-by: Anton Grishin <anton.grishin@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1810,6 +1810,45 @@ vaQueryProcessingRate(
     unsigned int       *processing_rate
 );
 
+/**
+ * \brief Device info structure
+ */
+typedef struct _VADeviceInfo {
+    /** \brief Vendor id */
+    int32_t vendor_id;
+    /** \brief Device id */
+    int32_t device_id;
+    union
+    {
+        struct
+        {
+            /** availability of the device local memory. */
+            uint32_t has_local_memory : 1;
+            /** Reserved for future use, must be zero. */
+            uint32_t reserved : 31;
+         } bits;
+         uint32_t value;
+     } deviceinfo_flags;
+    /** \brief Reserved bytes for future use, must be zero. */
+    uint32_t reserved[VA_PADDING_LOW];
+} VADeviceInfo;
+
+/**
+ * \brief Gets device info from backend driver.
+ *
+ * This function gets device information. Upon successful return, the device information
+ * will be stored in \c device_info.
+ *
+ * @param[in] dpy             the VA display
+ * @param[out] device_info    the device info structure
+ *
+ */
+VAStatus
+vaGetDeviceInfo(
+    VADisplay           dpy,
+    VADeviceInfo       *device_info
+);
+
 typedef enum
 {
     VAEncMiscParameterTypeFrameRate 	= 0,
--- a/va/va.c
+++ b/va/va.c
@@ -983,6 +983,23 @@ VAStatus vaQueryProcessingRate (
   return vaStatus;
 }
 
+VAStatus vaGetDeviceInfo(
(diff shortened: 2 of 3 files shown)


va/va.h
```diff
@@ -1806,6 +1806,42 @@ vaQueryProcessingRate(
...
+    VAAdapterDiscrete
+} VAAdapterType;
+
+/**
+ * \brief Device info structure
+ */
+typedef struct _VADeviceInfo {
+    /** \brief Device id */
+    int32_t device_id;
+    /** \brief Adapter type. 0 - integrated, 1 - discrete */
```
alariono: good catch!

XinfengZhang:
what's the purpose to get such information?

dvrogozh (changes requested):

dvrogozh:
@XinfengZhang 

> what's the purpose to get such information?

I think that such an API and nicely selected information to expose would be useful. I have few cases when this is needed in mind:
1. To answer classic questions like: "I have just installed driver, does it work?", "I have dual GPU system, which GPU works?" As of now we need to explain a pretty low level theory of operation - see below. But for the user who just installed something and typed 'vainfo' we can significantly simplify life by just printing out device ID!
2. In case of X11 or Wayland displays such info is more trickier to get - again printing this out in vainfo will help.
3. Sometimes higher level libraries differentiate features depending on device IDs and device caps - that's what mediasdk is doing and that's where this request comes from. As of now mediasdk is limited to i915 driver only and such API will permit to relax this restiction.

> You might have 2 cards available as:
>                 /dev/dri/card0
>                 /dev/dri/card1
>                 /dev/dri/renderD128
>                 /dev/dri/renderD129
> 
> D128 corresponds to card0. Make sure that you user login can access both devices – check that your account is a member of groups which own above devices. If not – add to both. Usually that’s render and video groups:
>                 sudo usermod -aG video <username>
>                 sudo usermod -aG render <username>
> Don’t forget to relogin to the system for above commands take effect.
> 
> To identify which card is which run this:
>                 # cat /sys/class/drm/card0/device/device
>                 xxxx
>                 # cat /sys/class/drm/card1/device/device
>                 yyyy
> The output is device ID. Check against device ID lists what is what.
>

va/va.h
```diff
@@ -1810,6 +1810,45 @@ vaQueryProcessingRate(
...
 );
 
+/**
+ * \brief Device info structure
+ */
+typedef struct _VADeviceInfo {
+    /** \brief Device id */
+    int32_t device_id;
+    /** \brief Vendor id */
+    int32_t vendor_id;
```
dvrogozh, line 1820: Per my taste `vendor_id` should go before `device_id`. Please, reorder.

dvrogozh (commented):
add minor change, please. Other than that I like it.

va/va.h
AntonGrishin: Done

XinfengZhang:
hi @dvrogozh 
1. all component has such question, not only media, so I think you may need another standalone method which is not related with media or 3d or ocl
2. application will convey the device fd into media , such as vainfo --device .
3. feature set should depend on the capability report , not device id

FurongZhang:
I reviewed this PR and did not have concern about the specific code changes. Just not very clear about the motivation of code changes although I saw simliar code changes in MSDK(the usage of MSDK is in sample app which demonstrates to the user how to use dual adapter).

If it is for workload submission to the different device, app can query caps and submit the corresponding workload. Currently, iHD does not expose the specific features which is related to local memory. 

Thanks,
Furong

PR #389, opened 2020-03-12
[trace] av1 decode buffers trace

IlyaSergeyev:
AV1 decoder buffers added to the UMD log


XinfengZhang:
@IlyaSergeyev could you help to change the commit message from UMD to libva

PR #390, opened 2020-03-13
Add VAProfileHEVCSccMain444_10 for HEVC SCC

Tianhaol:
Signed-off-by: Tianhao Liu <tianhao.liu@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -395,7 +395,8 @@ typedef enum
     VAProfileHEVCSccMain10              = 30,
     VAProfileHEVCSccMain444             = 31,
     VAProfileAV1Profile0                = 32,
-    VAProfileAV1Profile1                = 33
+    VAProfileAV1Profile1                = 33,
+    VAProfileHEVCSccMain444_10          = 34
 } VAProfile;
 
 /**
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -392,7 +392,8 @@ typedef enum
     VAProfileHEVCMain444_12             = 28,
     VAProfileHEVCSccMain                = 29,
     VAProfileHEVCSccMain10              = 30,
-    VAProfileHEVCSccMain444             = 31
+    VAProfileHEVCSccMain444             = 31,
```
xuguangxin: Do we support no scc(nature content) in scc profile?
If yes, since we have palette_mode_enabled_flag to indicate scc enabled or not, whey we need a new scc profile?
If no, how we support content switch, from nature content to a screen content.

cwang105:
SCC profiles encapsulates all main or main10, 444, etc. tools, correspondingly. So None screen contents are also supported in SCC profiles.

“palette_mode_enabled_flag” is available only under SCC profile. But under SCC profile, palette mode still can be disabled.

If Palette mode is enabled, whether or not it is actually applied is decided at block level. That is why both nature scene and screen content can both be supported.

va/va.h
Tianhaol: Thanks for clarification from cwang105.  SCC profiles is back compatible, can support the preceding profiles.

XinfengZhang:
LGTM

XinfengZhang:
@Tianhaol  please help add related code in va_str.c

PR #401, opened 2020-04-02
remove check of vaPutSurface implementation

### LINKED ISSUE #361: X11 PutSurface dependency issue
https://github.com/intel/libva/blob/7eec6f8ffb399ac33d545b0368079c29ebe6af24/va/va.c#L550

I believe this line should be dependent on whether X11 is available or enabled. I just went through the process of removing X11 dependencies in favor of DRM for MediaSDK and media-driver (iHD) and was getting failures because `vaPutSurface` did not exist in the vtable for iHD driver.

XinfengZhang:
Fixes #361
some backend driver does not implement vaPutSurface

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -548,7 +548,6 @@ static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
                     CHECK_VTABLE(vaStatus, ctx, EndPicture);
                     CHECK_VTABLE(vaStatus, ctx, SyncSurface);
                     CHECK_VTABLE(vaStatus, ctx, QuerySurfaceStatus);
-                    CHECK_VTABLE(vaStatus, ctx, PutSurface);
                     CHECK_VTABLE(vaStatus, ctx, QueryImageFormats);
                     CHECK_VTABLE(vaStatus, ctx, CreateImage);
                     CHECK_VTABLE(vaStatus, ctx, DeriveImage);


va/va.c
```diff
@@ -548,7 +548,6 @@ static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
                     CHECK_VTABLE(vaStatus, ctx, EndPicture);
                     CHECK_VTABLE(vaStatus, ctx, SyncSurface);
                     CHECK_VTABLE(vaStatus, ctx, QuerySurfaceStatus);
-                    CHECK_VTABLE(vaStatus, ctx, PutSurface);
```
ceyusa, line 551: Shouldn't they just return VA_STATUS_ERROR_UNIMPLEMENTED?

XinfengZhang: if there are no implmented, it will set vaStatus = VA_STATUS_ERROR_UNKNOWN and return at https://github.com/intel/libva/blob/master/va/va.c#L572

XinfengZhang: https://github.com/intel/libva/blob/master/va/va.c#L55

ceyusa: Right. What I meant is, imo,  that the driver implementators should add a function that returns VA_STATUS_ERROR_UNIMPLEMENTED.

Meanwhile, returning VA_STATUS_ERROR_UNKNOWN might be misleading to the api user.

But anyway, I'm not opposing to this change. Just stating a point of view.

XinfengZhang: @ceyusa  I filed another PR #404  to return  VA_STATUS_ERROR_UNIMPLEMENTED

PR #402, opened 2020-04-02
Fix HDR10 MaxCLL and MaxFALL documentation

uartie:
MaxCLL and MaxFALL should be in units of 1 nit as
specified in CEA-861.3, Appendix A.

A uint16_t is not large enough to specify the full
range of possible MaxCLL/MaxFALL values in units of
0.0001 cd/m^2.  So conclude that the documentation
was just a copy/paste error.

Signed-off-by: U. Artie Eoff <ullysses.a.eoff@intel.com>

### DIFF
--- a/va/va_vpp.h
+++ b/va/va_vpp.h
@@ -805,15 +805,15 @@ typedef struct _VAHdrMetaDataHDR10
      */
     uint32_t    min_display_mastering_luminance;
     /**
-     * \brief The maximum content light level.
+     * \brief The maximum content light level (MaxCLL).
      *
-     * The value is in units of 0.0001 candelas per square metre.
+     * The value is in units of 1 candelas per square metre.
      */
     uint16_t    max_content_light_level;
     /**
-     * \brief The maximum picture average light level.
+     * \brief The maximum picture average light level (MaxFALL).
      *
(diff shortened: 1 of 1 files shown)


FurongZhang:
@uartie I need some time to confirm that. But after a rough look, I prefer not to accept this patch.

uartie:
@FurongZhang if the fields are uint16_t and represent 0.0001 nit units, then we can only effectively set them to a maximum of 6 nits.  Certainly, MaxCLL and MaxFALL support much, much higher nits than that!  Also, based on https://github.com/intel/media-driver/issues/827, the current iHD driver interprets these fields as 1 nit unit anyway.

uartie:
https://www.mysterybox.us/blog/2016/10/19/hdr-video-part-3-hdr-video-terms-explained (see Metadata section)

FurongZhang:
I checked the specification, maxcll and maxfall are in units of candelas per square metre.

FurongZhang:
OK.

FurongZhang:
@uartie  I have double confirmed the specification, maxcll and maxfall are in units of candelas per square metre. I have approved this PR. Please rebase if needed.

Issue #396, opened 2020-03-26
libva-2.7.0 tries to load iris_drv_video.so


frantisekz:
Since upgrade from libva-2.6 to libva-2.7.0-0.1.pre1, libva tries to open /usr/lib64/dri/iris_drv_video.so by default.

I believe this is wrong. I have installed both media-driver and intel-vaapi-driver. They provide /usr/lib64/dri/iHD_drv_video.so and /usr/lib64/dri/i965_drv_video.so . 

I have Kaby Lake gen GPU, so I'd expect iHD_drv_video.so to be used by default when both are present. Currently, I need to specify LIBVA_DRIVER_NAME=iHD/i965 to get VAAPI working.

XinfengZhang:
the "iris" should come from DRI2 query function , suppose you are not using DRM, https://github.com/intel/libva/blob/master/va/drm/va_drm_utils.c#L39, it will just get iHD and i965

frantisekz:
Don't know what do you mean by DRM, I am on F32 system with GNOME Xorg session, if that does help.

frantisekz:
That error happenned when using vainfo. When i run vlc, it hits the error, but proceeds into using iHD driver later:

`$ vlc
VLC media player 3.0.9 Vetinari (revision 3.0.8-222-g602c5525e8)
[0000556bc61a0670] main libvlc: Running vlc with the default interface. Use 'cvlc' to use vlc without interface.
[0000556bc622f7a0] main playlist: playlist is empty
libva info: VA-API version 1.7.0
libva info: Trying to open /usr/lib64/dri/iris_drv_video.so
libva info: va_openDriver() returns -1
[00007f3694005aa0] glconv_vaapi_x11 gl error: vaInitialize: unknown libva error
libva info: VA-API version 1.7.0
libva info: Trying to open /usr/lib64/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_6
libva info: va_openDriver() returns 0
[00007f36bcc0e550] avcodec decoder: Using Intel iHD driver - 1.0.0 for hardware decoding
`

frantisekz:
Hmm, @XinfengZhang , it seems like the culprit is in Intel driver change in Mesa 20.0: i965 to Iris driver. I've updated mesa and libva at once.

I can reproduce this issue with libva-2.6.1 too. Looks like driver name mapping needs to be updated.

kwizart:
The driver mapping will have to be handled at the xorg-x11-server level.
[...]

PR #403, opened 2020-04-04
Update dri/vaapi driver mapping for iris

kwizart:
As reported https://github.com/intel/libva/issues/396

Update the driver mapping to take the new Iris intel driver into account

Signed-off-by: Nicolas Chauvet <kwizart@gmail.com>

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -37,6 +37,8 @@ struct driver_name_map {
 };
 
 static const struct driver_name_map g_driver_name_map[] = {
+    { "iris",       4, "iHD"    }, // Intel Media driver with Iris DRI driver
+    { "iris",       4, "i965"   }, // Intel OTC GenX driver with Iris DRI driver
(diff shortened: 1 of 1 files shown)


uartie:
Hmm... Based on the execution trace that I did, I was under the impression that the problem reported in #396 was caused by the DRI2 query called in the X11 backend here: https://github.com/intel/libva/blob/42ce2a4af77456a5bf1cc42ed4a6fc0f1168b328/va/x11/va_dri2.c#L149

So I'm confused how updating the DRM backend driver map will fix that.  AFAICT, the X11 backend needs to add support for dri driver -> vaapi driver mapping to fix the reported issue, right?

uartie:
Is there a case where "iris" is reported via the DRM backend execution path?  Please enlighten me.

kwizart:
Sorry, seems like I've miss-understood the needed fix.
I should also have stated that I don't have the needed hw to test this (I don't have any iris capable intel gpu).

Closing.

PR #407, opened 2020-04-22
Add vaSyncSurface2 with timeout

### LINKED ISSUE #397: [enhance] Adding timeout for vaSyncSurface
Current ```vaSyncSurface``` interface pushes application to use a separate thread for synchronization, since app. can't afford to block main thread for unknown time.

My suggestion is to a new function add a ```vaSyncSurface2``` w/ ```timeout``` parameter -
```C
VAStatus vaSyncSurface2(
    VADisplay dpy,
    VASurfaceID render_target
    unsigned int timeout
);
```
Either ```VA_STATUS_ERROR_SURFACE_BUSY```\ ```VA_STATUS_ERROR_HW_BUSY``` statuses could be re-used to indicate the timeout is expired or a new one status ```VA_STATUS_ERROR_TIMEOUT``` could be introduced.

dmitryermilov:
Fixes #397

Signed-off-by: Dmitry Ermilov <dmitry.ermilov@intel.com>

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -1639,6 +1639,28 @@ VAStatus vaSyncSurface (
   return va_status;
 }
 
+VAStatus vaSyncSurface2 (
+    VADisplay dpy,
+    VASurfaceID surface,
+    uint64_t timeout_ns
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  if (ctx->vtable->vaSyncSurface2)
+      va_status = ctx->vtable->vaSyncSurface2( ctx, surface, timeout_ns );
+  else
+      va_status = VA_STATUS_ERROR_UNIMPLEMENTED;
+  VA_TRACE_LOG(va_TraceSyncSurface2, dpy, surface, timeout_ns);
+  VA_TRACE_RET(dpy, va_status);
+
+  return va_status;
+}
+
 VAStatus vaQuerySurfaceStatus (
     VADisplay dpy,
     VASurfaceID render_target,
--- a/va/va.h
+++ b/va/va.h
@@ -213,6 +213,8 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_STATUS_ERROR_UNSUPPORTED_MEMORY_TYPE 0x00000024
 /** \brief Indicate allocated buffer size is not enough for input or output. */
 #define VA_STATUS_ERROR_NOT_ENOUGH_BUFFER       0x00000025
+/** \brief Indicate an operation isn't completed because time-out interval elapsed. */
+#define VA_STATUS_ERROR_TIMEDOUT                0x00000026
 #define VA_STATUS_ERROR_UNKNOWN			0xFFFFFFFF
 
 /** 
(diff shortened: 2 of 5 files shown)


va/va.h
```diff
@@ -3722,6 +3722,21 @@ VAStatus vaSyncSurface (
     VASurfaceID render_target
 );
 
+/**
+ * This function blocks during specified timeout (in milliseconds) until
```
dvrogozh, line 3726: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/drm/i915_drm.h?h=v5.7-rc2#n1516

It accepts timeout in nanoseconds. I think we should not lower timeout precision. I would expose timeout in nanoseconds and make argument uint64_t or int64_t. What will be infinite timeout? should we have -1 and int64_t?

va/va.h
```diff
@@ -3722,6 +3722,21 @@ VAStatus vaSyncSurface (
...
+ * If timeout is zero, the function returns immediately.
+ * Return values:
+ * VA_STATUS_SUCCESS - all pending operations completed successfully;
+ * VA_STATUS_ERROR_HW_BUSY - operations are still in progress;
+ * the other error statuses indicating that operations failed;
+ */
+VAStatus vaSyncSurface2 (
+    VADisplay dpy,
+    VASurfaceID render_target,
+    unsigned int timeout
```
dvrogozh, line 3737: better to use timeout_ms or timeout_ns to emphasize values...

va/va.h
```diff
@@ -3722,6 +3722,21 @@ VAStatus vaSyncSurface (
...
+);
+
```
dvrogozh, line 3739: I believe we also need vaSyncBuffer... (separate commit or PR).

dmitryermilov: yes, this is my major task for tomorrow.

va/va.h
dmitryermilov: Okay.. Thanks!

va/va.h
dmitryermilov: Thanks! Good point. I'll address.
What do you think about the following aspect: should we highlight that if underlying kmd/hw tasks are still in progress, vaSyncSurface2 can wait more that timeout specified (similar to sleep/usleep)?

dmitryermilov:
@XinfengZhang , @dmitryermilov , do you have any other suggestions ?

va/va.h
```diff
@@ -213,6 +213,8 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_STATUS_ERROR_UNSUPPORTED_MEMORY_TYPE 0x00000024
 /** \brief Indicate allocated buffer size is not enough for input or output. */
 #define VA_STATUS_ERROR_NOT_ENOUGH_BUFFER       0x00000025
+/** \brief Indicate a synchronous operation is still running */
+#define VA_STATUS_ERROR_SYNCHRONIZATION_TIMED_OUT 0x00000026
```
dvrogozh, line 217: `VA_STATUS_ERROR_SYNCHRONIZATION_TIMED_OUT` I think that's too big name and SYNCHRONIZATION would be granted from function name. I would name new status just in generic manner like `VA_STATUS_ERROR_TIMEDOUT`.

dmitryermilov: Thanks. Addressed.

va/va.c
```diff
@@ -1640,6 +1640,25 @@ VAStatus vaSyncSurface (
...
+    uint64_t timeout_ns
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  va_status = ctx->vtable->vaSyncSurface2( ctx, surface, timeout_ns );
```
xhaihao, line 1655: Need check ctx->vtable->vaSyncSurface2 against NULL firstly. BTW could you delete the whitespace between brace and parameters to keep the same code style in this file?

dmitryermilov: Thanks! Done.

As for the whitespaces, they're present in all other functions, e.g.:
`  va_status = ctx->vtable->vaSyncSurface( ctx, render_target );`
`  va_status = ctx->vtable->vaQuerySurfaceStatus( ctx, render_target, status );`
`  va_status = ctx->vtable->vaQuerySurfaceError( ctx, surface, error_status, error_info );`

dmitryermilov:
@dvrogozh , @xhaihao , @XinfengZhang , @dmitry-gurulev , @wyc55069407,

Please give your input: do we need VA_TIMEOUT_INFINITE here?

wyc55069407:
> @dvrogozh , @xhaihao , @XinfengZhang , @dmitry-gurulev , @wyc55069407,
> 
> Please give your input: do we need VA_TIMEOUT_INFINITE here?

I think VA_TIMEOUT_INFINITE can be defined as max(uint64_t) or uint64_t(-1) at va.h.  And add comments in the new libva interface to show this value is specific to infinite.
(Like #define VA_TIMEOUT_INFINITE (0xffffffffffffffff))

When you updated va.h. i'll update https://github.com/intel/media-driver/pull/943 & https://github.com/intel/media-driver/pull/945

PR #408, opened 2020-04-23
Add vaSyncBuffer

dmitryermilov:
Signed-off-by: Dmitry Ermilov <dmitry.ermilov@intel.com>

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -1700,6 +1700,29 @@ VAStatus vaQuerySurfaceError (
   return va_status;
 }
 
+VAStatus vaSyncBuffer (
+    VADisplay dpy,
+    VABufferID buf_id,
+    uint64_t timeout_ns
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  VA_TRACE_LOG(va_TraceSyncBuffer, dpy, buf_id, timeout_ns);
+
+  if (ctx->vtable->vaSyncBuffer)
+      va_status = ctx->vtable->vaSyncBuffer( ctx, buf_id, timeout_ns );
+  else
+      va_status = VA_STATUS_ERROR_UNIMPLEMENTED;
+  VA_TRACE_RET(dpy, va_status);
+
+  return va_status;
+}
+
 /* Get maximum number of image formats supported by the implementation */
 int vaMaxNumImageFormats (
     VADisplay dpy
--- a/va/va.h
+++ b/va/va.h
@@ -3811,6 +3811,46 @@ VAStatus vaQuerySurfaceError(
 );
 
 /**
+ * \brief Synchronizes pending operations associated with the supplied buffer.
+ *
+ * This function blocks during specified timeout (in nanoseconds) until
+ * all pending operations on the supplied buffer have been completed.
+ * If timeout is zero, the function returns immediately.
+ *
+ * Possible errors:
+ * - \ref VA_STATUS_ERROR_UNIMPLEMENTED: the VA driver implementation
+ *   does not support this interface
+ * - \ref VA_STATUS_ERROR_INVALID_DISPLAY: an invalid display was supplied
+ * - \ref VA_STATUS_ERROR_INVALID_BUFFER: an invalid buffer was supplied
+ * - \ref VA_STATUS_ERROR_TIMEDOUT: synchronization is still in progress,
+ *   client should call the function again to complete synchronization
+ *
+ * @param[in] dpy         the VA display
+ * @param[in] buf_id      the buffer for which synchronization is performed
+ * @param[in] timeout_ns  the timeout in nanoseconds
+ *
+ */
+VAStatus vaSyncBuffer(
+    VADisplay dpy,
+    VABufferID buf_id,
+    uint64_t timeout_ns
+);
+
+/**
+ * Notes about synchronization interfaces:
+ * vaSyncSurface:
+ * 1. Allows to synchronize output surface (i.e. from decoding or VP)
+ * 2. Allows to synchronize all bitstreams being encoded from the given input surface (1->N pipelines).
+ *
+ * vaSyncSurface2:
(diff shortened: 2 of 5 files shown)


va/va.h
```diff
@@ -3777,6 +3777,22 @@ VAStatus vaQuerySurfaceError(
...
+ * Return values:
+ * VA_STATUS_SUCCESS - all pending operations completed successfully;
+ * VA_STATUS_ERROR_HW_BUSY - operations are still in progress;
+ * the other error statuses indicating that operations failed;
+ */
+VAStatus vaSyncBuffer(
+    VADisplay dpy,
+    VABufferID buf_id,
+    uint64_t timeout_ns
+);
```
dmitryermilov, line 3794: Need to think on how to get status reports (like encoded buffer size, QP, num passes, etc). Generally we can do something similar to https://github.com/intel/libva/blob/82122961145d4be616400cc99485867adfddd971/va/va.h#L3419 but there are just 4 bits left. We need something extensible.

va/va_backend.h
```diff
@@ -485,8 +485,14 @@ struct VADriverVTable
...
 
+        VAStatus (*vaSyncBuffer) (
+            VADriverContextP ctx,
+            VABufferID buf_id,
+            uint64_t timeout_ns
+        );
+
         /** \brief Reserved bytes for future use, must be zero */
-        unsigned long reserved[57];
+        unsigned long reserved[55];
```
xhaihao, line 495: The type is unsigned long, so the size of elements in this array is 56 after adding a function callback.

va/va.h
```diff
@@ -3777,6 +3777,22 @@ VAStatus vaQuerySurfaceError(
...
 );
 
+/**
+ * This function blocks during specified timeout (in nanoseconds) until
+ * all pending operations associated with given buf_id have been completed.
+ * If timeout is zero, the function returns immediately.
+ * A common example of usage is synchronization of output encoded buffers, output statistic buffers.
+ * Return values:
+ * VA_STATUS_SUCCESS - all pending operations completed successfully;
+ * VA_STATUS_ERROR_HW_BUSY - operations are still in progress;
```
xhaihao, line 3787: Does `VA_STATUS_ERROR_HW_BUSY` mean the underlying HW hang? How does user know that `timeout_ns` has passed but the underlying HW is still handling the operations?

dmitryermilov: >Does VA_STATUS_ERROR_HW_BUSY mean the underlying HW hang?

I can't say exactly. va.h doesn't describe it in details. If VA_STATUS_ERROR_HW_BUSY really means GPU hang then yes, it's incorrect to return this status here. In this case we would need something like VA_STATUS_ERROR_OPERATION_TIMED_OUT which in fact not an error but rather a warning indicating that not all operations associated with given VABufferID have been completed within timeout. 
What do you think?

va/va_backend.h
dmitryermilov: Thanks. Fixed.

va/va.h
dmitryermilov: Addressed by VA_STATUS_ERROR_TIMEDOUT

va/va.c
```diff
@@ -1679,6 +1679,25 @@ VAStatus vaQuerySurfaceError (
...
+    uint64_t timeout_ns
+)
+{
+  VAStatus va_status;
+  VADriverContextP ctx;
+
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  va_status = ctx->vtable->vaSyncBuffer( ctx, buf_id, timeout_ns );
```
xhaihao, line 1694: It is possible the underlying driver doesn't implement this callback function, so we should check ctx->vtabl->vaSyncBuffer against NULL before calling it. 
```suggestion
  if (ctx->vtable->vaSyncBuffer)
       va_status = ctx->vtable->vaSyncBuffer(ctx, buf_id, timeout_ns);
  else
       va_status = VA_STATUS_ERROR_UNIMPLEMENTED;
```

dmitryermilov: Thanks! Done.

dmitryermilov:
@dvrogozh , @xhaihao , @XinfengZhang , @dmitry-gurulev , @wyc55069407,

Please give your input: do we need VA_TIMEOUT_INFINITE here?

wyc55069407:
> @dvrogozh , @xhaihao , @XinfengZhang , @dmitry-gurulev , @wyc55069407,
> 
> Please give your input: do we need VA_TIMEOUT_INFINITE here?

same comments as in #407

XinfengZhang:
hi @dmitryermilov , LGTM, but it is new , which maybe will lead to different call sequence with before, could you help to add sample into libva-utils?

dmitryermilov:
> hi @dmitryermilov , LGTM, but it is new , which maybe will lead to different call sequence with before, could you help to add sample into libva-utils?

Hi @XinfengZhang , yes, I can do it. Unfortunately libva-utils has only simple encode tools. I can replace vaSyncSurface with vaSyncBuffer e.g. in avcenc. But I have to say that, IMO, it won't be very descriptive for readers since the most benefits will be in 1->N cases. avcenc doesn't support it. I can do it after 1-2 weeks due to other commitments. libva and libva-utils are different repo's, I'd propose to merge current PR now (it's blocking other tasks). Is it okay for you?

va/va.c
```diff
@@ -1700,6 +1700,28 @@ VAStatus vaQuerySurfaceError (
...
+  VADriverContextP ctx;
+
+  CHECK_DISPLAY(dpy);
+  ctx = CTX(dpy);
+
+  if (ctx->vtable->vaSyncBuffer)
+      va_status = ctx->vtable->vaSyncBuffer( ctx, buf_id, timeout_ns );
+  else
+      va_status = VA_STATUS_ERROR_UNIMPLEMENTED;
+  VA_TRACE_LOG(va_TraceSyncBuffer, dpy, buf_id, timeout_ns);
```
XinfengZhang, line 1719: please move it to  line 1714

dmitryermilov: done

PR #412, opened 2020-05-04
API was extended for look ahead support

splotnikv:
including LA depth, capability and hints reporting.

Signed-off-by: Sergey Plotnikov <sergey.plotnikov@intel.com>

### DIFF
--- a/va/va_enc_hevc.h
+++ b/va/va_enc_hevc.h
@@ -338,8 +338,14 @@ typedef struct _VAEncSequenceParameterBufferHEVC {
         } bits;
         uint32_t value;
     } scc_fields;
+	
+	/** \brief Lookahead depth in frames. */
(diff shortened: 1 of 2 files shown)


va/va_enc_hevc.h
```diff
@@ -338,8 +338,14 @@ typedef struct _VAEncSequenceParameterBufferHEVC {
         } bits;
         uint32_t value;
     } scc_fields;
+	
+	/** \brief Lookahead depth in frames. */
+    uint8_t     lookahead_depth;
```
XinfengZhang, line 343: so , maximum depth is 256?

XinfengZhang:
@splotnikv , this interface is not very straight forward, could you file PRs in https://github.com/intel/media-driver & https://github.com/intel/libva-utils to implement the reference code?

PR #414, opened 2020-05-13
Adding interpolation method for scaling. 

FurongZhang:
Adding interpolation mode for scaling: nearest, bilinear, bicubic, lanczos, super resolution.

PR #415, opened 2020-05-13
Add interpolation method for scaling.

FurongZhang:
Signed-off-by: FurongZhang <furong.zhang@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -256,6 +256,13 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_FILTER_SCALING_NL_ANAMORPHIC 0x00000300
 #define VA_FILTER_SCALING_MASK          0x00000f00
 
+/** Interpolation method for scaling */
+#define VA_FILTER_INTERPOLATION_DEFAULT                    0x00000000
+#define VA_FILTER_INTERPOLATION_NEAREST_NEIGHBOR           0x00001000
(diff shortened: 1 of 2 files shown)


XinfengZhang:
duplicate with #414 ?

va/va.h
```diff
@@ -256,6 +256,13 @@ typedef int VAStatus;	/** Return status type from functions */
 #define VA_FILTER_SCALING_NL_ANAMORPHIC 0x00000300
 #define VA_FILTER_SCALING_MASK          0x00000f00
 
+/** Interpolation method for scaling */
+#define VA_FILTER_INTERPOLATION_DEFAULT                    0x00000000
```
XinfengZhang, line 260: should you add some comments here? https://github.com/intel/libva/blob/master/va/va_vpp.h#L971

FurongZhang: What kinds of comments? Here, I have document that for interpolation method.

XinfengZhang: you will use it just like https://github.com/intel/libva/blob/master/va/va_vpp.h#L971, right?

FurongZhang: @XinfengZhang yes, I agreed with you. Will add some comments in filter_flags.

FurongZhang: Thanks @XinfengZhang for the comments. Updated.

PR #416, opened 2020-05-14
add attributes for context priority setting

XinfengZhang:
application could set priority for each VAContext
then HW will check the priority for the tasks related with the VAContext

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -783,6 +783,10 @@ typedef enum
      * implementation, multiple frames encode/decode can improve HW concurrency
      */
     VAConfigAttribMultipleFrame         = 40,
+    /** \brief priority setting for the context. Read-Write
+     *  attribute value is \c VAConfigAttribValContextPriority
+     */
+    VAConfigAttribContextPriority       = 41,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
@@ -1140,6 +1144,18 @@ typedef union _VAConfigAttribValMultipleFrame {
     } bits;
     uint32_t value;
 }VAConfigAttribValMultipleFrame;
+
+/** brief Attribute value VAConfigAttribValContestPriority */
+typedef union _VAConfigAttribValContextPriority{
+    struct{
+        /** \brief the priority , for the Query operation (read) it represents highest priority
+         * for the set operation (write), value should be [0~highest priority] , 0 is lowest priority*/
+        uint32_t priority     :16;
+        /** \brief reserved bits for future, must be zero*/
+        uint32_t reserved     :16;
+    }bits;
+    uint32_t value;
+}VAConfigAttribValContextPriority;
 
 /** @name Attribute values for VAConfigAttribProcessingRate. */
 /**@{*/
--- a/va/va_str.c
+++ b/va/va_str.c
@@ -129,6 +129,7 @@ const char *vaConfigAttribTypeStr(VAConfigAttribType configAttribType)
     TOSTR(VAConfigAttribMaxFrameSize);
     TOSTR(VAConfigAttribPredictionDirection);
     TOSTR(VAConfigAttribMultipleFrame);
+    TOSTR(VAConfigAttribContextPriority);
     case VAConfigAttribTypeMax: break;
     }
     return "<unknown config attribute type>";


va/va.h
```diff
@@ -781,6 +781,11 @@ typedef enum
      * implementation, multiple frames encode/decode can improve HW concurrency
      */
     VAConfigAttribMultipleFrame         = 40,
+    /** \brief priority setting for the context. Read-Write
+     *  attribute value is \c VAConfigAttribValContextPriority
+     *  when read the value, it will return the range of priorty values
+     */
+    VAConfigAttribContextPriority       = 41,
```
FurongZhang, line 788: Is there any value for VAConfigAttribContextPriority? High? Medium and Low? For example, VAConfigAttribRateControl has the below values: VA_RC_NONE, VA_RC_CBR, etc.

XinfengZhang: @FurongZhang , KMD could set from -1023 to 1023. so it is range, driver will report the range from 0, maybe 0 to 2046, , application could set any value from 0 to 2046, then it is converted to -1023 to 1023.  maybe for other vendor , it will be different, we are considering whether 3 or 5 level is enough.

FurongZhang: Out of curiosity, why we don't align with KMD -1023 to 1023?

va/va.h
```diff
@@ -1137,6 +1142,17 @@ typedef union _VAConfigAttribValMultipleFrame {
     uint32_t value;
 }VAConfigAttribValMultipleFrame;
 
+/** brief Attribute value VAConfigAttribValContestPriority */
+typedef union _VAConfigAttribValContextPriority{
+    struct{
+        /** \brief the priority , for the Query operation it will return the range of priority */
```
dmitryermilov, line 1148: I assume this "for the Query operation it will return the range of priority" sounds weird. We don't have **ranges** here, we have only **one value**. If it's 'max' I propose to say it explicitly in the code comment that range is between [0, priority].

va/va.h
```diff
@@ -1137,6 +1142,17 @@ typedef union _VAConfigAttribValMultipleFrame {
...
+        uint32_t priority     :16;
```
dmitryermilov, line 1149: I assume this "for the Query operation it will return the range of priority" sounds weird. We don't have ranges here, we have only one value. If it's 'max' I propose to say it explicitly in the code comment that range is between [0, priority].

hongbolv:
@XinfengZhang When will we have the final decision about the interface definition?

va/va.h
dvrogozh: echo @dmitryermilov. It is not clear what range of priority means and how it can be queried? We have couple of choices here:
1. Declare in VAAPI spec that we support range of priorities as [0, max] where 0 is lowest priority, max is a highest and this highest value can be returned by the query. Downside is that this will be VAAPI priority declaration which driver will need to map to OS specifics.
2. Same as 1), but declare [min, max]. I this case we need to expose 2 attributes, one for min another for max
3. Again, have [min, max], but have structure for that (or different bits for min and for max)

I would go with 1).

hongbolv: In the graphics driver (mesa), the priority is defined three priority (LOW(-512), DEFAULT(0), HIGH(512)). We should consider the alignment between media and graphic, even computing for some user scenario.

dvrogozh: And KMD has -1023 to 1023 as I heard. I wonder how mesa maps to KMD in this case? is the strict = and they effectively miss part of the range? or the remap 512 to 1023?

We are not mesa, we have own interface and should consider what aligns with our own API style and what makes sense to us. Whenever possible we eventually should align with others, but max priority makes more sense to me in case of libva.

hongbolv: In the mesa, it just map LOW/MEDIUM/HIGH to three KMD priority values (-512/0/512). You could say it dropped the range, just picked three typical values.
#define GEN_CONTEXT_LOW_PRIORITY ((I915_CONTEXT_MIN_USER_PRIORITY-1)/2)
#define GEN_CONTEXT_MEDIUM_PRIORITY (I915_CONTEXT_DEFAULT_PRIORITY)
#define GEN_CONTEXT_HIGH_PRIORITY ((I915_CONTEXT_MAX_USER_PRIORITY+1)/2)

Yes, we are own the interface. In the MediaSDK, we also use three priority types other than a range for the user. Itself is already aligned with Mesa definition. 
Back to VA interface, it become a simple question.
1) expose three typical priority values to user, like mesa.
     Less capability for user, more straightforward, low risk when resource competing across component. 
2) expose a range to user, user can define their priority strategy. 
     More capability for user, user should carefully define the priority value.

dvrogozh: Thank you for insight, @hongbolv. One more question: do I understand correctly that -1023, 1023 range is i915 defined one. So, some other KMD may potentially have some other range?

FurongZhang: Personally prefer to expose a range [min, max] to the VAAPI user.

hongbolv: Yes, the  range (-1023, 1023) is i915 defined.

va/va.h
XinfengZhang: it depend on driver implementation, driver could report any range from 0

va/va.h
dmitryermilov: This addressed my previous concern:
•	for the Query operation (read) it represent highest priority
•	for the set operation (write), value should be [0~highest priority] , 0 is lowest priority.

Thank you.

va/va.h
```diff
@@ -1141,6 +1145,18 @@ typedef union _VAConfigAttribValMultipleFrame {
     uint32_t value;
 }VAConfigAttribValMultipleFrame;
 
+/** brief Attribute value VAConfigAttribValContestPriority */
+typedef union _VAConfigAttribValContextPriority{
+    struct{
+        /** \brief the priority , for the Query operation (read) it represent highest priority
```
dmitryermilov, line 1151: it represent -> it represents

PR #418, opened 2020-05-20
Fixes file descriptor leak

gstark:
If code calls va_isDRI2Connected multiple times there is no protection
against calling open(device_name, O_RDWR) again and losing the reference
to the original file descriptor.

A recent change (bc8a12) calls the method twice each time libva is
initialized, thus causing a leak of device file descriptors.

This change moves the initializing of the fd and auth_type to the
creation of the dri_state and then checks to see if they have already
been initialized. If so, the method va_isDRIConnected returns true
and fills in driver_name.

### DIFF
--- a/va/x11/dri2_util.c
+++ b/va/x11/dri2_util.c
@@ -182,8 +182,7 @@ va_isDRI2Connected(VADriverContextP ctx, char **driver_name)
     char *device_name = NULL;
     drm_magic_t magic;        
     *driver_name = NULL;
-    dri_state->base.fd = -1;
-    dri_state->base.auth_type = VA_NONE;
+
     if (!VA_DRI2QueryExtension(ctx->native_dpy, &event_base, &error_base))
         goto err_out;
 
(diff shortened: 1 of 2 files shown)


gstark:
Here is a very small app that will demonstrate the file descriptor leak:

```c
#include <stdarg.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>
#include <getopt.h>
#include <va/va_str.h>

#include "va_display.h"

int open_and_close()
{
  VADisplay va_dpy;
  VAStatus va_status;
  int major_version, minor_version;
  int ret_val = 0;

  va_dpy = va_open_display();
  if (NULL == va_dpy)
  {
      return 2;
  }

  va_status = vaInitialize(va_dpy, &major_version, &minor_version);
  if (va_status != VA_STATUS_SUCCESS) {
      ret_val = 3;
  }

  vaTerminate(va_dpy);
  va_close_display(va_dpy);

  return ret_val;
}

int main(int argc, const char* argv[])
{
    for(;;)
    {
        open_and_close();
    }
}
```

PR #419, opened 2020-05-23
Remove the execute bit from all source code files

primeos:
C source code and header files shouldn't be executable. When building
libva via Meson the affected files would be installed with the execute
file mode bit set (with GNU Autotools this wasn't the case).

To check: find -name '*.[ch]' -executable

Signed-off-by: Michael Weiss <dev.primeos@gmail.com>


primeos:
I also just added a Travis CI check to spot such problems in advance. E.g. without the fix (first commit) the build would fail with the following error message:
https://travis-ci.org/github/primeos/libva/builds/690399141#L588

PR #433, opened 2020-07-07
Move installed backend headers up

ceyusa:
In order to use these headers in an uninstalled environment, such
as gst-build, the exported headers should be achievable in the
same directory, as installed, but relative to the repository.


uartie:
With your gst-build, is it possible to just add the extra include search paths for your uninstalled use-case?

dvrogozh:
> In order to use these headers in an uninstalled environment

Can you, please, clarify why this is needed?

Issue #431, opened 2020-07-02
Deprecate tg_start and tg_end in VASliceParameterBufferAV1


rosetta-jpn:
VASliceParameterBufferAV1 is a structure for a single Tile.
tg_start and tg_end are unnecessary but just confusing.
They should be deprecated.

XinfengZhang:
thanks, @rosetta-jpn , different HW vendor maybe has different implementation, not sure about whether it should be deprecated now, lets wait feedback from different channel.

rosetta-jpn:
It was done in https://github.com/intel/libva/commit/1d807f7a77c3b517e1674d8ff201ade93a931a3d.

PR #435, opened 2020-07-17
Fix errors in av1 decoder interfaces

rosetta-jpn:
Signed-off-by: Hirokazu Honda <hiroh@chromium.org>

### DIFF
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -364,7 +364,11 @@ typedef struct  _VADecPictureParameterBufferAV1
     uint8_t                 tile_cols;
     uint8_t                 tile_rows;
 
-    /* specifes the width/height of a tile minus 1 in units of superblocks */
+    /* The width/height of a tile minus 1 in units of superblocks. Though the
+     * maximum number of tiles is 64, since ones of the last tile are computed
+     * from ones of the other tiles and frame_width/height, they are not
+     * necessarily specified.
+     */
     uint16_t                width_in_sbs_minus_1[63];
     uint16_t                height_in_sbs_minus_1[63];
 
@@ -425,7 +429,7 @@ typedef struct  _VADecPictureParameterBufferAV1
     uint8_t                 superres_scale_denominator;
 
     /** \brief Interpolation filter.
-     *  value range [0..9]
+     *  value range [0..4]
      */
     uint8_t                 interp_filter;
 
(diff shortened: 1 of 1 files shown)


va/va_dec_av1.h
```diff
@@ -365,8 +365,8 @@ typedef struct  _VADecPictureParameterBufferAV1
     uint8_t                 tile_rows;
 
     /* specifes the width/height of a tile minus 1 in units of superblocks */
-    uint16_t                width_in_sbs_minus_1[63];
-    uint16_t                height_in_sbs_minus_1[63];
+    uint16_t                width_in_sbs_minus_1[64];
```
XinfengZhang, line 368: it will break backward compatibility

rosetta-jpn: Should I decrease this by 1? https://github.com/intel/libva/blob/master/va/va_dec_av1.h#L650

dvrogozh: This won't work and still breaks backward compatibility. You can not add fields or increase size of arrays in the middle of the structure. You need to add fields in the very end taking out space from va_reserved. 

Why exactly do we need this change?

cwang105:
why do we want to change width_in_sbs_minus_1[63] to width_in_sbs_minus_1[64]?
The max number of tile columns is 64. So the width of last tile column can always be derived from frame width, no?

Thanks,
Ce

rosetta-jpn:
It sounds good to me if the driver computes tile_width as sb columns from them.
It was not obvious that a driver needs to compute it from the comment.

XinfengZhang:
@rosetta-jpn , to avoid backward compatible issue , change the comments is enough, right? will you change it?

rosetta-jpn:
Sure. Updated.

XinfengZhang:
@rosetta-jpn , could you help to squash two commits together.

rosetta-jpn:
> @rosetta-jpn , could you help to squash two commits together.

Done!

XinfengZhang:
LGTM, @cwang105 @dvrogozh @ZefuLI if there are no other concern, will merge it. thanks

rosetta-jpn:
ping other reviewers.

By the way, I would like to ask you all if marking deprecated to tg_start and tg_end is a correct thing.
https://github.com/intel/libva/issues/431#issuecomment-652735631

cwang105:
Looks good to me.

Thanks,
Ce

XinfengZhang:
from my personal understanding, tg_start and tg_end should not be deprecated now. 
yes, https://github.com/intel/media-driver does not care about it at all, but the interface is designed not only for Intel. maybe some underlying driver need these parameters. I am not whether these parameter is independent and will not impact implementation.

rosetta-jpn:
Sorry for late reply.

> from my personal understanding, tg_start and tg_end should not be deprecated now.
tg_start and tg_end were confusing for me at least.
Since there are tg_column and tg_row, looks like its structure should be used for a single tile.
But tg_start and tg_end are contradicts to it.

Yes, Deprecated might not be correct.
Shall we clarify this by some comment in the structure? Any suggestion about the comment?

rosetta-jpn:
Gentle ping

XinfengZhang:
@rosetta-jpn , sorry I missed the notification mail. I am fine with current change, if no other concern, I will merge it , thanks

PR #440, opened 2020-08-19
add misc parameter to adjust context priority dynamically

XinfengZhang:

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -792,6 +792,7 @@ typedef enum
     VAConfigAttribMultipleFrame         = 40,
     /** \brief priority setting for the context. Read-Write
      *  attribute value is \c VAConfigAttribValContextPriority
+     *  this setting also could be update by \c VAContextParameterUpdateBuffer
      */
     VAConfigAttribContextPriority       = 41,
     /**@}*/
@@ -1152,7 +1153,7 @@ typedef union _VAConfigAttribValMultipleFrame {
     uint32_t value;
 }VAConfigAttribValMultipleFrame;
 
-/** brief Attribute value VAConfigAttribValContestPriority */
+/** brief Attribute value VAConfigAttribValContextPriority */
 typedef union _VAConfigAttribValContextPriority{
     struct{
         /** \brief the priority , for the Query operation (read) it represents highest priority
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -2547,6 +2549,23 @@ typedef struct _VAEncMiscParameterCustomRoundingControl
         uint32_t    value;
     }   rounding_offset_setting;
 } VAEncMiscParameterCustomRoundingControl;
+
```
dmitryermilov, line 2552: I assume it misses a written description of purpose of the structure and its usage.

va/va.h
```diff
@@ -2547,6 +2549,23 @@ typedef struct _VAEncMiscParameterCustomRoundingControl
...
+{
+    union{
+        struct {
+            /** \brief indicate whether context priority changed */
+            uint32_t context_priority_update :1;
+            uint32_t reserved                :31;
+        } bits;
+        uint32_t value;
+    }flags;
+    /** \brief task priority adjustment dynamically */
```
dmitryermilov, line 2563: What does dynamically  mean here?

va/va.h
XinfengZhang: accept

va/va.h
```diff
@@ -1878,7 +1878,9 @@ typedef enum
     /** \brief Buffer type used for FEI input frame level parameters */
     VAEncMiscParameterTypeFEIFrameControl = 18,
     /** \brief encode extension buffer, ect. MPEG2 Sequence extenstion data */
-    VAEncMiscParameterTypeExtensionData = 19
+    VAEncMiscParameterTypeExtensionData = 19,
+    /** \brief adjust context parameters dynamically, refer \c VAEncMiscParameterContextParameterAdjustment */
+    VAEncMiscParameterTypeContextParameterAdjust = 20
```
hongbolv, line 1883: The context priority is a general parameter for decoder, encoder, vpp.  The value name started with "VAEncMiscParameter" will cause the confusion.

XinfengZhang: accept

va/va.h
hongbolv: It means the user can change priority per frame other than per stream.

hongbolv (approved):

dmitryermilov (approved):

PR #441, opened 2020-08-21
Refine the va_TraceVAPictureParameterBufferAV1.

xhaihao:
Signed-off-by: He Junyan <junyan.he@intel.com>


rosetta-jpn:
Would you mind what this patch is refined concretely?
I wonder there are missing parts formerly.

HeJunyan:
It add all missing fields such as film_grain_info, seg_info, loop_filter_info_fields and wm. All these fields are not printed out in current version. We also reorder all the fields to the VADecPictureParameterBufferAV1 declaring order.

rosetta-jpn:
Thanks. I actually tried tracing with this patch. It looks obviously better than original.

PR #442, opened 2020-09-03
Add fourcc values for YUV 4:4:4 formats

weizhu-intel:
The Intel driver for gen11+ needs these formats for HEVC codecs in YUV 4:4:4
at 8- and 10-bit depth.  The names are taken from libdrm.

Signed-off-by: Mark Thompson <sw@jkqxz.net>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -3957,6 +3957,11 @@ VAStatus vaSyncBuffer(
  * Four bytes per pixel: A, Y, U, V.
  */
 #define VA_FOURCC_AYUV          0x56555941
+/** XYUV packed 8-bit YUV 4:4:4
+ *
+ * Four bytes per pixel: unspecified, Y, U, V.
+ */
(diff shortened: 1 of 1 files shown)


XinfengZhang:
@fhvwy Hi Mark , the origin patch is from you, any concern with this patch , if no concern , I will merge it

XinfengZhang (changes requested):
two concerns:
1. permission from Mark
2. if ffmpeg-vaapi map XYUV  to 0YUV, MSDK also need to change to XYUV. Or, there are different mapping, ffmpeg-vaapi XYUV to 0YUV, ffmpeg-QSV  AYUV to 0YUV. 
before these two concern was addressed. we should not merge the code.

fhvwy:
Applying this is fine with me.
On the original version I did wonder whether there are better fourccs for the 10/16-bit versions - I copied XV30/XV48 from libdrm, but I have never seen them used anywhere else.

PR #443, opened 2020-09-04
Added multi GPU device suppport on Android

TangZhiZhen:
Fixed compilation issue

Signed-off-by: Zhizhen Tang <zhizhen.tang@intel.com>


va/Android.mk
```diff
@@ -25,8 +25,8 @@
 
 LOCAL_PATH:= $(call my-dir)
 
-LIBVA_DRIVERS_PATH_32 := /vendor/lib/dri
-LIBVA_DRIVERS_PATH_64 := /vendor/lib64/dri
+LIBVA_DRIVERS_PATH_32 := /vendor/lib/:/system/lib
```
XinfengZhang, line 28: I wonder whether it will impact old setting

TangZhiZhen: none impact, because it will use /system/lib or /system/lib64 to find share library，I can add dri, but it can't used on android.

XinfengZhang:
@xuguangxin , could you help to review this pull request. it add new function to select different HW

va/android/va_android.cpp
```diff
@@ -167,6 +158,68 @@ VADisplay vaGetDisplay (
...
+
+    drm_state->auth_type = VA_DRM_AUTH_CUSTOM;
+
+    pDriverContext->drm_state = drm_state;
+
+    return (VADisplay)pDisplayContext;
+}
+
+VADisplay vaGetDisplay2 (
+    void *native_dpy, /* implementation specific */
```
dvrogozh, line 177: It seems that originally `native_dpy` was supposed to be used to distinguish different devices. From this perspective you introduce a function with 2 parameters dedicated for the same.

Can you, please, clarify how do you currently use `native_dpy`. What's that, how do you initialize it?

dvrogozh (changes requested):

va/android/va_android.cpp
TangZhiZhen: Here's the mediasdk code(samples/sample_common/src/vaapi_utils_android.cpp):
typedef unsigned int vaapiAndroidDisplay;

#define VAAPI_ANDROID_DEFAULT_DISPLAY 0x18c34078

AndroidLibVA::AndroidLibVA(void)
    : CLibVA(MFX_LIBVA_AUTO)
    , m_display(NULL)
{
    VAStatus va_res = VA_STATUS_SUCCESS;
    mfxStatus sts = MFX_ERR_NONE;
    int major_version = 0, minor_version = 0;
    vaapiAndroidDisplay* display = NULL;

    m_display = display = (vaapiAndroidDisplay*)malloc(sizeof(vaapiAndroidDisplay));
    if (NULL == m_display) sts = MFX_ERR_NOT_INITIALIZED;
    else *display = VAAPI_ANDROID_DEFAULT_DISPLAY;

    if (MFX_ERR_NONE == sts)
    {
        m_va_dpy = vaGetDisplay(m_display);
......

dvrogozh: I was always wondering what this magic `0x18c34078` means? Basically, can Android implementation be reviewed here? What native_dpy magic value actually brings? Can this be properly defined (for example just make it real /dev/dri/renderD128)?

xuguangxin: Hi @dvrogozh , 
it's a legacy interface, android driver will open renderD128 for caller.

Hi @TangZhiZhen
Instead of add new interface, could we use vaGetDisplayDRM? 
Two thing you may need check:
1. do we have libdrm on android, I remember libva need this for drm backend ?
2. Do we have issue when we share surface between decode and SurfaceFlinger?

TangZhiZhen: do we have libdrm on android, I remember libva need this for drm backend ?
[Zhizhen: ]According to android.mk file, it only generated two library (libva.so and libva_android.so) in android.

Do we have issue when we sharing surface between decoder and SurfaceFlinger?
[Zhizhen: ] no issue, we can sharing surface between decoder and surfaceFlinger by gralloc.

dvrogozh: > Hi @dvrogozh ,
> it's a legacy interface, android driver will open renderD128 for caller.

I understand that, but we actually hardcode 0x18c34078 to correspond to renderD128. Is there any Android API to get render node from the legacy native dpy handle which 0x18c34078 looks to be? If Android don't go to provide such an API, I don't quite get why we need the legacy API at the first place since we already have vaGetDisplayDRM which covers what's needed.

> [Zhizhen: ]According to android.mk file, it only generated two library (libva.so and libva_android.so) in android

Android.mk can be fixed to generate other targets if needed.

xuguangxin: > I understand that, but we actually hardcode 0x18c34078 to correspond to renderD128. Is there any Android API to get render node from the legacy native dpy handle which 0x18c34078 looks to be? If Android don't go to provide such an API, I don't quite get why we need the legacy API at the first place since we already have vaGetDisplayDRM which covers what's needed.

remove API will create backward combability issues.  
msdk omx, hw composer, maybe more users still used this.

xuguangxin: add vaGetDisplayDRM maybe a more reasonable solution

dvrogozh: I did not mean to remove the API, sorry for confusion. But my point is:
1. You already have vaGetDisplay(void* native_dpy) API on android which has capability to accept device identification. Consider extending it thu the work with native_dpy. Something like (and this is not VAAPI):

```
void* android_get_native_display(const char* dev); // accepts /dev/dri/cardN|/dev/dri/renderDN, returns native display
const char* android_get_drm_device(void* native_dpy); // accepts native display, returns device
```

2. If not possible for any reason, consider to use vaGetDisplayDRM. But in this case I think we need to consider to deprecate vaGetDisplay.

xuguangxin: Hi @dvrogozh , 
@TangZhiZhen plan to use ffmpeg on android, with approach 1 we need modify ffmpeg code to make it runnable on android. 
So, instead of add a new API, I suggest @TangZhiZhen add  vaGetDisplayDRM back to android build. It Just like your suggestion on approach 2.  and we do not need change ffmpeg code. 

 @TangZhiZhen, could you help do this change in android? or you can ask Hongfu/ @LukeZheZhu for help.

dvrogozh: > So, instead of add a new API, I suggest @TangZhiZhen add vaGetDisplayDRM back to android build. 

I am fully ok with that.

Issue #454, opened 2020-10-07
meson and autotool versions mismatch


dvrogozh:
For meson build system we really have version as 2.9.0.0 while in autotools we have just 2.9.0 (if I read autotool scripts correctly):
https://github.com/intel/libva/blob/2.9.0/meson.build#L10
https://github.com/intel/libva/blob/2.9.0/configure.ac#L50

onabiull:
And since https://github.com/intel/libva/commit/d9676538e84319d6954f7d6fd70be0f3073d044f
Autotools report API version 1.10.0, and meson - 1.9.0, discovered in https://github.com/Intel-Media-SDK/MediaSDK/pull/2363

dvrogozh:
> Autotools report API version 1.10.0, and meson - 1.9.0

@onabiull : I am taking care of this in #464. Please, review.

PR #455, opened 2020-10-09
correct the version for meson build 

XinfengZhang:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@
 # - reset micro version to zero when VA-API major or minor version is changed
 project(
   'libva', 'c',
-  version : '2.9.0.0',
+  version : '2.9.0',
   meson_version : '>= 0.37.0',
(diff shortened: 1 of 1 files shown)


XinfengZhang:
it fix #454

meson.build
```diff
@@ -7,7 +7,7 @@
 # - reset micro version to zero when VA-API major or minor version is changed
 project(
   'libva', 'c',
-  version : '2.9.0.0',
+  version : '2.9.0',
```
dvrogozh, line 10: can you bump the version to become 2.9.1? Just removing ending .0 without bumping the version might cause issues with package managers on version check (which package has greater version 2.9.0 vs. 2.9.0.0).

dvrogozh (changes requested):

dvrogozh:
I think this one will be better: #465

PR #464, opened 2020-10-23
meson: automatically correlated project version and api version number

dvrogozh:
In configure.ac we have project version correlated with api version
in automated way. This was not done for meson and we start to see
project support issues on this regard. Let's try to automate that
and avoid these issues in the future.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,8 +1,9 @@
 # libva package version number, (as distinct from shared library version)
 # XXX: we want the package version to remain at 1.0.x for VA-API 0.32.y
 #
-# - major version is automatically generated from VA-API major version
-# - minor version is automatically generated from VA-API minor version
+# libva package version and VA-API version number are related:
+# - package major version is automatically generated from VA-API major version (w/ +1)
+# - package minor version is automatically generated from VA-API minor version (as =)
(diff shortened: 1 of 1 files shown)


onabiull (changes requested):
It should go vice-versa. Binary version should be derived from the API. Could you please correct?

XinfengZhang:
yes, api version is base,  libva is implementation of vaapi definition
please refer https://github.com/intel/libva/blob/master/configure.ac#L65

dvrogozh:
Let me try. I am not expert in meson and not sure whether it will accept variable in the `project()` statement declaration. But I might be just frightened by cmake. So, I'll try...

dvrogozh:
@XinfengZhang. @onabiull : this won't work(. Unfortunately meson requires `project()` to be the very first statement. It does not accept variable declarations before it. Error is:
```
ERROR: First statement must be a call to project
```

So, if we want some automated correlation for meson between package and api versions we need to play from package version for meson. I.e. this PR as it stands now. Do we want that?

ceyusa:
For meson the project version is a first class element for compilation and distribution. API version is specific to libva.

PR #468, opened 2020-10-27
Add a configuration attribute type for look ahead encoding

JeremyShang:
Signed-off-by: Jeremy Shang jeremy.shang@intel.com

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -795,6 +795,12 @@ typedef enum
      *  this setting also could be update by \c VAContextParameterUpdateBuffer
      */
     VAConfigAttribContextPriority       = 41,
+    /** \brief Look ahead encoding is to add another encoding pass to look ahead several frames before normal encode pass.
+     *  Lookahead pass collect encoding statistics and encode pass analyze the statistics to optimize encoding parameters.
+     *  If capability reports this feature can be supported, application can query capability for max frames to look ahead,
+     *  then set look ahead depth no more than that max value to enable look ahead encoding.
+     */
+    VAConfigAttribEncLookAhead          = 42,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -1165,6 +1169,26 @@ typedef union _VAConfigAttribValContextPriority{
...
+    struct {
+        /** \brief Number of frames to lookahead. Range is [0~100]. 0 means lookahead disabled.*/
+        uint32_t look_ahead_depth               : 8;
+        /** \brief indicate whether the current encoding is in the Look Ahead pass
+         *  if look_ahead_phase == 0, the current encoding is in the actual encoding pass
+         *  if look_ahead_phase == 1, the current encoding is in the Look Ahead pass.*/
+        uint32_t look_ahead_phase               : 1;
+        /** \brief A flag indicates whether Look Ahead Analysis is supported
+         *  if lookahead analysis is supported, application can ignore encoded bitstream.*/
+        uint32_t look_ahead_analysis_support    : 1;
```
XinfengZhang, line 1183: what is lookahead analysis?

va/va.h
```diff
@@ -1165,6 +1169,26 @@ typedef union _VAConfigAttribValContextPriority{
...
+        /** \brief A flag indicates whether Look Ahead BRC is supported.*/
```
XinfengZhang, line 1184: what is look ahead BRC?

va/va.h
```diff
@@ -795,6 +795,10 @@ typedef enum
      *  this setting also could be update by \c VAContextParameterUpdateBuffer
      */
     VAConfigAttribContextPriority       = 41,
+    /** \brief Add another pass to lookahead several frames before normal encode pass.
```
XinfengZhang, line 798: if there are such capability reported, how to use it?

XinfengZhang (changes requested):

va/va.h
```diff
@@ -1165,6 +1169,26 @@ typedef union _VAConfigAttribValContextPriority{
     uint32_t value;
 }VAConfigAttribValContextPriority;
 
+/** \brief Attribute value for VAConfigAttribLookAhead*/
+typedef union _VAConfigAttribValEncLookAhead {
+    struct {
+        /** \brief Number of frames to lookahead. Range is [0~100]. 0 means lookahead disabled.*/
```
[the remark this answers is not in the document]
JeremyShang: Sure. New comments are added. It's the value for app to set, and the field for maximum value is added(max_num_look_ahead_frames). If 50 is reported , application could set (0,50].

va/va.h
JeremyShang: Updated commit has combined these two caps as one as look_ahead_support.

va/va.h
JeremyShang: Updated commit has combined these two caps as one as look_ahead_support.

va/va.h
JeremyShang: Comments are added. If capability reports this feature can be supported, application can query capability for max frames to look ahead, then set look ahead depth no more than that max value to enable look ahead encoding.

va/va.h
```diff
@@ -1165,6 +1171,26 @@ typedef union _VAConfigAttribValContextPriority{
...
+        uint32_t look_ahead_depth               : 8;
+        /** \brief max num of frames to look ahead before actual encoding pass,
+         *  application can set look_ahead_depth as (0, max_num_look_ahead_frames] */
+        uint32_t max_num_look_ahead_frames      : 8;
+        /** \brief indicate whether the current encoding is in the Look Ahead pass,
+         *  if look_ahead_phase == 0, the current encoding is in the actual encoding pass
+         *  if look_ahead_phase == 1, the current encoding is in the look ahead pass. */
+        uint32_t look_ahead_phase               : 1;
+        /** \brief A flag indicates whether look ahead is supported */
+        uint32_t look_ahead_support             : 1;
```
XinfengZhang, line 1187: look_ahead_support is enough? as I know, there are different output types. so maybe need several bits to indicate it?

PR #470, opened 2020-11-02
va/va_dec_vp9: Added loop filter and Quantization params for VP9 deco…

veeraamd:
Added the quantization and loop filter parameters for vp9 videos.
This will avoid the need for re-parsing of headers in the driver implementation.

Signed-off-by: Veerabadhran Gopalakrishnan <veerabadhran.gopalakrishnan@amd.com>

### DIFF
--- a/va/va_dec_vp9.h
+++ b/va/va_dec_vp9.h
@@ -189,8 +189,22 @@ typedef struct  _VADecPictureParameterBufferVP9
      */
     uint8_t                 bit_depth;
 
+    /** \brief loop filter scemantics
+     */
+    uint8_t                 mode_ref_delta_enabled;
+    uint8_t                 mode_ref_delta_update;
+
+    /** \brief Quantization params
+     */
+    uint8_t                 base_qindex;
+    int8_t                  y_dc_delta_q;
+    int8_t                  uv_ac_delta_q;
+    int8_t                  uv_dc_delta_q;
+
+    /* padding variable*/
+    uint8_t                 va_padding[2];
     /** \brief Reserved bytes for future use, must be zero */
-    uint32_t                va_reserved[VA_PADDING_MEDIUM];
+    uint32_t                va_reserved[VA_PADDING_MEDIUM - 2];
 
 } VADecPictureParameterBufferVP9;
 


veeraamd:
@XinfengZhang Could you please review this commit?

va/va_dec_vp9.h
```diff
@@ -189,8 +189,22 @@ typedef struct  _VADecPictureParameterBufferVP9
      */
     uint8_t                 bit_depth;
 
+    /** \brief loop filter scemantics
+     */
+    uint8_t                 mode_ref_delta_enabled;
+    uint8_t                 mode_ref_delta_update;
```
fhvwy, line 195: (I'm guessing these two fields are intended to be the spec variables `loop_filter_delta_enabled` and `loop_filter_delta_update`, since these are the names of the matching variables in AV1.)
Why?  The loop filter levels are present per-segment in `VASegmentParameterVP9.filter_level`, in which the adjustment fields these are indicating the presence of have already been applied.

va/va_dec_vp9.h
```diff
@@ -189,8 +189,22 @@ typedef struct  _VADecPictureParameterBufferVP9
...
+
+    /** \brief Quantization params
+     */
+    uint8_t                 base_qindex;
+    int8_t                  y_dc_delta_q;
+    int8_t                  uv_ac_delta_q;
+    int8_t                  uv_dc_delta_q;
```
fhvwy, line 202: What are you going to use the top-level quant values for?  They don't do anything in themselves because the segment features can modify them, and the derived per-segment values are present in `VASegmentParameterVP9`.

rosetta-jpn:
What happens if an application doesn't fill these parameters?
Doesn't this break a compatibility of an app behavior?

va/va_dec_vp9.h
veeraamd: In case, the segmentation is not enabled, the top level value is needed.  We saw failures in playback without these values.

veeraamd:
@rosetta-jpn Since we are adding new params, we will not be breaking any existing applications.
We are also planning to take it up with ffmpeg to update these values their lib.

rosetta-jpn:
Once a driver use the filled values, an application that doesn't fill these (probably equivalent to filling zero) is affected, isn't it?

va/va_dec_vp9.h
fhvwy: Can you explain the details of the case you are having problems with?  When segmentation is not enabled all `segment_id`s are zero so all mode info blocks use the `quant_scale` values supplied for segment zero.

sreerenjb: The patch doesn't make sense to me either. It is not mandatory to send the uncompressed header to the driver which means there wouldn't be a way to parse these fields inside the UMD.
Also, please note that the SegParam[0] will have all the required values even if segmentation is disabled.

veeraamd:
@rosetta-jpn @sreerenjb @fhvwy
Hi Everyone,
After multiple discussions with other teams, we learnt that the top level quantization and loop filter values are required for proper decoding of VP9 streams by the AMD driver and firmware.
The top level quantization and loop filter values for VP9 are already present in the vdpau headers.
Therefore I request to include these parameters.
I agree about the problem with the applications using the older version of the headers and filling these values.
What do you suggest to tackle this problem?

veeraamd:
All,
@rosetta-jpn @sreerenjb @fhvwy
Could you please update on the request to add the parameters.

rosetta-jpn:
@veeraamd , Sorry for my slow response.

> After multiple discussions with other teams, we learnt that the top level quantization and loop filter values are required for proper decoding of VP9 streams by the AMD driver and firmware.
> The top level quantization and loop filter values for VP9 are already present in the vdpau headers.

As @sreerenjb said in the comment, in the case segmentation is disabled, the top segmentation value is filled in VASliceParameterBufferVP9::seg_param[0].
Does the driver read them? Could you share with us how you concluded with your team?

veeraamd:
Hi everyone, 
Thanks for all the inputs and review comments.
We are taking this up in the firmware.
Closing this pull request.

PR #472, opened 2020-11-10
Pass offset and size of pred_weight_table

Tianhaol:
These two parameters are gotten from packed slice header, are used
to write  pred_weight_tabe into packed slice header
of the final bit stream.


XinfengZhang:
could you add more comments inside code , and add more description in the commit message. especially please describe 3 usage clearly. how to enable force weight prediction from application , how to disable it, how to enable driver adaptive  weight prediction.

XinfengZhang (approved):

PR #474, opened 2020-11-18
[AV1] Update documented ranges for loop filter and quantization params.

andrescj-chromium:
Currently, the following fields of _VADecPictureParameterBufferAV1 are
documented as having a range of [-63, 63]:

ref_deltas
mode_deltas
y_dc_delta_q
u_dc_delta_q
u_ac_delta_q
v_dc_delta_q
v_ac_delta_q

This is incorrect because per the spec, these parameters are parsed
using su(1+6) which yields a range of [-64, 63] (see sections 4.10.6,
5.9.11, 5.9.12, and 5.9.13). This patch fixes the documented ranges.

### DIFF
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -463,13 +463,13 @@ typedef struct  _VADecPictureParameterBufferAV1
 
     /** \brief The adjustment needed for the filter level based on
      *  the chosen reference frame.
-     *  value range [-63..63].
+     *  value range [-64..63].
      */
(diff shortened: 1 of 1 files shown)


sreerenjb (approved):
Thank you for the patch.
LGTM

sreerenjb:
@XinfengZhang @ZefuLI

XinfengZhang:
LGTM

ZefuLI:
LGTM

XinfengZhang (approved):

PR #475, opened 2020-12-01
Video processing HVS Denoise: Added 4 modes including default, auto

FurongZhang:
BDRate, auto Subjective, manual mode.

Signed-off-by: furongzh <furong.zhang@intel.com>


va/va_vpp.h
```diff
@@ -1313,6 +1333,15 @@ typedef struct _VAProcFilterParameterBufferHVSNoiseReduction {
      *  Value 10 is the default value.
      */
     uint16_t            strength;
+    /**
+     *  \brief HVS Denoise Mode which controls denoise method.
+     *  It is a value in VAProcHVSDenoiseMode.
+     *  Value VAProcHVSDenoiseDefault means default mode;
```
dvrogozh, line 1339: I don't think you need to repeat values descriptions here - you have them above within `VAProcHVSDenoiseMode`  declaration.

dvrogozh (dismissed):

va/va_vpp.h
```diff
@@ -1313,8 +1333,13 @@ typedef struct _VAProcFilterParameterBufferHVSNoiseReduction {
      *  Value 10 is the default value.
      */
     uint16_t            strength;
+    /**
+     *  \brief HVS Denoise Mode which controls denoise method.
+     *  It is a value in VAProcHVSDenoiseMode.
+     */
+    uint16_t            mode;
```
dvrogozh, line 1340: @XinfengZhang : what's our approach to specify integer type for enum value?

dvrogozh: Yeah, but if we will have `VAProcHVSDenoiseMode mode`, what will be its size? Generally speaking it's compiler dependent (but I again might be over paranoid).

XinfengZhang: it will be 32bit on both 64bit and 32 bit compiler. but your concern make sense. maybe we could use macro instead of enum to avoid such risk.

FurongZhang:
@XinfengZhang

dvrogozh (approved):

XinfengZhang:
LGTM now

va/va_vpp.h
```diff
@@ -1298,6 +1298,18 @@ typedef struct _VAProcFilterParameterBufferTotalColorCorrection {
     float                             value;
 } VAProcFilterParameterBufferTotalColorCorrection;
 
+/** @name Video Processing Human Vision System (HVS) Denoise Mode.*/
+/**@{*/
+/** \brief Default Mode. This mode is decided in driver to the appropriate mode. */
+#define VA_PROC_HVS_DENOISE_DEFAULT               0
+/** \brief Auto BDRate Mode. */
```
XinfengZhang, line 1305: hi @FurongZhang , could you help to add more comments for these options? from application view, I dont know how to select it

VickyZengg (approved):

PR #476, opened 2020-12-01
wayland: don't try to authenticate with render nodes

emersion:
When the parent compositor is running with a render node, Mesa will send
the render node path via wl_drm instead of the primary path. DRM
authentication doesn't work with render nodes.

This patch makes the Wayland logic consistent with the DRM logic [1].

Tested using Sway running with wlroots' headless backend.

[1]: https://github.com/intel/libva/blob/b4d0af7f43f0ef2d51f070c75429c0d6acca663b/va/drm/va_drm.c#L68

Signed-off-by: Simon Ser <contact@emersion.fr>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -75,7 +75,7 @@ m4_define([libva_lt_age],
           [m4_eval(libva_binary_age - libva_interface_age)])
 
 # libdrm minimun version requirement
-m4_define([libdrm_version], [2.4])
+m4_define([libdrm_version], [2.4.60])
 
 # Wayland minimum version number
 # 1.11.0 for wl_proxy_create_wrapper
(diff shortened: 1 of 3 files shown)


emersion:
@XinfengZhang: gentle ping

XinfengZhang:
sorry for late response, 1 question: is there minimum  version requirement for function drmGetNodeTypeFromFd?

emersion:
That's a good point, it requires 2.4.60. I've updated the patch to bump to version constraint.

emersion:
Ping @XinfengZhang

emersion:
Ping @XinfengZhang

XinfengZhang (approved):

PR #481, opened 2020-12-20
AV1 decode documentation edit for superres_scale_denominator

sushma08:
superres_scale_denominator value was only reflecting range [9..16] which
doesn't match the range mentioned in spec [8..16]. This commit clarifies
that when use_superres=1, superres_scale_denominator value range is
[9..16] otherwise, it is 8.

### DIFF
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -444,7 +444,8 @@ typedef struct  _VADecPictureParameterBufferAV1
     } pic_info_fields;
 
     /** \brief Supper resolution scale denominator.
-     *  value range [9..16]
+     *  When use_superres=1, superres_scale_denominator must be in the range [9..16].
+     *  When use_superres=0, superres_scale_denominator must be 8.
      */
     uint8_t                 superres_scale_denominator;
 


va/va_dec_av1.h
```diff
@@ -444,7 +444,8 @@ typedef struct  _VADecPictureParameterBufferAV1
     } pic_info_fields;
 
     /** \brief Supper resolution scale denominator.
-     *  value range [9..16]
+     *  When use_superres=1, value range [9..16].
+     *  When use_superres=0, denominator value will be 8.
```
andrescj-chromium, line 448: I think this should be clarified in one of two ways:

1) Say that "When use_superres=0, superres_scale_denominator is ignored and the driver shall assume the denominator to be 8".

2) Say that "When use_superres=0, superres_scale_denominator must be 8".

(1) is how the iHD driver currently implements it per your comments. (2) would be safer in order to handle any other driver that might have implemented AV1 decoding already (if any).

andrescj-chromium (commented):
Thanks for this! I have a suggestion.

va/va_dec_av1.h
sushma08: Thank you for the suggestion Andres. I will go with (2).

va/va_dec_av1.h
```diff
@@ -444,7 +444,8 @@ typedef struct  _VADecPictureParameterBufferAV1
     } pic_info_fields;
 
     /** \brief Supper resolution scale denominator.
-     *  value range [9..16]
+     *  When use_superres=1, super_scale_denominator range [9..16].
```
andrescj-chromium, line 447:
```suggestion
     *  When use_superres=1, superres_scale_denominator must be in the range [9..16].
```

va/va_dec_av1.h
```diff
@@ -444,7 +444,8 @@ typedef struct  _VADecPictureParameterBufferAV1
...
+     *  When use_superres=0, super_scale_denominator must be 8.
```
andrescj-chromium, line 448:
```suggestion
     *  When use_superres=0, superres_scale_denominator must be 8.
```

andrescj-chromium (changes requested):
Thanks! This looks good to me (for what it's worth) % typo.

va/va_dec_av1.h
sushma08: Done :)

andrescj-chromium (approved):

va/va_dec_av1.h
```diff
@@ -444,7 +444,8 @@ typedef struct  _VADecPictureParameterBufferAV1
     } pic_info_fields;
 
     /** \brief Supper resolution scale denominator.
-     *  value range [9..16]
+     *  When use_superres=1, superres_scale_denominator must be in the range [9..16].
+     *  When use_superres=0, superres_scale_denominator must be 8.
```
XinfengZhang, line 448: when use_superres==0 should has  same effect with superres_scale_denominator == 8.
they should have same meaning.  if (use_superres == 0 || supperres_scale_denominator == 8) supper res is disabled, right?

sushma08: Whenever, use_superres = 0, supperres_scale_denominator will be set to 8 Carl. From driver code - cmd.DW9.SuperResDenom = m_picParams->PicFlags.fields.use_superres ? params->m_picParams->superres_scale_denominator : 8;

XinfengZhang: I mean, there are already use_supperes == 0 which is similiar with superres_scale_denominator ==8. [8..16] & [9..16] looks same. application could get expected result from the combination of these two varaible , only difference is [8..16] is aligned with spec, and it is straightforward, right?

sushma08: Our intent is to make sure that we show Super resolution scale denominator value range is [8..16] as per AV1 spec. However, the extra details provide information to the reader on when we use 8 vs [9..16].

XinfengZhang:
LGTM, there are no compatible issue, even the value range changed, because there are another flag as a switch.

PR #485, opened 2020-12-31
disable travis build

XinfengZhang:
already enable github actions


dvrogozh:
It seems Coverity scan is the only thing which is available in travis and is missing in our github actions setup. Everything else, i.e. ubuntu and freebsd builds, are there in github actions now. 

Thus the question is: do we need Coverity scan for the libva? Badge on the top page seems to be broken (goes to https://scan.coverity.com/projects/intel-libva which gives 404 error to me). So, I am even not sure if the scan is still alive. If it is not, yep, let's disable travis.

XinfengZhang:
the coverity scan should base on branch https://github.com/intel/libva/tree/coverity_scan, not master, we did not update the branch long time

dvrogozh (approved):

dvrogozh:
@XinfengZhang : I am going to merge this in.

XinfengZhang:
sure, no problem.

PR #487, opened 2021-01-05
Remove tile_rows and tile_cols restriction to match AV1 spec

sushma08:
There is no restriction of tile_rows and tile_cols value in AV1 spec so,
this commit excludes the comment that says value of tile_rows and
tile_cols should be power of 2.


sushma08:
Can we please merge this PR? Thank you.

XinfengZhang (approved):

PR #492, opened 2021-01-12
Add 3DLUT Filter in Video Processing.

FurongZhang:

### DIFF
--- a/va/va_vpp.h
+++ b/va/va_vpp.h
@@ -253,6 +253,8 @@ typedef enum _VAProcFilterType {
     VAProcFilterHVSNoiseReduction,
     /** \brief High Dynamic Range Tone Mapping. */
     VAProcFilterHighDynamicRangeToneMapping,
+    /** \brief Three-Dimensional Look Up Table (3DLUT). */
+    VAProcFilter3DLUT,
     /** \brief Number of video filters. */
     VAProcFilterCount
 } VAProcFilterType;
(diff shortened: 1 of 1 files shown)


va/va_vpp.h
```diff
@@ -1333,6 +1335,88 @@ typedef struct _VAProcFilterParameterBufferHDRToneMapping {
...
 } VAProcFilterParameterBufferHDRToneMapping;
 
+/** @name 3DLUT Channel Layout and Mapping */
+/**@{*/
+/** \brief 3DLUT Channel layout is R, G, B, the default layout. Map RGB to RGB. */
+#define VA_3DLUT_CHANNEL_RGB_RGB              0x00000000
+/** \brief 3DLUT Channel layout is Y, U, V. Map YUV to RGB. */
+#define VA_3DLUT_CHANNEL_YUV_RGB              0x00000001
+/** \brief 3DLUT Channel layout is V, U, Y. Map VUY to RGB. */
+#define VA_3DLUT_CHANNEL_VUY_RGB              0x00000002
```
xhaihao, line 1345: VAProcPipelineParameterBuffer includes the input and output surface color standard, and we may know the layout / channel via surface fourcc, why do we need this mapping?

FurongZhang: Thanks Haihao for the feedback! 
Yes, VAProcPipelineParameterBuffer includes the input and output surface color standard.
But here, the enum definition has the different purpose. 3DLUT is a filter, this filter needs to know the input is YUV or RGB, OR VUY, this could be different from the VPP input. 
For example ,The VPP input is P010->CSC convert to RGB -> 3DLUT(RGB in, RGB out)->VPP output.

xhaihao: Can this case be done by CSC ? According to the comment below, 3DLUT is one of CSC. 

"3DLUT is usually used to transform input RGB/YUV values in one color space to output RGB values in another
   color space" 
 from RGB to RGB ?

FurongZhang: It can not be done by CSC. 3DLUT is more than CSC. It can do tone mapping, but CSC can not do tone mapping.

XinfengZhang (approved):

va/va_vpp.h
xhaihao: We already have the filter for tone mapping, is it possible to use tone mapping filter for your case? BTW I didn't see meta data related info in your new filter, did I miss something in your patch?

PR #493, opened 2021-01-13
Modify AV1 frame_width_minus1 and frame_height_minus1 comment

sushma08:
AV1 does not have any frame resolution restrictions. However, the comment
"value may not be multiple of 8" was confusing so, I am excluding it to be
in sync with the AV1 spec.


andrescj-chromium (approved):
Looks good to me :)

XinfengZhang (approved):

PR #498, opened 2021-02-02
trace: add image information trace for vaDeriveImage

XinfengZhang:

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -1776,6 +1776,9 @@ VAStatus vaCreateImage(
     ctx = CTX(dpy);
 
     va_status = ctx->vtable->vaCreateImage(ctx, format, width, height, image);
+    if (VA_STATUS_SUCCESS == va_status) {
+        VA_TRACE_LOG(va_TraceImageInfo, dpy, image);
+    }
     VA_TRACE_RET(dpy, va_status);
     return va_status;
 }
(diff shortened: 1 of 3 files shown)


va/va.c
```diff
@@ -1915,6 +1915,10 @@ VAStatus vaDeriveImage (
   ctx = CTX(dpy);
 
   va_status = ctx->vtable->vaDeriveImage ( ctx, surface, image );
+  if(VA_STATUS_SUCCESS == va_status)
+  {
+     VA_TRACE_LOG(va_TraceImageInfo, dpy, image);
```
dvrogozh, line 1920: add the same to `vaCreateImage`?

XinfengZhang: done

va/va_trace.c
```diff
@@ -5867,3 +5867,26 @@ void va_TraceStatus(VADisplay dpy, const char * funcName, VAStatus status)
...
 }
+
+void va_TraceImageInfo(VADisplay dpy, VAImage *image)
+{
+    int i;
+
+    DPY2TRACE_VIRCTX(dpy);
+
+    TRACE_FUNCNAME(idx);
+
```
dvrogozh, line 5878: We probably need to check that image is not null. Won't be good to crash during tracing...

PR #503, opened 2021-03-02
add device information query/set interface

XinfengZhang:
this interface is used to detect device information
device information may include:
memory locations, memory regions, deviceIDs etc.
it is common attributes between vaContext.

query interface:
this function should be called after vaInitialize

set interface:
this function should be called after vaInitialize
and before any other VAAPI functions, such as vaCreateSurfaces, etc.


va/va.h
```diff
@@ -4694,6 +4694,108 @@ VAStatus vaDeriveImage (
...
+        /** \brief memory region count, read - only */
+        uint32_t local_memory_regions      : 4;
+        /** \brief current region index read - write
+         * vaQueryDeviceAttibutes , it will return pre-assigned memory region.
+         * vaSetDeviceAttributes,  it is used indicate which memory region is selected*/
+        uint32_t current_memory_region     : 4;
+        /** \brief reserved bits for future, must be zero*/
+        uint32_t reserved  :8;
+        /** \brief bit mask to indicate which memory_region is available, read only
+         * VADeviceAttribValMemoryRegion reg;
```
dvrogozh, line 4750: wrap code w/ `\code ... \endcode`, please

va/va.h
```diff
@@ -4694,6 +4694,108 @@ VAStatus vaDeriveImage (
...
+         */
+        uint32_t memroy_region_masks : 16;
+    }bits;
+    uint32_t value;
+}VADeviceAttribValMemoryRegion;
+
+/**
+ * Query all attributes for a display handle
+ * The caller must provide an "attrib_list" array, the num_attibs should be set.
+ * if the The actual number of attributes < *num_attibs, the actual number of
```
dvrogozh, line 4773: fix text: "if the The". Should be: "If the". Mind capital I.

va/va.h
```diff
@@ -4694,6 +4694,108 @@ VAStatus vaDeriveImage (
...
+ * attributes will be returned in *num_attibs. if the *num_attibs < actual number
```
dvrogozh, line 4774: Capital I in if

va/va.h
```diff
@@ -4694,6 +4694,108 @@ VAStatus vaDeriveImage (
...
+ * of attributes, only part of the attibutes is set in attib_list. other should be treated as
+ * unsupported.
+ */
+
+VAStatus vaQueryDeviceAttributes(
```
dvrogozh, line 4779: Should this be named Query or Get?

va/va.h
```diff
@@ -4694,6 +4694,108 @@ VAStatus vaDeriveImage (
...
+     VADisplay dpy,
+     VADeviceAttrib *attrib_list,
+     uint32_t       *num_attribs
+);
+
+/**
+ * Set Device configuration by this function, only when the device attribues is writable
+ * if APP want to use this function to set device configuration,  this function must be
```
dvrogozh, line 4787: `if APP want` - is that a new sentence? add . and capitalize I.

va/va.h
```diff
@@ -4694,6 +4694,108 @@ VAStatus vaDeriveImage (
...
+ * called after vaInitialize and before any other VAAPI function. if this function is called
```
dvrogozh, line 4788: s/if/If/

dvrogozh (changes requested):

dvrogozh (commented):
I think we need to finish discussion in #505 before merging this one since similar terms like `local_memory`, regions and indices are discussed here as well.

@fhvwy : fyi, you might wish to look into this PR as well

va/va.c
```diff
@@ -2432,3 +2432,44 @@ VAStatus vaProtectedSessionExecute(
...
+  {
+    vaStatus = VA_STATUS_ERROR_UNIMPLEMENTED;
+  }
+
+  return vaStatus;
+}
+
+VAStatus vaSetDeviceAttributes(
+     VADisplay dpy,
+     VADeviceAttrib *attrib_list,
```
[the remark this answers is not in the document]
dvrogozh: Initial concept for this API was slightly different, in particular consideration was to have attribs settable before vaInitialize. In this sense old API did not work for us... But then concept has changed we forgot to evaluate the new API vs. existing one.

@XinfengZhang : indeed, 2 sets of APIs start to look weird. Should we converge and just declare couple new display attributes?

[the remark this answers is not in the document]
XinfengZhang: I wonder whether display attributes is suitable , because current display attributes seems just related with "Display" not adaptor

va/va.h
```diff
@@ -4694,6 +4694,110 @@ VAStatus vaDeriveImage (
...
+}VADeviceAttribValMemoryInfo;
+
+typedef union _VADeviceAttribValMemoryRegion{
+    struct{
+        /** \brief memory region count, read - only */
+        uint32_t local_memory_regions      : 4;
+        /** \brief current region index read - write
+         * vaQueryDeviceAttibutes , it will return pre-assigned memory region.
+         * vaSetDeviceAttributes,  it is used indicate which memory region is selected*/
+        uint32_t current_memory_region     : 4;
```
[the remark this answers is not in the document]
XinfengZhang: yes

va/va.c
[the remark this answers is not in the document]
dvrogozh: I think yes, considering that adaptor/display is kind of merged thing in libva, and I would say even on Linux device exposure level (since we can use /dev/dri/cardN for display and encoding). So, having all attributes together seems ok to me - they just configure different aspects of the same entity.

PR #504, opened 2021-03-02
add new PRIME type for more memory information

XinfengZhang:
the additional memory information include memory location and some addtional information

### DIFF
--- a/va/va_drmcommon.h
+++ b/va/va_drmcommon.h
@@ -87,6 +87,11 @@ struct drm_state {
  */
 #define VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2		0x40000000
 
+/** \brief DRM PRIME memory type (enhanced version)
+ *
+ * Used with VADRMPRIMESurfaceDescriptorExt.
+ */
+#define VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_3		0x80000000
 /**
  * \brief External buffer descriptor for a DRM PRIME surface.
  *
@@ -154,5 +159,24 @@ typedef struct _VADRMPRIMESurfaceDescriptor {
     } layers[4];
 } VADRMPRIMESurfaceDescriptor;
 
+
+typedef struct _VADRMPRIMESurfaceDescriptorExt {
+    VADRMPRIMESurfaceDescriptor surface_desc;
+    /* memory location , local memory or system memory */
+    union{
+        struct{
+            /* whether the memory is allocated on local memory */
+            uint32_t local_memory :1;
+            /* whether the memory region is enabled */
+            uint32_t memory_region_enable: 1;
+            /* memory region index */
+            uint32_t memory_region_index: 8;
+            /* reserved bits for future, must be zero */
+            uint32_t reserved: 22;
+        }bits;
+        uint32_t value;
+    } memory_location;
+    uint32_t                    va_reserved[5];
+}VADRMPRIMESurfaceDescriptorExt;
 
 #endif /* VA_DRM_COMMON_H */


fhvwy:
The intent of this is to connect to the memory_region stuff in the i915 kernel DRM driver, right?  Given that it won't apply to anything other than i915, maybe that should be in the name rather than calling it `DRM_PRIME_3` as if it applied to other DRM PRIME cases - something like `VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_I915_MEMORY_REGION`, perhaps?

Relatedly, what lies on the other side of this API?  Is there a matching OpenCL/Vulkan extension or KMS feature which accepts/provides this information?

va/va_drmcommon.h
```diff
@@ -155,4 +160,23 @@ typedef struct _VADRMPRIMESurfaceDescriptor {
 } VADRMPRIMESurfaceDescriptor;
 
 
+typedef struct _VADRMPRIMESurfaceDescriptorExt {
+    VADRMPRIMESurfaceDescriptor surface_desc;
+    /* memory location , local memory or system memory */
+    union{
+        struct{
+            /* whether the memory is allocated on local memory */
+            uint32_t local_memory :1;
```
fhvwy, line 169: What does "local memory" mean?  Is it memory local to the CPU running the user code, memory local to the device being used, something else?

va/va_drmcommon.h
```diff
@@ -155,4 +160,23 @@ typedef struct _VADRMPRIMESurfaceDescriptor {
...
+            /* whether the memory region is enabled */
+            uint32_t memory_region_enable: 1;
```
fhvwy, line 171: What does it mean for a memory region to be enabled?  I can't find any concept like this in the i915 driver.

va/va_drmcommon.h
```diff
@@ -155,4 +160,23 @@ typedef struct _VADRMPRIMESurfaceDescriptor {
...
+            /* memory region index */
+            uint32_t memory_region_index: 8;
```
fhvwy, line 173: Index of what?  Is this actually the ID of the memory region, as in `enum intel_region_id`?

XinfengZhang:
> The intent of this is to connect to the memory_region stuff in the i915 kernel DRM driver, right? Given that it won't apply to anything other than i915, maybe that should be in the name rather than calling it `DRM_PRIME_3` as if it applied to other DRM PRIME cases - something like `VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_I915_MEMORY_REGION`, perhaps?
> 
yes, one of the reason is memory_region, but the main purpose is that OCL want us provide the memory location, want to know the surface is allocated on system memory or local memory( on chip/device memory).

new memory type always indicate new structure, right? even we use VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_I915_MEMORY_REGION, we still need a structure, but the structure also need to contain the information already in VADRMPRIMESurfaceDescriptor. actually, I suppose we forgot to reserve some bytes in VADRMPRIMESurfaceDescriptor. 


> Relatedly, what lies on the other side of this API? Is there a matching OpenCL/Vulkan extension or KMS feature which accepts/provides this information?

Kernel user api does not provide a interface to query such information though a handle of buffer object. it's why I want to extend VADRMPRIMESurfaceDescriptor.  about "Is there a matching OpenCL/Vulkan extension or KMS feature which accepts/provides this information?" I am not sure, need to check.

va/va_drmcommon.h
XinfengZhang: memory local to the device being used. it is on chip/device memory.

va/va_drmcommon.h
XinfengZhang: if there are multiple memory chip, and has different control. it should be true.

va/va_drmcommon.h
XinfengZhang: the ID of the memory region, sorry , seems it is confusing, because there are already intel_region_id defined. so. maybe I need use other name to distinguish with enum intel_region_id

dvrogozh: If there is an index, then there should be a range for this index which does not seem present in this PR. This is confusing and needs a correction: please, specify how to get a range?

va/va_drmcommon.h
dvrogozh: I am not sure that this flag is actually needed. If there is `memory_region_index` specified, does not that mean that "memory region is enabled?"

dvrogozh (changes requested):

PR #506, opened 2021-03-04
Fix VASurfaceAttribPixelFormat docs

emersion:
The docs seem to refer to the legacy getSurfaceAttributes behavior.
Update them for the new querySurfaceAttributes behavior which
completely ignores what the user provides in the attributes array.
While at it, also document its usage with createSurfaces.


emersion:
Ping @XinfengZhang

XinfengZhang (approved):

PR #508, opened 2021-03-11
AV1: change doc reference from ref_frame_id[] to ref_frame_idx[].

andrescj-chromium:
Additionally, clarify some ambiguities in the documentation.

### DIFF
--- a/va/va_dec_av1.h
+++ b/va/va_dec_av1.h
@@ -339,12 +339,17 @@ typedef struct  _VADecPictureParameterBufferAV1
 
     /** \brief Surface indices of reference frames in DPB.
      *
-     *  Contains a list of uncompressed frame buffer surface indices as references.
-     *  Application needs to make sure all the entries point to valid frames
-     *  except for intra frames by checking ref_frame_id[]. If missing frame
-     *  is identified, application may choose to perform error recovery by
-     *  pointing problematic index to an alternative frame buffer.
-     *  Driver is not responsible to validate reference frames' id.
+     *  Contains a list of uncompressed frame buffer surface indices as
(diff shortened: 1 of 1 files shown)


va/va_dec_av1.h
```diff
@@ -339,12 +339,17 @@ typedef struct  _VADecPictureParameterBufferAV1
...
     /** \brief Surface indices of reference frames in DPB.
      *
-     *  Contains a list of uncompressed frame buffer surface indices as references.
-     *  Application needs to make sure all the entries point to valid frames
-     *  except for intra frames by checking ref_frame_id[]. If missing frame
-     *  is identified, application may choose to perform error recovery by
-     *  pointing problematic index to an alternative frame buffer.
-     *  Driver is not responsible to validate reference frames' id.
+     *  Contains a list of uncompressed frame buffer surface indices as
+     *  references for the current frame. If the current frame is not an intra
```
ChunjuanLiu, line 343: “for the current frame” is not accurate, these surfaces are not only used for the current frame, but also for future frames. Maybe some surface is not used for current frame, but used for future frame.

va/va_dec_av1.h
```diff
@@ -339,12 +339,17 @@ typedef struct  _VADecPictureParameterBufferAV1
...
+     *  frame (i.e., it's neither a key frame nor an intra-only frame), the
+     *  application needs to make sure all the entries in ref_frame_idx[] point
+     *  to valid frames in ref_frame_map[]. If a missing frame is identified,
+     *  the application may choose to perform error recovery by pointing the
```
ChunjuanLiu, line 347: Minor: remove "the" before "application" and "driver"?

PR #510, opened 2021-03-22
Add H.264 Scalable (SVC) profiles

thaithong:
Signed-off-by: Thong Thai <thong.thai@amd.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -507,7 +507,9 @@ typedef enum
     VAProfileAV1Profile1                = 33,
     VAProfileHEVCSccMain444_10          = 34,
     /** \brief Profile ID used for protected video playback. */
-    VAProfileProtected                  = 35
+    VAProfileProtected                  = 35,
+    VAProfileH264ScalableBaseline       = 36,
+    VAProfileH264ScalableHigh           = 37
 } VAProfile;
 
 /**


XinfengZhang:
@thaithong , just these two profile? how to check the support of temporal ,spatial, quality ?

thaithong:
@XinfengZhang Just these two profiles for now. I'm planning to add structs for the "scalability_info" SEI message later, which will contain the additional information. I'm currently working on encoding/decoding temporal, SVC Baseline.

cwang105:
what is the purpose of adding SVC profiles? only to support temporal scalability?

thaithong:
Yes, for now, just to encode/decode H.264 temporal scalability videos.

XinfengZhang:
sorry for late response, from my understanding, if we are only want SVC-T
the implementation maybe be transparent to driver
application could set the temporal layers and manage reference relationship.

thaithong:
Hi @XinfengZhang , So right now I'm using the packed headers to pass information to the driver, but another reason I wanted to add the profile was have the profile_idc appear properly in the encoded video.

So that programs like `mediainfo` would show the proper profile, like this:

`Format profile                           : Scalable Baseline@L1.3`

...instead of it's currently showing:

`Format profile                           : Baseline@L1.3`

I made changes locally, with gstreamer-vaapi to add the SVC profile, but they're just hardcoded numbers instead of named enums, since libva doesn't have the profile yet.

The plan is to add the SVC profiles to libva, and then update gstreamer-{vaapi, h264parser} with the SVC profiles and structs (for NALU prefix, SEI scalability_info).

thaithong:
@XinfengZhang ping

XinfengZhang:
actually, I dont understand why mediainfo will use vaapi definition, from my understanding , it should parse the container or bitstream to get the information , then display the information (mapping between the profile_idc and string of profile). 
but now, seems there are another mapping (profile_idc to vaapi definition , vaapi definition to string of media info showing)?
could you also file your gst PR , and link it into this one?

thaithong:
I didn't mean it as mediainfo uses VAAPI to parse. I meant adding the profile to libva (as a first step) to having SVC videos encoded by VAAPI, use the profile in the bitstream headers, so that programs like mediainfo will show the correct profile when it parses the bitstream (that was encoded using VAAPI).

I didn't intend to merge my gst changes yet - the changes are still kind of messy. I'll try to clean up the code and post it soon then.

XinfengZhang:
hi @thaithong but just like you mentioned in https://github.com/intel/libva/pull/510#issuecomment-825644210, we are using packed header to insert the profile_id into bitstream. so, the profile_idc in packed header will be inserted into bitstream, then it will be parsed by mediainfo.    it's what I am confused.

my real concern is about the caps report. for example, for vaQueryConfigProfiles, I dont know whether backend driver should report the new defined VAProfileH264ScalableBaseline & VAProfileH264ScalableHigh. if backend driver report this , does it means backend driver will support it. but till now, as I know. almost all vendors depend on the sw (on top of libva) to implement SVC.  If backend does not report it. it also looks weird.  it is tricky.

PR #514, opened 2021-03-31
report the capability of memory support

XinfengZhang:
extend display attibute to include adaptor/hw/platform attribute
and add two new attributes to indicate whether there are on device memory
and whether there are different memory region on the device

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -4879,7 +4879,17 @@ VAStatus vaDeassociateSubpicture (
  * brightness etc. in the rendering process.  The application can query what
  * attributes are supported by the driver, and then set the appropriate attributes
  * before calling vaPutSurface()
+ *
+ * Display attributes can also be used to query/set platform or display adaptor (vaDisplay)
+ * related information. These attributes do not depend on vaConfig, and could not be used
+ * for vaPutSurface. Application can use vaQueryDisplayAttributes/vaGetDisplayAttributes
+ * at anytime after vaInitialize, but (for settable attributes) vaSetDisplayAttributes should be
+ * called after vaInitialize and before any other function call.
+ *
+ * To distinguish these two types of display attributes, display adaptor related attributes
+ * should be marked as "HW attribute" in the description.
  */
+
 /* PowerVR IEP Lite attributes */
 typedef enum
 {
@@ -4928,6 +4938,47 @@ typedef enum
 #define VA_RENDER_DEVICE_UNDEFINED  0
 #define VA_RENDER_DEVICE_LOCAL      1
 #define VA_RENDER_DEVICE_EXTERNAL   2
+
+/**\brief sub device info
+ * Sub-device is the concept basing on the "device" behind "vaDisplay".
+ * If a device could be divided to several sub devices, the task of
+ * decode/encode/vpp could be assigned on one sub-device. So, application
+ * could choose the sub device before any other operations. After that,
+ * all of the task execution/resource allocation will be dispatched to
+ * the sub device. If application does not choose the sub device, driver
+ * will assign one as default.
+ *
+ * If the value == VA_ATTRIB_NOT_SUPPORTED, it mean that the attribute
+ * is unsupport or UNKNOWN.
+ */
+
+typedef union _VADisplayAttribValSubDevice{
+    struct{
+        /** \brief current sub device index, read - write */
+        uint32_t current_sub_device     : 4;
+        /** \brief sub devices count, read - only */
+        uint32_t sub_device_count       : 4;
+        /** \brief reserved bits for future, must be zero*/
+        uint32_t reserved               : 8;
+        /** \brief bit mask to indicate which sub_device is available, read only
+         * \code
+         * VADisplayAttribValSubDevice reg;
+         * VADisplayAttribute reg_attr;
+         * reg_attr.type = VADisplayAttribSubDevice;
+         * vaGetDisplayAttributes(dpy, &reg_attr, 1);
+         * reg.value = reg_attr.value;
+         *
+         * for(int i = 0; i < reg.bits.sub_device_count; i ++ ){
+         *    if((1<<i) & reg.bits.sub_device_mask){
+         *        printf("sub device  %d can be selected", i);
+         *    }
+         *}
+         * \endcode
+         */
+        uint32_t sub_device_mask       : 16;
+    }bits;
+    uint32_t value;
+}VADisplayAttribValSubDevice;
 
 /** Currently defined display attribute types */
 typedef enum
@@ -5001,6 +5052,10 @@ typedef enum
      * specify vaPutSurface render area if there is no drawable on the monitor
      */
     VADisplayAttribRenderRect          = 18,
+    /*
+     * HW attribute, read/write, specify the sub device configure
+     */
+    VADisplayAttribSubDevice           = 19,
 } VADisplayAttribType;
 
 /* flags for VADisplayAttribute */


XinfengZhang:
it is another option for #503

va/va.h
```diff
@@ -4828,7 +4828,17 @@ VAStatus vaDeassociateSubpicture (
  * brightness etc. in the rendering process.  The application can query what
  * attributes are supported by the driver, and then set the appropriate attributes
  * before calling vaPutSurface()
+ *
+ * Display attributes also are used to query/set display adaptor (vaDisplay) related information
```
dvrogozh, line 4832: . in the end of the sentence

va/va.h
```diff
@@ -4828,7 +4828,17 @@ VAStatus vaDeassociateSubpicture (
...
+ * These attributes does not depend on vaConfig, only related with platform or display adaptor
```
dvrogozh, line 4833: typo: "do not".

'.' in the end of the sentence

va/va.h
```diff
@@ -4828,7 +4828,17 @@ VAStatus vaDeassociateSubpicture (
...
+ * It could not used for vaPutSurface.
+ * App could call vaQueryDisplayAttributes/vaGetDisplayAttributes to query these attributes
+ * at anytime after vaInitialize, but only could call vaSetDisplayAttributes for these attributes
+ * after vaInitialize and before any other function call if the attributes is settable.
+ * To distinguish these two types display attributes, this display adaptor related attributres
```
dvrogozh, line 4838: to types >>>of<<<

va/va.h
```diff
@@ -4828,7 +4828,17 @@ VAStatus vaDeassociateSubpicture (
  * brightness etc. in the rendering process.  The application can query what
  * attributes are supported by the driver, and then set the appropriate attributes
  * before calling vaPutSurface()
+ *
+ * Display attributes also are used to query/set display adaptor (vaDisplay) related information
+ * These attributes does not depend on vaConfig, only related with platform or display adaptor
+ * It could not used for vaPutSurface.
```
dvrogozh, line 4834: type: >>>They<< could not >>>be<<< used

dvrogozh (approved):

va/va.h
```diff
@@ -4835,7 +4835,17 @@ VAStatus vaDeassociateSubpicture (
  * brightness etc. in the rendering process.  The application can query what
  * attributes are supported by the driver, and then set the appropriate attributes
  * before calling vaPutSurface()
+ *
+ * Display attributes also are used to query/set display adaptor (vaDisplay) related information.
+ * These attributes do not depend on vaConfig, only related with platform or display adaptor
+ * They could not be used for vaPutSurface.
+ * App could call vaQueryDisplayAttributes/vaGetDisplayAttributes to query these attributes
+ * at anytime after vaInitialize, but only could call vaSetDisplayAttributes for these attributes
+ * after vaInitialize and before any other function call if the attributes is settable.
```
HeJunyan, line 4844: Can we modify the behavior to:

App could call vaQueryDisplayAttributes/vaGetDisplayAttributes to query these attributes at anytime after vaInitialize,
but only could call vaSetDisplayAttributes for these attributes BEFORE vaInitialize.

Which make the vaInitialize a time point  to bind the tile?

XinfengZhang: if we make the change just like your suggestion, it means that there are still no vaSetDisplayAttributes function pointer before initialize,  part of the logic should in libva instead of backend driver.

HeJunyan: OK, then I can accept this.

va/va.h
```diff
@@ -4885,6 +4895,44 @@ typedef enum
...
+        /** \brief bit mask to indicate which memory_region is available, read only
+         * \code
+         * VADisplayAttribValMemoryRegion reg;
+         * VADisplayAttribute reg_attr;
+         * reg_attr.type = VADisplayAttribMemoryRegion;
+         * vaGetDisplayAttributes(dpy, &reg_attr, 1);
+         * reg.value = reg_attr.value;
+         *
+         * for(int i = 0; i < reg.bits.local_memory_regions; i ++ ){
+         *    if((1<<i) & reg.bits.memory_region){
```
liangali, line 4925: shall we use "reg.bits.memory_region_masks" instead?

liangali (approved):

dvrogozh (approved):

va/va.h
```diff
@@ -4885,6 +4895,47 @@ typedef enum
...
+         * reg.value = reg_attr.value;
+         *
+         * for(int i = 0; i < reg.bits.local_memory_regions; i ++ ){
+         *    if((1<<i) & reg.bits.memory_region_masks){
+         *        printf("memory region %d can be selected", i);
+         *    }
+         *}
+         * \endcode
+         */
+        uint32_t memroy_region_masks : 16;
```
dvrogozh, line 4934: fix typos: s/memroy/memory/. And should we use plural in a different way: `memory_regions_mask`?

va/va.h
```diff
@@ -4835,7 +4835,17 @@ VAStatus vaDeassociateSubpicture (
...
  * before calling vaPutSurface()
+ *
+ * Display attributes can also be used to query/set platform or display adaptor (vaDisplay)
+ * related information. These attributes do not depend on vaConfig, and could not be used
+ * for vaPutSurface. Application can use vaQueryDisplayAttributes/vaGetDisplayAttributes
+ * at anytime after vaInitialize, but (for settable attributes) vaSetDisplayAttributes should be
+ * called after vaInitialize and before any other function call.
+ *
+ * To distinguish these two types of display attributes, display adaptor related attributes
+ * should be marked as "HW attribute" in the description.
```
xhaihao, line 4846: Does it mean platform related information is not 'HW attribute' ? 

How about to use VADisplayAdaptorAttrib instead of VADisplayAttrib as prefix?

va/va.h
```diff
@@ -4885,6 +4895,47 @@ typedef enum
...
+         *    if((1<<i) & reg.bits.memory_region_masks){
+         *        printf("memory region %d can be selected", i);
+         *    }
+         *}
+         * \endcode
+         */
+        uint32_t memory_regions_masks : 16;
+    }bits;
+    uint32_t value;
+}VADisplayAttribValMemoryRegion;
```
xhaihao, line 4937: Use VADisplayAdaptorAttrib as prefix ?

va/va.h
```diff
@@ -4885,6 +4895,47 @@ typedef enum
...
+         * reg.value = reg_attr.value;
+         *
+         * for(int i = 0; i < reg.bits.local_memory_regions; i ++ ){
+         *    if((1<<i) & reg.bits.memory_region_masks){
+         *        printf("memory region %d can be selected", i);
+         *    }
+         *}
+         * \endcode
+         */
+        uint32_t memory_regions_masks : 16;
```
xhaihao, line 4934: Is 16 enough for future use?

XinfengZhang: if not , we could consider use min_value and max_value.

va/va.h
XinfengZhang: it is good idea, do we need to change the VADisplayAttribSubDevice also, if we changed it, in the enum, the prefix looks not aligned.

va/va.h
```diff
@@ -4929,6 +4939,47 @@ typedef enum
 #define VA_RENDER_DEVICE_LOCAL      1
 #define VA_RENDER_DEVICE_EXTERNAL   2
 
+/**\brief sub device info
+ * Sub-device is the concept basing on the "device" behind "vaDisplay".
+ * If a device could be divided to several sub device, the task of
```
dvrogozh, line 4944: "to several sub devices", plural

va/va.h
```diff
@@ -4929,6 +4939,47 @@ typedef enum
...
+ * decode/encode/vpp could be assigned on one sub-device. So, application
+ * could choose the sub device and attach it to vaDisplay before
+ * any other operations. after that, all of the task execution/resource
```
dvrogozh, line 4947: "After", capital letter

va/va.h
```diff
@@ -4929,6 +4939,47 @@ typedef enum
 #define VA_RENDER_DEVICE_LOCAL      1
 #define VA_RENDER_DEVICE_EXTERNAL   2
 
+/**\brief sub device info
+ * Sub-device is the concept basing on the "device" behind "vaDisplay".
+ * If a device could be divided to several sub device, the task of
+ * decode/encode/vpp could be assigned on one sub-device. So, application
+ * could choose the sub device and attach it to vaDisplay before
```
dvrogozh, line 4946: I don't quite like "attach". It sounds like in the future we might be able to attach multiple sub-devices to vaDisplay. Is that we want to have? This will require API change however since since right now we have only `current_sub_device`, but we will need a mask.

va/va.h
```diff
@@ -4929,6 +4939,47 @@ typedef enum
...
+         * reg.value = reg_attr.value;
+         *
+         * for(int i = 0; i < reg.bits.local_sub_devices; i ++ ){
+         *    if((1<<i) & reg.bits.sub_device_masks){
+         *        printf("sub device  %d can be selected", i);
+         *    }
+         *}
+         * \endcode
+         */
+        uint32_t sub_devices_masks      : 16;
```
dvrogozh, line 4978: too many plurals. I suggest `sub_devices_mask`

dvrogozh (changes requested):

dvrogozh (approved):

PR #513, opened 2021-03-31
report the capability of vaCopy support

XinfengZhang:
Signed-off-by: Carl Zhang <carl.zhang@intel.com>

PR #519, opened 2021-04-08
Introduce VADisplayAttribSurfaceAttribs

emersion:
This allows applications to query the supported surface attributes.

This is useful to allow applications to make sure their attributes
are taken into account, and allow them to have a fallback when a
surface attribute is not supported.

ffmpeg already needs [a workaround][1] to figure out whether a driver
supports VASurfaceAttribMemoryType.

The concept of a "HW attribute" is borrowed from https://github.com/intel/libva/pull/513 and https://github.com/intel/libva/pull/514.

[1]: https://github.com/FFmpeg/FFmpeg/blob/d393c45051ddaf6146e7e29ec2ea97035a727529/libavutil/hwcontext_vaapi.h#L53

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -4827,7 +4827,16 @@ VAStatus vaDeassociateSubpicture (
  * Display attributes are used to control things such as contrast, hue, saturation,
  * brightness etc. in the rendering process.  The application can query what
  * attributes are supported by the driver, and then set the appropriate attributes
- * before calling vaPutSurface()
+ * before calling vaPutSurface().
+ *
+ * Display attributes also are used to query/set display adaptor (vaDisplay)
+ * related information. These attributes do not depend on vaConfig. They can't
+ * be used with vaPutSurface. Apps can call vaQueryDisplayAttributes/
+ * vaGetDisplayAttributes to query these attributes after vaInitialize, but can
+ * only call vaSetDisplayAttributes for these attributes after vaInitialize and
+ * before any other function call if the attributes is settable. To distinguish
+ * these two types of display attributes, this display adaptor related
+ * attributes should be marked as "HW attribute" in the description.
  */
 /* PowerVR IEP Lite attributes */
 typedef enum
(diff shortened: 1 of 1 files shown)


XinfengZhang:
not sure whether I understand your purpose, I dont think we need a special attribute to report whether we support another attributes. when a attributes is un-support. it should return with a value VA_ATTRIB_NOT_SUPPORTED, if it is a pointer, it should be set to null.  so, why we need to use vaQueryDisplayAttributes to query whether a surface attribute is support or not?

emersion:
> when a attributes is un-support. it should return with a value VA_ATTRIB_NOT_SUPPORTED

Ah, I was wondering about this. It doesn't seem like intel media driver is doing it. I've filled a pull request to fix that: https://github.com/intel/media-driver/pull/1176

> so, why we need to use vaQueryDisplayAttributes to query whether a surface attribute is support or not?

It's a little tricky for the caller to figure out which attribute caused the failure. If the application provided both `VASurfaceAttribUsageHint` and `VASurfaceAttribMemoryType`, which one caused the failure?

If they know in advanced the surface attributes supported by the display, applications can have a fallback. For instance, if `VASurfaceAttribUsageHint` is not supported, applications can remove it, since it's just a hint.

XinfengZhang:
the normal call sequence should be:
vaQuerySurfaceAttributes/vaGetSurfaceAttributes  then call vaCreateSurfaces

so , before vaCreateSurfaces, it should already know which attribute could be supported.

with your PR, the call sequence seems be changed to 
vaQueryDisplayAttributes/vaGetDisplayAttributes ===> vaCreateSurfaces

it looks a little strange.

emersion:
Ah, I guess my question is: how does the application know that the driver supports attributes which aren't readable (only writable)? Examples: `VASurfaceAttribExternalBufferDescriptor` and `VASurfaceAttribDRMFormatModifiers`.

Should the driver return the writable attribute without any value in `vaQuerySurfaceAttributes`?

XinfengZhang:
query function support this flag https://github.com/intel/libva/blob/master/va/va.h#L1600 and the value could be
VA_SURFACE_ATTRIB_GETTABLE  or VA_SURFACE_ATTRIB_SETTABLE

emersion:
Closing since drivers should return settable attributes in `vaQuerySurfaceAttributes`. I checked intel-media-driver and Mesa, both already do so for `VASurfaceAttribExternalBufferDescriptor`.

PR #524, opened 2021-05-05
Add VA_RT_FORMAT_BGRP to support Planar BGR format.

FurongZhang:

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -989,6 +989,7 @@ typedef struct _VAConfigAttrib {
 #define VA_RT_FORMAT_RGB32	0x00020000	///< Packed RGB, 32 bits per pixel, 8 bits per colour sample.
 #define VA_RT_FORMAT_RGBP	0x00100000	///< Planar RGB, 8 bits per sample.
 #define VA_RT_FORMAT_RGB32_10	0x00200000	///< Packed RGB, 32 bits per pixel, 10 bits per colour sample.
+#define VA_RT_FORMAT_BGRP	0x00400000	///< Planar RGB, 8 bits per sample.
 
 #define VA_RT_FORMAT_PROTECTED	0x80000000
 


xhaihao:
It would be better to submit another PR if you want to change the coding style.

FurongZhang:
@xhaihao , it makes sense to me. I have changed that. Thank you for reviewing it.

va/va.h
```diff
@@ -989,6 +989,7 @@ typedef struct _VAConfigAttrib {
 #define VA_RT_FORMAT_RGB32	0x00020000	///< Packed RGB, 32 bits per pixel, 8 bits per colour sample.
 #define VA_RT_FORMAT_RGBP	0x00100000	///< Planar RGB, 8 bits per sample.
 #define VA_RT_FORMAT_RGB32_10	0x00200000	///< Packed RGB, 32 bits per pixel, 10 bits per colour sample.
+#define VA_RT_FORMAT_BGRP	0x00400000	///< Planar RGB, 8 bits per sample.
```
XinfengZhang, line 992: BGRP means , first plane is Blue, second is green, final is red. revert order of RGBP.  and it is 8bit only?

xhaihao:
Do we really need a new RT format for RGB planar format ? I think the combination of  VA_RT_FORMAT_RGBP and VA_FOURCC_BGRP can be used for BGR planar format,  like what we did for VA_RT_FORMAT_RGB32 ( VA_RT_FORMAT_RGB32 can be used for different RGBA/RGBx formats in VA-API).

va/va.h
xhaihao: VA_FOURCC_BGRP is used to specify the layout.

FurongZhang: Yes

FurongZhang:
@xhaihao , yes
/**
 * \brief Creates an array of surfaces
 *
 * Creates an array of surfaces. The optional list of attributes shall
 * be constructed based on what the underlying hardware could expose
 * through vaQuerySurfaceAttributes().
 *
 * @param[in] dpy               the VA display
 * @param[in] format            _the desired surface format. See \c VA_RT_FORMAT_*_
 * @param[in] width             the surface width
 * @param[in] height            the surface height
 * @param[out] surfaces         the array of newly created surfaces
 * @param[in] num_surfaces      the number of surfaces to create
 * @param[in] attrib_list       the list of (optional) attributes, or \c NULL
 * @param[in] num_attribs       the number of attributes supplied in
 *     \c attrib_list, or zero
 */
VAStatus
vaCreateSurfaces(
    VADisplay           dpy,
    unsigned int        format,
    unsigned int        width,
    unsigned int        height,
    VASurfaceID        *surfaces,
    unsigned int        num_surfaces,
    VASurfaceAttrib    *attrib_list,
    unsigned int        num_attribs
);

XinfengZhang:
@xhaihao , it is RT format, not fourcc, as previous discussion, RT format could include several Fourcc, so adding this RT format is only for attributes reporting?

FurongZhang:
VA_RT_FORMAT_RGB32 can be used for different RGBA/RGBx formats in VA-API. Has the same memory layout.

FurongZhang:
@xhaihao @XinfengZhang , as we discussed, use RGBP RT format. Closed this PR.

PR #525, opened 2021-05-05
Add CHANGELOG.md to track the changes.

FurongZhang:


XinfengZhang (changes requested):
TBH, it is not a good idea, it bring additional effort to each commit from different contributors.

Issue #536, opened 2021-08-12
crash in libva when using iHD with mpv and vo=vaapi on Xwayland


LaserEyess:
Most information is copy and pasted from: https://github.com/intel/media-driver/issues/1095.

Essentially, the following command can reproduce a null pointer dereference inside libva when running under wayland
```
LIBVA_DRIVER_NAME=iHD mpv --no-config --vo=vaapi $file
```
`vo=vaapi` is Xorg only, so when you run this command under wayland it uses Xwayland. Quoted from the other issue.

> I  have rebuilt a libva (https://github.com/intel/libva/commit/6e86b4fb4dafa123b1e31821f61da88f10cfbe91) and this repository (https://github.com/intel/media-driver/commit/f315c36020eed2d2b52b9557cb888ca535f17886) and got mpv to work with them via `LD_PRELOAD` and `LIBVA_DRIVERS_PATH`. Both of them have full debug symbols and I can set breakpoints if needed to step through the code.
 >
> Here is a current backtrace: https://0x0.st/-J1l.txt
>
> Jexu's PR fixed an assert the line right before the crash, it calls into [`dri_vtable->get_drawable()`](https://github.com/intel/media-driver/blob/master/media_driver/linux/common/ddi/media_libva_putsurface_linux.cpp#L392) here and crashes.
>
> It looks like what happens is that inside that function is that it immediately calls into a funciton called `do_drawable_hash` in libva. In there it hits [this line](https://github.com/intel/libva/blob/master/va/x11/va_dricommon.c#L74), and the problem there is
> ```
> (gdb) info locals
>
> dri_state = 0x7fab78023100
> index = 2
> dri_drawable = 0x0
> (gdb) p dri_state->createDrawable
> $1 = (struct dri_drawable *(*)(VADriverContextP, XID)) 0x0
> ```
> So that's the issue I guess, a NULL pointer as this function? I did not look any further than this, but it looks like `dri_state` is [created here](https://github.com/intel/libva/blob/master/va/x11/va_dricommon.c#L64).
>
Let me know if I need to provide anything else. This is trivial to reproduce.

Jexu:
@LaserEyess Can you help verify and review the fix?

LaserEyess:
No this did not work. If you look at my previous backtrace `dri_drawable` was also NULL, so you have only delayed the crash by exactly one line. Now it crashes on `dri_drawable->x_drawable = drawable;`.

LaserEyess:
[...]

PR #537, opened 2021-08-12
Check the function pointer before using

Jexu:
Fixes: #536

Signed-off-by: Xu, Zhengguo <zhengguo.xu@intel.com>

### DIFF
--- a/va/x11/va_dricommon.c
+++ b/va/x11/va_dricommon.c
@@ -71,11 +71,17 @@ do_drawable_hash(VADriverContextP ctx, XID drawable)
         dri_drawable = dri_drawable->next;
     }
 
-    dri_drawable = dri_state->createDrawable(ctx, drawable);
-    dri_drawable->x_drawable = drawable;
-    dri_drawable->is_window = is_window(ctx->native_dpy, drawable);
-    dri_drawable->next = dri_state->drawable_hash[index];
-    dri_state->drawable_hash[index] = dri_drawable;
+    if(dri_state->createDrawable)
+    {
+        dri_drawable = dri_state->createDrawable(ctx, drawable);
+    }
+    if(dri_drawable)
+    {
+        dri_drawable->x_drawable = drawable;
+        dri_drawable->is_window = is_window(ctx->native_dpy, drawable);
+        dri_drawable->next = dri_state->drawable_hash[index];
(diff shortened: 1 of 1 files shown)


va/x11/va_dricommon.c
```diff
@@ -71,7 +71,10 @@ do_drawable_hash(VADriverContextP ctx, XID drawable)
         dri_drawable = dri_drawable->next;
     }
 
-    dri_drawable = dri_state->createDrawable(ctx, drawable);
+    if(dri_state->createDrawable)
+    {
+        dri_drawable = dri_state->createDrawable(ctx, drawable);
+    }
     dri_drawable->x_drawable = drawable;
```
LaserEyess, line 78: There is no guarantee that `dri_drawable` isn't NULL here, that's what `dri_state->createDrawable()` is for I guess. So you could add another conditional to check if `dri_drawable` is null, but then you'd get to `dri_state->drawable_hash[index] = dri_drawable;`, and then you'd return a NULL dri_drawable, etc.

So I'm wondering if the problem isn't much, much deeper than a superficial check here. I'm not familiar at all with libva or how vaapi drivers work, but it seems to me that there is some missing null check in `ctx` somewhere. Some assumption about Xorg is violated when using Xwayland, and this specifically happens on iHD only, so maybe the fix really belongs there.

XinfengZhang (approved):

va/x11/va_dricommon.c
Jexu: driver checks nothing but call va func 'va_dri_get_drawable' to get the drawable. And func pointer 'createDrawable' is inited in va_isDRI2Connected. since iHD uses dri3 instead of dri2, then this pointer is not inited. so we saw the null pointer when call 'createDrawable'.

LaserEyess: Ok, that sounds like a more involved fix which requires something like https://github.com/intel/libva/pull/180 to work. I guess you're not looking to fix this by implementing that. If that's the case, why not just do
```C
if !(dri_state->createDrawable)
{
    return dri_drawable
}
```

That gives the same result as what you have right now? Basically you're just accepting that `dri_drawable` will return NULL.

XinfengZhang (approved):

PR #541, opened 2021-09-28
meson: Copy public headers to build directory

ceyusa:
In order to use libva in a [subproject](https://mesonbuild.com/Subprojects.html) or in an "uninstalled" environment, libva needs to fix its public headers layout in its source directory. A [previous fix was proposed](intel#433) which moved those headers as expected, but it was closed because it was too
invasive.

This new approach is less invasive and oriented to meson subprojects or uninstalled setups, which has a wider audience.

The fix is two-fold:

1. Bump the required version of meson in order to support this new feature. And update deprecated syntax.
2. Copy each public header file into subproject's build directory.

### DIFF
--- a/va/meson.build
+++ b/va/meson.build
@@ -46,6 +46,9 @@ libva_headers = [
   version_file,
 ]
 
+# headers that have be copied for subprojects
+libva_headers_subproject = []
+
 libva_headers_priv = [
   'sysdeps.h',
   'va_fool.h',
@@ -87,6 +90,8 @@ if WITH_DRM
   libva_drm_headers = [
     'drm/va_drm.h',
   ]
+
+  libva_headers_subproject += libva_drm_headers
 
   libva_drm_headers_priv = [
     'drm/va_drm_auth.h',
@@ -134,11 +139,14 @@ if WITH_X11
     'x11/va_x11.c',
   ]
 
-  libva_x11_headers = [
-    'va_x11.h',
+  libva_dri_headers = [
     'x11/va_dri2.h',
     'x11/va_dricommon.h',
   ]
+
+  libva_x11_headers = ['va_x11.h'] + libva_dri_headers
+
+  libva_headers_subproject += libva_dri_headers
 
   libva_x11_headers_priv = [
     'x11/va_dri2str.h',
@@ -178,6 +186,8 @@ if WITH_GLX
     'glx/va_glx.h',
   ]
 
+  libva_headers_subproject += libva_glx_headers
+
   libva_glx_headers_priv = [
     'glx/va_glx_impl.h',
     'glx/va_glx_private.h',
@@ -215,6 +225,8 @@ if WITH_WAYLAND
     'wayland/va_backend_wayland.h',
     'wayland/va_wayland.h',
   ]
+
+  libva_headers_subproject += libva_wayland_headers
 
   libva_wayland_headers_priv = [
     'wayland/va_wayland_drm.h',
@@ -256,3 +268,8 @@ if WITH_WAYLAND
     include_directories : configinc,
     dependencies : deps)
 endif
+
+fs = import('fs')
+foreach header : libva_headers_subproject
+  configure_file(output: fs.name(header), input: header, copy: true)
+endforeach
--- a/meson.build
+++ b/meson.build
@@ -8,7 +8,7 @@
 project(
   'libva', 'c',
   version : '2.14.0.1',
-  meson_version : '>= 0.37.0',
+  meson_version : '>= 0.53.0',
   default_options : [ 'warning_level=1',
                       'buildtype=debugoptimized' ])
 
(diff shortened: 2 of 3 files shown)


va/meson.build
```diff
@@ -256,3 +268,16 @@ if WITH_WAYLAND
...
+  python3 = import('python').find_installation()
+  copy_header = '''
+import os, shutil, sys
+outdir = os.path.join(os.environ["MESON_BUILD_ROOT"], os.environ["MESON_SUBDIR"])
+shutil.copy(sys.argv[1], outdir)
+'''
+
+  foreach header : libva_headers_subproject
+    run_command(python3, '-c', copy_header, header, check: true)
+  endforeach
```
tp-m, line 282: Can't you just do something like this (with more path/subdir shenanigans if needed)?
```meson
foreach header : libva_headers_subproject
  configure_file(output: header, input: header, copy: true)
endforeach
```

ceyusa: done! thanks!

eli-schwartz:
Hmm, it seems like the cross-referenced PR was not at all gstreamer oriented, but was instead the correct (targeted) fix.

The fundamental problem which both PRs try to solve is:

the include path for installed (public API) headers is different between the source tree and the install layout. So, there is nowhere in the source tree that you can add as an include path to compile against the official header API (`va/*.h`).

This happens to be exposed by any use of meson subprojects, of which gst-build is one example.

Another thing that is affected, is meson's builtin support for pkg-config `*-uninstalled.pc` files -- you can add `PKG_CONFIG_PATH=<builddir>/meson-uninstalled` as an environment variable and then other projects doing `pkg-config --cflags --libs libva` in their configure scripts or looking up `dependency('libva')` or what have you, will detect the uninstalled project because pkg-config inherently prefers -uninstalled.pc as a development environment.
That being said, libva does not use meson's `import('pkgconfig').generate()` API, but rather uses handwritten pkg-config files. So, no -uninstalled.pc is created anyway.

...

One PR tried to solve this by natively making the source tree paths match an expected layout.

This PR instead tries to solve this by making additional copies of each file, and the additional copies match the expected layout.

ceyusa:
@eli-schwartz thanks for noticing the pkgconfig issue. I'm fixing it. I confirm that compiling media-driver is easier with `*-uninstalled.pc` files.

ceyusa:
The problem, of course, is that these new `*-uninstalled.pc` files only works within a meson subproject environment, since the headers layout is not fixed natively.

eli-schwartz:
No, you just need to fix this kwarg:

> `subdirs` which subdirs of `include` should be added to the header search path, for example if you install headers into `${PREFIX}/include/foobar-1`, the correct value for this argument would be `foobar-1`

ceyusa:
@eli-schwartz but that subdir kwarg would be only valid for uninstalled, right? I don't see a way to specify that.

eli-schwartz:
Wait, sorry, I don't think that's needed. Right. The current include path is includedir.

... So why doesn't this work when installed? Why only in a subproject?

eli-schwartz:
> The problem, of course, is that these new *-uninstalled.pc files only works within a meson subproject environment, since the headers layout is not fixed natively.

What's the problem here?

ceyusa:
> What's the problem here?

iiuc, using uninstalled pc files outside of a subproject setup won't work, since the missing headers are not copied.

eli-schwartz:
```
if meson.is_subproject()
```

Right, that would do it. You can simply remove that if and always copy them.

ceyusa:
Now, with this script I can build media-driver inside a gst-build/gst-env 

https://gist.github.com/ceyusa/c3f7e4be11a80a19c459db5c88f40074

XinfengZhang:
not familiar with meson build. I just tried your patch , and suppose no difference with "installed" headers, , a new folder generated in builddir---"meson-uninstalled", also in builddir/va, new files added (va_backend_glx.h, va_backend_wayland.h, va_dri2.h,va_dricommon.h,va_drm.h,va_glx.h, va_wayland.h).

ceyusa:
Thanks @XinfengZhang.

As you mentioned this patch doesn't change any while used as an isolated project (build and install). But it enables libva for [meson's subprojects](https://mesonbuild.com/Subprojects.html), which is something that GStreamer uses for developers.

The problem (as mentioned in [old bug](https://github.com/intel/libva/pull/433)) is that these headers -va_backend_glx.h, va_backend_wayland.h, va_dri2.h,va_dricommon.h,va_drm.h,va_glx.h, va_wayland.h- aren't reachable to the compiler if they are not installed, which is the way how meson subprojects work (uninstalled headers and pkgconfig files). This patch makes them reachable to the compiler by copying them to `builddir/va`.

XinfengZhang:
thanks, I have no other concern. @xhaihao @dvrogozh any concern from your side?

XinfengZhang (approved):

XinfengZhang:
@ceyusa , if possible, please help to rebase the PR

ceyusa:
> @ceyusa , if possible, please help to rebase the PR

done

ceyusa:
GStreamer patch  https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/1397

XinfengZhang (approved):

PR #547, opened 2021-10-31
VA/X11: add crocus mapping

jbeich:
Fixes https://github.com/intel/intel-vaapi-driver/issues/533
Fixes https://github.com/intel/libva/issues/548
Closes https://github.com/intel/libva/pull/549 (duplicate)
Closes https://github.com/intel/libva/pull/550 (duplicate)
See https://gitlab.freedesktop.org/mesa/mesa/-/blob/21.3/include/pci_ids/crocus_pci_ids.h
Dogfooding via https://github.com/freebsd/freebsd-ports/commit/bbdb8c12e8ac

Disclaimer: Runtime hasn't been tested. I only have Skylake where this is nop.

PR #549, opened 2021-11-10
VA/X11: add mapping for crocus DRI driver

airlied:
Just add a mapping for crocus


jbeich:
Duplicate of https://github.com/intel/libva/pull/547 ?

PR #550, opened 2021-11-12
Add DRM/X11 mapping for crocus

kwizart:


va/drm/va_drm_utils.c
```diff
@@ -37,6 +37,8 @@ struct driver_name_map {
 };
 
 static const struct driver_name_map g_driver_name_map[] = {
+    { "crocus",     6, "iHD"    }, // Intel Media driver
+    { "crocus",     6, "i965"   }, // Intel OTC GenX driver
```
jbeich, line 41: `crocus` isn't a valid kernel DRM driver name.
https://github.com/intel/libva/blob/453002ce69779c713f6f8315bedce140d34ba805/va/drm/va_drm_utils.c#L61-L69

PR #553, opened 2021-11-18
Add description of the fd for surface importing

XinfengZhang:

### DIFF
--- a/va/va_drmcommon.h
+++ b/va/va_drmcommon.h
@@ -116,6 +116,10 @@ struct drm_state {
  * support NV12 surfaces when they are contained within a single DRM
  * object, and therefore fail to create such surfaces if the two
  * planes are in different DRM objects.
+ * Note that backend driver will retrieve the resource represent by fd,
+ * and a valid surface ID is generated. Backend driver will not close
+ * the file descriptor. Application should handle the release of the fd.
+ * releasing the fd will not impact the existence of the surface.
  */
 typedef struct _VADRMPRIMESurfaceDescriptor {
     /** Pixel format fourcc of the whole surface (VA_FOURCC_*). */


rosetta-jpn (approved):
LGTM! Thanks.

va/va_drmcommon.h
```diff
@@ -116,6 +116,11 @@ struct drm_state {
  * support NV12 surfaces when they are contained within a single DRM
  * object, and therefore fail to create such surfaces if the two
  * planes are in different DRM objects.
+ * Note that backend driver will retrieve the resource represent by fd,
+ * and a valid surface ID is generated. Backend driver will not close
+ * the file descriptor. Application should handle the release of the fd.
+ * releasing the fd  will not impact the existence of the surface.
+ *
```
xhaihao, line 123: How is the surface generated ? If the surface depends on `fd`,  does it make sense  that the surface can be used after the release of the `fd` ?

XinfengZhang: fd is created and released by application. vaSurface associate also have pair of (vaCreateSurfaces, vaDestroySurfaces).  
it is good question that whether surface can be used after the release of fd.
from my understanding , it depend on who will handle the real memory allocate/free. fd or surface? my initial idea is: fd and vaSurface both represent the surface, when the fd already exist, create surface basing on the fd means the ref+1 for the memory. the real memory free will happen when ref==0. 
but maybe there are another behavior with other backend driver developer. the real memory is combined with fd. the call sequence must be
create fd
create surface
destroy surface
destroy fd
so, to be safe, let me remove the last sentence, thanks.

XinfengZhang: I added the last sentence back, because if there are special implement that destroy fd will impact the surface usage, the fd should  be "could  not be closed" or "failed to close"

xhaihao (approved):
LGTM

PR #554, opened 2021-11-23
Add nvidia-drm to the DRM driver map

elFarto:
This PR adds the nvidia-drm driver to the DRM driver map. It maps it to existing 'nvidia' VA-API driver, matching the X11 driver.

This change is needed to support an upcoming VA-API implementation based off NVDEC, currently not public.

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -45,6 +45,7 @@ static const struct driver_name_map g_driver_name_map[] = {
     { "nouveau",    7, "nouveau"  }, // Mesa Gallium driver
     { "radeon",     6, "r600"     }, // Mesa Gallium driver
     { "amdgpu",     6, "radeonsi" }, // Mesa Gallium driver
+    { "nvidia-drm",10, "nvidia"   }, // NVIDIA driver
     { NULL,         0, NULL }
 };
 


XinfengZhang:
so, NVDEC backend driver of libva is not public now,  could the binary be downloaded now?, is there a method to test it? it will be a open source one or close source one?

elFarto:
Correct, the backend is currently not released. The only useful application for this backend is Firefox, which requires some addition patches to be able to use it (in addition to this patch). 

It will be an open source backend. I can give you access to the repo if you wish to test it yourself.

XinfengZhang:
> Correct, the backend is currently not released. The only useful application for this backend is Firefox, which requires some addition patches to be able to use it (in addition to this patch).
> 
> It will be an open source backend. I can give you access to the repo if you wish to test it yourself.

thanks , I maybe not test it, but just read

elFarto:
The repository has now been made public: https://github.com/elFarto/nvidia-vaapi-driver

elFarto:
I don't think it's the place of libva to pick the specific implementation that should be used for the nvidia-drm driver. This can be achieved using symlinks from nvidia_drv_video.so to the chosen implementation rather than hardcoding it into library.

If NVIDIA want to have their own implementation, it will simply supersede any other implementation (since it should in theory work perfectly...). If you were to hardcode the name other than `nvidia`, then they'd need to raise a similar request to get the library to use changed.

In terms of actual names, only `nvidia` and `nvidia-drm` make any sense, and since the X11 driver already points to `nvidia`, I think we should use the same as the implementation should be able to handle any differences between being called via X11 and via DRM.

Of course, this issue only arises because you have the names hardcoded in the library, with no external config mechanism (like /usr/share/glvnd/egl_vendor.d/*.json for EGL implementations) and no fallback mechanism, like trying the kernel driver name + "_drv_video.so".

SupervisedThinking:
@XinfengZhang 
Well until today NVIDIA gave a fu... about VA-API support for dedicated graphic cards & it doesn't look like they are going to deliver support anytime soon. 

If some random, nobody dev (@elFarto no offense 😝 ) has to step up to write an open source wrapper to fix this for countless users I don't see why NVIDIA should make it's move now. He did the job for them & all what's left is some fixing & optimizations because his driver works very well even at a pretty early state.

If NVIDIA will ever offer an open source driver by themselves https://github.com/elFarto/nvidia-vaapi-driver is superflous anyways because, as @elFarto stated, it should work perfect / cover all needs anyway - then again they can raise a PR or just stick to `LIBVA_DRIVER_NAME=nvidia-unicorn` or whatever.

XinfengZhang:
thanks @elFarto @SupervisedThinking .  I have no other concerns

jonathanbian:
LGTM

jonathanbian (approved):
Looks good

XinfengZhang (approved):

kwizart:
> I don't think it's the place of libva to pick the specific implementation that should be used for the nvidia-drm driver. This can be achieved using symlinks from nvidia_drv_video.so to the chosen implementation rather than hardcoding it into library.

The problem with symlinks is that they needs to point to an implementation and there are some others implementations that exist (despite their limitations) such as libva-vdpau-driver and the vp9 fork(s)
From a package perspective , the symlink needs to be owned by the vaapi backend implementation package and the consequence for several packages to provide different symlinks is that they will conflicts and cannot co-exists.

Previous way to solve this issue was to use update-alternatives, but this doesn't cope with read-only rootfs distribution (such as rpm-ostree and alikes). So this is deprecated.

Just for clarity and comparison , this is not how mesa drivers are loaded, by choosing to update a symlink to choose the correct implementation. 
I think symlink defeat the point of using  a mapping here as a means to determine precise name for the implementation of the vaapi backend.
So I prefer a correct mapping with a better name backend than relying on a symlink.

So I would suggest to use nvidia_codec or nvdec_nvenc (if ever nvenc might be supported at some point)...

marioroy:
Using NVIDIA graphics in 2022? There are two VA-API implementations, NVDEC and VDPAU. See the discussion.

https://bugzilla.redhat.com/show_bug.cgi?id=2133752

To summarize, I chose the path that both packages may co-exist. The nvidia symbolic link exists and points to nvdec by default. I renamed nvidia_drv_video.so to nvdec_drv_video.so in the nvdec package.

Currently:

Two packages exist, but only one can be installed due to naming conflict.

Co-exist path:

Both packages co-existing. The nvdec package provides the symbolic link, whereas the vdpau package does not.
```text
-rwxr-xr-x. 1 root root  87096 Oct 11 04:17 nvdec_drv_video.so
lrwxrwxrwx. 1 root root     18 Oct 11 04:17 nvidia_drv_video.so -> nvdec_drv_video.so
-rwxr-xr-x. 1 root root 120032 Oct 11 04:17 vdpau_drv_video.so
```

Minimally, one does not need to set LIBVA_DRIVER_NAME unless running Chrome. The vdpau backend is what works in Chrome at this time (though, only in Xorg).

The bugzilla ticket is awareness on how one may enjoy hardware acceleration in Firefox and Chrome.

mirh:
I'm sorry, but something isn't quite checking out here. 
If this PR had actually been working as expected.. why is it so that after #617 you started to have to manually specify `LIBVA_DRIVER_NAME`, [least](https://github.com/intel/libva/issues/675) nothing works? Be it the old ass VDPAU or the new NVDEC drivers. 
And even then, if you don't eventually re-enable NVCTRL you are still getting *vaGetDriverNames* errors.

Issue #38, opened 2017-04-07
Use a unified coding style in libva source code


xhaihao:
We can run the script in the top-level directory to unify the coding style.

PR #556, opened 2021-12-08
ci: fix freebsd build by switching to vmactions/freebsd-vm@v0.1.5

dvrogozh:

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -12,11 +12,11 @@ on:
 
 jobs:
   freebsd:
-    runs-on: macos-latest
+    runs-on: macos-10.15
     steps:
     - uses: actions/checkout@v2
     - name: test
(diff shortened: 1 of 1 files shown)


dvrogozh:
@jbeich : any suggestions why this fails? But it seems to me this is an issue with the host (which is macos) or action itself. So, I will file an issue to them.

XinfengZhang (approved):

dvrogozh:
I think I understood what's wrong. We were using "macos-latest" for the host which previously was 10.15.7. Now systems start getting update to 11.6.1 which miss `vboxmanage`. I s/macos-latest/macos-10.15/ and ci started to pass.

dvrogozh:
This is change solely in freebsd.yml, ubuntu workloads won't start. Can be merged.

dvrogozh:
fyi, I submitted [#38](https://github.com/vmactions/freebsd-vm/issues/38) to note behavior on macos-latest runners.

PR #564, opened 2022-01-20
Change the default log level from info to errors

nmiell:
libva by default writes log messages to stdout, but the stdouts of GUI
applications are connected to the systemd journal these days.

This results in a lot of unnecessarily repeated log messages of the form
> libva info: VA-API version 1.13.0
> libva info: Trying to open /usr/lib64/dri/radeonsi_drv_video.so
> libva info: Found init function __vaDriverInit_1_13
> libva info: va_openDriver() returns 0
as e.g. Firefox creates and destroys video decoders for page elements.

This patch reduces the default log level from info to errors only.

Signed-off-by: Nicholas Miell <nmiell@gmail.com>

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -119,8 +119,12 @@ int vaDisplayIsValid(VADisplay dpy)
  * Global log level configured from the config file or environment, which sets
  * whether default logging appears or not (always overridden by explicitly
  * user-configured logging).
+ *
+ * 0 = nothing
+ * 1 = errors only
+ * 2 = informational messages
  */
-static int default_log_level = 2;
+static int default_log_level = 1;
 
 static void default_log_error(void *user_context, const char *buffer)
 {


XinfengZhang:
I have no much concern about this change, people always could use LIBVA_MESSAGING_LEVEL to set the log level.

nmiell:
> I have no much concern about this change, people always could use LIBVA_MESSAGING_LEVEL to set the log level.

Yeah, that's what I suggested for the Fedora package but they told me to take up with you. I'll go back to them now.

kwizart:
> I have no much concern about this change, people always could use LIBVA_MESSAGING_LEVEL to set the log level.
The point is that "individuals" can tune the verbosity on a per-system basis. Sure.

But users managing many systems or application developers probably expect verbosity level to be same across several components.
After all, It's not that usual, for libraries, to emit non error messages in the "back" of the application.

The point is for this request is to have a "sane default" , so that's why, as the Fedora package maintainer, I'm requesting to have this request held upstream over on a distro basis.

Maybe we can imagine an API to fine tune the verbosity setting. So using vainfo will enable level 2, while firefox will use libva while using level 1 only. (or level 2 when the firefox verbosity related to video is raised).

XinfengZhang:
it make sense, if most of libraries behind frontend application just emit error messages , libva better to have similar default log levels.
if there are issues reported, developer could increase the level to get the enough information. 
ask a dumb question, there are a lot of duplicate libva log? because currently the log function  only be called during initialization process, so there are multiple times initializations ?

nmiell:
Firefox appears to initialize the library every time it encounters a `<video>` element or similar.

Issue #569, opened 2022-02-02
[glx] fbconfig[0] is not 8 bits RGB with Mesa 18.x or newer


kebszeone:
Mesa has enabled "allow_rgb10_configs" by default. Note that 10bpc affects both
GLX and EGL. The requirement from EGL is that 10bpc formats must be first in
the list of configs. This breaks applications that just use the first config,
which used to be 8bpc and now it's 10bpc. Here's related bug report and some
discussions from Mesa list:
https://bugs.freedesktop.org/show_bug.cgi?id=109548
https://lists.freedesktop.org/archives/mesa-dev/2019-February/214610.html

In va/glx/va_glx_impl.c: `static int create_tfp_surface(VADriverContextP ctx, VASurfaceGLXP pSurfaceGLX)` 
```
    glx_pixmap = pOpenGLVTable->glx_create_pixmap(
                     ctx->native_dpy,
                     fbconfig[0],
                     pixmap,
                     pixmap_attrs
```
Chossing fbconfig[0] breaks applications wanting 8bit config on mesa 18X or newer, or when `allow_rgb10_configs=true` is set.
You should iterate over FBConfig and select 8 bit color size !

kebszeone:
Acutally, juste specifying the alpha size seems to be enough.
```
diff --git a/va/glx/va_glx_impl.c b/va/glx/va_glx_impl.c
index 3ad01a6..c1ca999 100644
--- a/va/glx/va_glx_impl.c
+++ b/va/glx/va_glx_impl.c
@@ -521,6 +521,7 @@ static int create_tfp_surface(VADriverContextP ctx, VASurfaceGLXP pSurfaceGLX)
         GLX_RED_SIZE,           8,
         GLX_GREEN_SIZE,         8,
         GLX_BLUE_SIZE,          8,
+        GLX_ALPHA_SIZE,         8,
         /*
          * depth test isn't enabled in the implementaion of VA GLX,
          * so depth buffer is unnecessary. However to workaround a
(END)
```

jbrownsw:
The PR was closed without being merged and I just ran into it so I don't think it can be closed.

Issue #473, opened 2020-11-17
vaCopySurfaceGLX produces corrupt image


UnscrewLater:
I have an app that uses vaCopySurfaceGLX.  On some systems it works OK but others it produces a highly contoured magenta & yellow coloured image.

The problem is evident on an i7 8700 (Coffee Lake) system with integrated graphics running Ubuntu 20.04.1 with mesa 20.0.8.  However an old i5 661 (Iron Lake) and the same Ubuntu 20.04.1 works fine.  The problem persists when using libva built from git master.  I've also tried building the latest intel-vaapi-driver but no change.

I believe there is a bug in the function create_tfp_surface() when it creates a framebuffer by calling glXChooseFBConfig.  If I add 'GLX_ALPHA_SIZE, 8,' to the attribute list the problem is fixed.  This should be included as the GLX_RENDER_TYPE is set to GLX_RGBA_BIT.

I have attached a patch that corrects the problem:
[0001-GLX-Specify-an-alpha-size-when-requesting-a-GLX-fram.patch.txt](https://github.com/intel/libva/files/5552819/0001-GLX-Specify-an-alpha-size-when-requesting-a-GLX-fram.patch.txt)

imbens:
I can confirm the problem on Ubuntu 20.04.1 with i5 750 and Radeon RX550. The attached patch fixes the problem.

XinfengZhang:
could you help file PR with attached patch?

UnscrewLater:
I'm happy to help out but have some reservations:
1. Some GPUs are unaffacted. This indicates to me that the real fault lies outside of libva, possibly with glX (most likely) or in the intel-vaapi-driver or kernel as these have hardware dependent paths.
2. The patch supplied is insufficient as-is and needs to include more attributes from the target texture.  This works around the current undefined behavior in glX and/or driver.
3. I don't have, nor do I want, a github account to create a working tree to pull from.
4. I'm happy to provide a git compatible patch incorporating the changes discussed in (2). I have incorporated these changes into my own app which has tested OK over several platforms.

kebszeone:
Since mesa 18.x or higher allow_rgb10_configs is set to true by default ! Can you check with setting it to false ?
But still I think your patch is a good solution, or we could iterate over the fbconfig and choose the proper one

lancezac:
[...]

PR #570, opened 2022-02-02
va_glx_impl: set `GLX_ALPHA_SIZE, 8` when creating pixmap

kebszeone:
Since Mesa 18.X or higher allow_rgb1O_configs is set to true by default,
Not specifying the alpha leads to taking a 1O bit GLXFBConfig which leads
to corrupted rendering

Signed-off-by: Mouhamad Kebe <mouhamad.kebe@blade-group.com>

### DIFF
--- a/va/glx/va_glx_impl.c
+++ b/va/glx/va_glx_impl.c
@@ -521,6 +521,7 @@ static int create_tfp_surface(VADriverContextP ctx, VASurfaceGLXP pSurfaceGLX)
         GLX_RED_SIZE,           8,
         GLX_GREEN_SIZE,         8,
         GLX_BLUE_SIZE,          8,
+        GLX_ALPHA_SIZE,         8,
         /*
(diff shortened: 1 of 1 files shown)


kebszeone:
Lined issues https://github.com/intel/libva/issues/569 https://github.com/intel/libva/issues/473

UnscrewLater:
The function create_tfp_surface() needs several more related fixes to remove assumptions about the texture. The attached patch contains fixes that I have been using for more than a year now:
[0001-GLX-Specify-an-alpha-size-when-requesting-a-GLX-fram.txt](https://github.com/intel/libva/files/7987994/0001-GLX-Specify-an-alpha-size-when-requesting-a-GLX-fram.txt)
.

kebszeone:
@UnscrewLater your patch fixes the issue on my setup indeed. what's taking so long for them to integrate it ?

PR #571, opened 2022-02-07
Added Intel Media Engine Definition.

FurongZhang:
Added copy engine selection.

Signed-off-by: 
Furong Zhang <furong.zhang@intel.com>
Rogozhkin, Dmitry V <dmitry.v.rogozhkin@intel.com>
Zhang, Carl <carl.zhang@intel.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -371,6 +371,9 @@ typedef int VAStatus;   /** Return status type from functions */
 #define VA_EXEC_MODE_DEFAULT      0x0
 #define VA_EXEC_MODE_POWER_SAVING 0x1
 #define VA_EXEC_MODE_PERFORMANCE  0x2
+/** Values 3-7 are reserved for future extension. */
+/** Vendor specific operation mode*/
+#define VA_EXEC_MODE_CUSTOM       0x8
 
 /* Values used to describe device features. */
 /** The feature is not supported by the device.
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -5179,13 +5179,19 @@ typedef struct _VACopyObject {
...
+#define VA_INTEL_MEDIA_ENGINE_VDBOX                 0x1
+#define VA_INTEL_MEDIA_ENGINE_VEBOX                 0x2
+#define VA_INTEL_MEDIA_ENGINE_SFC                   0x4
+
 typedef union _VACopyOption {
     struct {
         /** \brief va copy synchronization, the value should be /c VA_EXEC_SYNC or /c VA_EXEC_ASYNC */
         uint32_t va_copy_sync : 2;
-        /** \brief va copy mode, the value should be VA_EXEC_MODE_XXX */
+        /** \brief va copy mode, if va_copy_engine_selection is 0, the value should be VA_EXEC_MODE_XXX; else, the value should be VA_INTEL_MEDIA_ENGINE_XXX on Intel platforms */
```
XinfengZhang, line 5190: how about the capability reporting, currently it was defined by (1<<VA_EXEC_MODE_XXX) | (1 << VA_EXEC_MODE_XXX)

dvrogozh: And what the value should be on non-Intel platforms?

As of now VAAPI is kept abstract from HW. We never exposed Intel specific engine names and always hide that under something like POWERSAVE, PERFORMANCE, etc.

Basically, existing modes were designed specifically for that and I don't see the point why we should introduce something else. I suggest you map existing stuff to driver capabilities internally:
```
/usr/include/va/va.h:370:#define VA_EXEC_MODE_DEFAULT      0x0
/usr/include/va/va.h:371:#define VA_EXEC_MODE_POWER_SAVING 0x1                
/usr/include/va/va.h:372:#define VA_EXEC_MODE_PERFORMANCE  0x2
```

To help customers use that we can define intel extension here in libva. For example:
```
$ cat /usr/include/va/va_intel.h
#define VA_INTEL_MEDIA_ENGINE_VDBOX VA_EXEC_MODE_XXX
#define VA_INTEL_MEDIA_ENGINE_VEBOX VA_EXEC_MODE_YYY
#define VA_INTEL_MEDIA_ENGINE_SFC VA_EXEC_MODE_ZZZ
```

dvrogozh (changes requested):

FurongZhang:
It seems there is some need to define some Intel specific items (this need is not specific to copy function). I am thinking about this.

I once thought about the option to split into another file. But then, 
1) the VAAPI user need to refer to the new file "va_intel.h" in their code. 
2) not quite sure if "va.h" needs "va_intel.h".

PR #573, opened 2022-02-08
add new execution mode for more hw capabilities

### LINKED ISSUE #572: new execution mode needed for Intel platform
Intel platform has 3 path vaCopy support, EU, BLT, VEBox
so, beside current definitions ,  POWER_SAVING, PERF , need new definitions to mapping these 3 path. 
backend driver should select one as DEFAULT.

XinfengZhang:
Fixes #572
Closes #571(duplicate, same purpose)

add VA_EXEC_MODE_POWER_SAVING_ALT definition

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -367,10 +367,24 @@ typedef int VAStatus;   /** Return status type from functions */
 /** asynchronization,application should call additonal sync operation to access output */
 #define VA_EXEC_ASYNC             0x1
 
-/** operation mode */
+/** operation mode
+  * Operation mode is defined to distinguish the code path if there are different algorithm
+  * or hw implementation candidates. Application could query which implementation supported
+  * in backend driver and select one by demand.
+  * VA_EXEC_MODE_DEFAULT is the basic recommended one from backend driver, it could be same with
+  * one of VA_EXEC_MODE_XXX on different platform.
+  */
+
 #define VA_EXEC_MODE_DEFAULT      0x0
 #define VA_EXEC_MODE_POWER_SAVING 0x1
 #define VA_EXEC_MODE_PERFORMANCE  0x2
+/*
+ * VA_EXEC_MODE_3 is defined for the third solution. It is for the developer with
+ * the expertise and know enough information about the backend driver, and want to balance
+ * the hardware workload for multiple session usage. Backend driver may provide more
+ * implementation to bring in more customization/flexibility.
+ */
+#define VA_EXEC_MODE_3            0x3
 
 /* Values used to describe device features. */
 /** The feature is not supported by the device.


hye5 (approved):

fenhu (approved):

FurongZhang (commented):
VA_EXEC_MODE_POWER_SAVING_ALT is not clear to the VAAPI user. Don't recommend this. 
2 questions: 
1) When the user should select this one VA_EXEC_MODE_POWER_SAVING_ALT? 
2) What the user should expect from these 3 flags?

FurongZhang (changes requested):
Request to address the questions, then to merge the PR.

va/va.h
```diff
@@ -367,11 +367,30 @@ typedef int VAStatus;   /** Return status type from functions */
...
+  * But, if there are similar HW functions, application could select second one if the first
+  * HW is busy. From design perspective, POWER_SAVING is always prior "power saving" mode
+  * than POWER_SAVING_ALT.
+  *
+  * Take Intel Alchemist as an example: it could implement "vaCopy" through EU(execution unit),
+  * BLT(blitter engine), VEBox(video enhancement engine). both BLT and VEBox is fix function.
+  * Intel may suggest people to select BLT for power saving purpose. But, if BLT is too busy,
+  * or there are some special request, they could select VEBox (POWER_SAVING_ALT) as backup path.*/
+
+#define VA_EXEC_MODE_POWER_SAVING_ALT 0x3
```
FurongZhang, line 392: I think this naming is causing more confusion to VAAPI user and not flexible to extension.

XinfengZhang: it is simple and straight forward to add a new mode, right? "alternative" is a word which already has the native meaning "second, backup, another choose", suppose in future, if you have two "perf" implementation, then a PERFORMANCE_ALT needed. or you think POWER_SAVING_2 or POWER_SAVING_ALTERNATE is better?

dvrogozh: Folks, I think we need to stop here and avoid introduce POWER_SAVING_2 or POWER_SAVING_ALTERNATE modes. This will give more confusion than help. :)

Really, let's just do this:
```
$ cat va.h
#define VA_EXEC_MODE_DEFAULT      0x0
#define VA_EXEC_MODE_POWER_SAVING 0x1
#define VA_EXEC_MODE_PERFORMANCE  0x2
#define VA_EXEC_MODE_CUSTOM1  0x3
#define VA_EXEC_MODE_CUSTOM2  0x4
#define VA_EXEC_MODE_CUSTOM3  0x5
...

$ cat va_intel.h

#define VA_EXEC_MODE_BLT         VA_EXEC_MODE_XXX
#define VA_EXEC_MODE_VDBOX   VA_EXEC_MODE_YYY
#define VA_EXEC_MODE_VEBOX    VA_EXEC_MODE_ZZZ
```

So, people which  really want these Intel specific settings will just include a new file `va_intel.h` and use that. Those who don't need this - will use one of the 3 existing modes with clear meaning (default, power_save, performance).

dvrogozh (changes requested):

va/va.h
XinfengZhang: if use 
#define VA_EXEC_MODE_BLT         VA_EXEC_MODE_XXX
#define VA_EXEC_MODE_VDBOX   VA_EXEC_MODE_YYY
#define VA_EXEC_MODE_VEBOX    VA_EXEC_MODE_ZZZ
it means that the application have to know the backend vendors, and provide different path for different vendor.
for example:
when you report value 1, application does not know what it means, it is BLT or  POWER_SAVING?  from perspective of application developer, it is confusing. if these definitions is just some description about intel implementation of current mode , why just not add comments?

adding vendor specific customized parameters is our target , and will add some API , such as vendor query...  but I dont think this resolve current problem. current problem is there are 3 modes, we just defined 2.

dvrogozh: To be honest, I think that writing va_intel.h would be something application developers will end up with as soon as we will introduce POWER_SAVE_2 since it's absolutely unclear when to use POWER_SAVE or POWER_SAVE_2. What's the difference? From these names it can't be deduced and clogs the code. To help understand the code redefinitions will be used. Better to introduce them ourselves in advance to avoid different definitions by different application vendors.

XinfengZhang: even for this purpose, I think we could add POWER_SAVE_ALT firstly , and add the comments to describe Intel mapping HW for these 3 definitions .  then, add va_intel.h in future. vendor specific extension is long term thing, also need related query APIs, different flags to indicate what kind of parameter is for vendors etc. 
and still need to check whether it is suitable that let application developers aware of these vendor differences. 

I think maybe we could filed an issue to discuss it , and collect the feature list which need to distinguish the vendors.
and how to handle it for different vendors.

FurongZhang:
Agree with @dvrogozh . @XinfengZhang , please take a look at https://github.com/intel/libva/pull/571. 

I disagree with the naming VA_EXEC_MODE_POWER_SAVING_ALT which does not make sense to me. 

If @XinfengZhang  really don't want to add Intel specific definition, please add the below as my PR for the temporary WA. 
/** Values 3-7 are reserved for future extension. */
/** Vendor specific operation mode*/
#define VA_EXEC_MODE_CUSTOM       0x8

Strongly object to merge VA_EXEC_MODE_POWER_SAVING_ALT into VAAPI.

va/va.h
FurongZhang: If only adding a new flag, how about the below in the PR https://github.com/intel/libva/pull/571. At least, it allows future extension. Meanwhile, it also serves the current purpose. 
/** Values 3-7 are reserved for future extension. /
/* Vendor specific operation mode*/
#define VA_EXEC_MODE_CUSTOM 0x8

dvrogozh: Is there some other meaningful attribute we can use to differentiate modes? Maybe VA_EXEC_MODE_LATENCY?

XinfengZhang: Yes, We need a meaningful attribute, LATENCY or LOW_LATENCY, actually VEBox path is not recommended now, but just as I mentioned before. I dont know who will use it, but maybe in future, there are extremely numerous sessions, or other task need BLT and EU, then maybe it could be a fallback path. 

have several concerns about #571, 1. it does not add the query support , when APP query the attributes, APP get value 1, but not know what it represent. 2. there are similar way, for example: add execution HW which using another 4-8 bit. but certainly will bring other concerns about vendor specific definition. 3. complex, from interface perspective, "simple” also should be considered.

xhaihao: From user's perspective, I don't think `VA_EXEC_MODE_CUSTOM` is meaningful because I don't know what this attribute will be used for.

va/va.h
```diff
@@ -367,10 +367,28 @@ typedef int VAStatus;   /** Return status type from functions */
 /** asynchronization,application should call additonal sync operation to access output */
 #define VA_EXEC_ASYNC             0x1
 
-/** operation mode */
+/** operation mode
+  * operation mode is defined to distinguish the code path if there are different algorithm
```
dvrogozh, line 371: Fix here and below: sentences should start with capital letters.

va/va.h
```diff
@@ -367,10 +367,28 @@ typedef int VAStatus;   /** Return status type from functions */
...
 #define VA_EXEC_MODE_DEFAULT      0x0
 #define VA_EXEC_MODE_POWER_SAVING 0x1
 #define VA_EXEC_MODE_PERFORMANCE  0x2
+/*
+ * VA_EXEC_MODE_3 is defined for the third solution. it is for the developer with
+ * the expertise and know enough information about the backend driver, and want to balance
+ * the hardware workload for multiple session usage. and backend driver may provide more
+ * implementation to bring in more customization/flexibility.
+ * The definition implies VA_EXEC_MODE_1 VA_EXEC_MODE_2 also is defined like
+ * #define VA_EXEC_MODE_1 VA_EXEC_MODE_POWER_SAVING
```
dvrogozh, line 387: I don't get the point here. You did not define neither VA_EXEC_MODE_1 nor VA_EXEC_MODE_2. Either define them or drop this comment.

dvrogozh (changes requested):

PR #578, opened 2022-03-01
trace: convert the log of profile/entrypoint to string

XinfengZhang:
string is better to be understood


va/va_trace.c
```diff
@@ -1091,12 +1091,12 @@ void va_TraceCreateConfig(
 
     TRACE_FUNCNAME(idx);
 
-    va_TraceMsg(trace_ctx, "\tprofile = %d\n", profile);
-    va_TraceMsg(trace_ctx, "\tentrypoint = %d\n", entrypoint);
+    va_TraceMsg(trace_ctx, "\tprofile = %s\n",  vaProfileStr(profile));
```
dvrogozh, line 1094: https://github.com/intel/libva/blob/9e4bdc5c4f7f8549187badb8a5111c1b07c3c41c/va/va.h#L502

is missing in vaProfileStr implementation:
https://github.com/intel/libva/blob/9e4bdc5c4f7f8549187badb8a5111c1b07c3c41c/va/va_str.c#L29

XinfengZhang: yes, but it is deprecated, H264Baseline is too complex, so ConstrainedBaseline is introduced for most of usage case.   I wonder whether there are real H264Baseline usage.

XinfengZhang: we could , but there will be such warning with compiling
va_str.c: In function ‘vaProfileStr’:
va_str.c:39:9: warning: ‘VAProfileH264Baseline’ is deprecated [-Wdeprecated-declarations]
         TOSTR(VAProfileH264Baseline);
         ^~~~~

XinfengZhang: or we could add default: "unknown or deprecated profile" in vaProfileStr

dvrogozh: Maybe we should print a number for unknown or deprecated profile?

XinfengZhang: yes, agree, will update

dvrogozh: I don't see this done. Did you forget to amend?

XinfengZhang: hi @dvrogozh  are you talking about https://github.com/intel/libva/pull/578/files#diff-9e128fca5567f303d7b25909040cf5c6ecf260303e64bfcdc5eebd19d8917dfeR1094-R1095? it should already been changed.

xhaihao:
The patch prints a string for type too, it'd better to update the commit short log

xhaihao (approved):
LGTM

PR #580, opened 2022-03-07
Improve tracing slightly for VP8/9

dwlsalmeida:


va/va_trace.c
```diff
@@ -5226,6 +5227,7 @@ static void va_TraceVP8Buf(
         va_TraceVASliceParameterBufferVP8(dpy, context, buffer, type, size, num_elements, pbuf);
         break;
     case VASliceDataBufferType:
+        va_TraceVABuffers(dpy, context, buffer, type, size, num_elements, pbuf);
```
XinfengZhang, line 5230: removing line 5229 is enough? because there are default:

XinfengZhang: @dwlsalmeida , will you change it?

PR #584, opened 2022-03-30
X11 DRI2: check if device is a render node

### LINKED ISSUE #582: automatic backend driver loading does not work in xorgxrdp
xrdp can spawn a Xorg server that uses drm  (/dev/dri/renderD*) as backend. As such opengl and vulkan work out of the box.
video acceleration seems to work to with also with at least radeonsi and iHD if forced with the appropriate environment variable (LIBVA_DRIVER_NAME=...)

Automatic selection of the driver however does not work.

I found that in va/x11/dri2_util.c there is a call to drmGetMagic that fails since there is no drm master in this case (Xorg runs on top of a render node, not card)

If I modify the code to check if dri_state->base.fd is a render node if drmGetMagic fails, skip VA_DRI2Authenticate it seems to do the trick.

Since I'm not a Xorg expert I don't know if this does not break something, but as far as I understand the whole drmGetMagic and subsequent authentication is only a /dev/dri/card thing (like in drm/va_drm.c in function va_DisplayContextGetNumCandidates, where it specifically checks if it is not a render node and if so it skips the magic/authentication steps)

akarl10:
and if so skip drmGetMagic and VA_DRI2Authenticate.

This is essentailly the same as in the pure drm non X11 case

fixes #582

### DIFF
--- a/va/x11/dri2_util.c
+++ b/va/x11/dri2_util.c
@@ -25,6 +25,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <assert.h>
+#include <sys/stat.h>
 
 #include <xf86drm.h>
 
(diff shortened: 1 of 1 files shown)


XinfengZhang (commented):
LGTM now, thanks

XinfengZhang (approved):

XinfengZhang:
hi @akarl10 , I will squash your two commit to one, and keep the description of first commit, is it ok to you?

akarl10:
@XinfengZhang of course thats ok 👍

XinfengZhang:
sorry, I could not apply "squash & merge" , could you help to squash these two commits, the second one looks a part of first commit.

XinfengZhang:
image

akarl10:
> image

I think I have enabled everything..
image

But nevertheless now there is only one commit in this commit

PR #598, opened 2022-05-12
Added VPP Super Resolution Filter.

FurongZhang:
Add Super Resolution VPP filter including parametrization and capability specification.

               Yanfeng Mi <yanfeng.mi@intel.com>
	       Noah Zhang <noah.zhang@intel.com>
	       Ben Lin <ben.y.lin@intel.com>

### DIFF
--- a/va/va_vpp.h
+++ b/va/va_vpp.h
@@ -255,6 +255,8 @@ typedef enum _VAProcFilterType {
     VAProcFilterHighDynamicRangeToneMapping,
     /** \brief Three-Dimensional Look Up Table (3DLUT). */
     VAProcFilter3DLUT,
+    /** \brief Super Resolution. */
+    VAProcFilterSuperResolution,
     /** \brief Number of video filters. */
     VAProcFilterCount
 } VAProcFilterType;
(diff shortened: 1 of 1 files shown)


jonathanbian:
A couple of comments:

How are "Conference" and "Camera" modes differ from the default mode? Maybe we can use more generic terms for these two modes so it's more intuitive for the user to know what is different about these modes, and these modes are not necessarily limited to conference or camera usages.

Does it make sense to have the user query the supported scaling factors in VAProcFilterCapSuperResolution?

FurongZhang:
Thank you so much @jonathanbian . Your comments make sense to me. I will think about them, then update the PR(it may need some time).

va/va_vpp.h
```diff
@@ -1447,6 +1449,75 @@ typedef struct _VAProcFilterCap3DLUT {
...
+    /**
+     * \brief Super Resolution Conference Mode.
+     * In this mode, Super Resolution is optimized for video conference scenario.
+     */
+    VAProcSuperResolutionConference,
+    /**
+     * \brief Super Resolution Camera Mode.
+     * In this mode, Super Resolution is optimized for camera scenario.
+     */
+    VAProcSuperResolutionCamera,
```
XinfengZhang, line 1473: sorry, I am confused by the definition of camera and conference,  conference also use Camera , right? could we define clear mode , or usage scenario ?

va/va_vpp.h
```diff
@@ -1447,6 +1449,75 @@ typedef struct _VAProcFilterCap3DLUT {
...
+    /**
+     * \brief Number of Super Resolution Mode.
+     */
+    VAProcSuperResolutionCount,
```
XinfengZhang, line 1477: better not define such maximum enum value to avoid use of implementation .

va/va_vpp.h
```diff
@@ -1447,6 +1449,75 @@ typedef struct _VAProcFilterCap3DLUT {
...
+ *     VAProcFilterSuperResolution,
+ *     &sr_caps, &num_sr_caps);
+ */
+typedef struct _VAProcFilterCapSuperResolution {
+    /** \brief Super Resolution mode, should be one of VAProcSuperResolutionXXX.*/
+    VAProcSuperResolutionMode            mode;
+    /** \brief The maximum value of input width SR filter is capable of supporting.*/
+    uint32_t                             max_input_width;
+    /** \brief The maximum value of input height SR filter is capable of supporting.*/
+    uint32_t                             max_input_height;
```
XinfengZhang, line 1515: how about max output resolution , for example: I have a 720P input , but suppose the capability of HW maybe just support 1080P output, dont support 4K, how to describe the restriction?

XinfengZhang (changes requested):

va/va_vpp.h
FurongZhang: ok. Agreed. Will change that.

va/va_vpp.h
FurongZhang: Super resolution scaling itself supports 2x. Hence, only max input is meaningful. Probably, need to document this information. And also consider Jonathan's comment.

FurongZhang:
Regarding Super Resolution Mode, I am proposing the below and discuss with other folks.
VAProcSuperResolutionDefault
VAProcSuperResolutionSharpen
VAProcSuperResolutionArtifactRemoval

va/va_vpp.h
FurongZhang: I updated the PR, please help to review again.

va/va_vpp.h
```diff
@@ -1447,6 +1449,102 @@ typedef struct _VAProcFilterCap3DLUT {
...
+ *  \brief Video Processing Super Resolution Predefined Mode.
+ *
+ *  Super Resolution is the process of recovering a High Resolution(HR) image/video
+ *  from a given Low Resolution(LR) image/video. A few modes are predefined for
+ *  different scenarios.
+ */
+typedef enum _VAProcSuperResolutionMode {
+    /**
+     * \brief Super Resolution Default Mode.
+     * This mode applies to video playback (especially natural video).
```
XinfengZhang, line 1462: 1,maybe changes to "this mode is tuned for video playback"? or "recommended to be used" as  next two mode.  and app always could use this mode for different usage scenarios , right? 
2, the default mode is not one of the other modes?

FurongZhang: > 1,maybe changes to "this mode is tuned for video playback"? or "recommended to be used" as next two mode. and app always could use this mode for different usage scenarios , right? 2, the default mode is not one of the other modes?

#1 Modified.
#2 Yes, the default is not one of the other modes.

xhaihao: for #2, it sounds this behavior doesn't keep consistency with other flags in VAAPI. The default is one of available modes for other flags.

va/va_vpp.h
```diff
@@ -1447,6 +1449,103 @@ typedef struct _VAProcFilterCap3DLUT {
...
+     *
+     * If output resolution / input resolution > max_scaling_ratio, it is up to
+     * backend implementation how to handle this case.
+     * Here, lists two possible options for the reference.
+     * Option #1: return VA_STATUS_ERROR_RESOLUTION_NOT_SUPPORTED;
+     * Option #2: input resolution * max_scaling_ratio with SR, the remaining
+     * scaling ratio could be performed by other scaling method if the backend
+     * can ganrantee the scaling quality.
+     */
+    float                                max_scaling_ratio;
```
xhaihao, line 1536: Because this is the capability returned from the backend driver, so `resolution / input resolution > max_scaling_ratio` shouldn't be allowed.

FurongZhang: max_scaling_ratio means the maximum scaling ratio which SR can supports.

PR #583, opened 2022-03-28
Add av1e hierarchy level info.

RachelChengRJ:
This is a change to add hierarchy information for AV1 encoder.

PR #607, opened 2022-06-17
Revert "Add av1e hierarchy level info."

XinfengZhang:
Reverts intel/libva#583

there are no issue with this PR itself
it is because media_driver has a violation, to faster enable features, it used reserved byte before the PR merged. after that , new libva + old media-driver will failed,  old libva + new media driver also will failed.
will re-aply similar patch which will use other reserved byte instead current 2 (reserved8b, reserved8bits0), 

and current 2 (reserved8b, reserved8bits0) will be kept there, should not change the variable name for next 2 releases

### DIFF
--- a/va/va_enc_av1.h
+++ b/va/va_enc_av1.h
@@ -249,10 +249,7 @@ typedef struct  _VAEncSequenceParameterBufferAV1 {
      */
     uint8_t     seq_tier;
 
-    /** \brief Indicates whether or not the encoding is in dyadic hierarchical GOP structure.
-     *  value range [0..1].
-     */
-    uint8_t    HierarchicalFlag; 
+    uint8_t     reserved8b;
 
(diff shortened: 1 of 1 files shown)


RachelChengRJ (approved):

xhaihao (approved):
LGTM

eero-t:
CI check preventing such regressions in future would also be good (e.g. in another PR, that checks whether latest media-driver release can be built with libva change).

XinfengZhang:
> CI check preventing such regressions in future would also be good (e.g. in another PR, that checks whether latest media-driver release can be built with libva change).

TBH, it is not a libva problem, it expose media_driver issue,  I dont think that media_driver should be built in libva project, libva should always keep compatibility, we could add some tools to check this kind of compatibility , but should not try to detect the issue of project which are using libva

PR #608, opened 2022-06-21
Add HierarchicalFlag & hierarchical_level_plus1 for AV1e.

RachelChengRJ:
This PR introduces new fields for VAEncSequenceParameterBufferAV1.

@eero-t , @xhaihao, @evelikov, @uartie : FYI - read caveat below.

Unfortunately, we've screwed up this some time ago in the driver and vpl rt and relied on reserved fields (which shouldn't have been done). So, that's a second attempt to reintroduce proper field values (first attempt - #607). Now the change is accompanied by respective changes in the driver and vpl rt. These changes needs to be merged basically at once which is not fully possible due to
current approach to merges for driver and vpl rt (which are merged after internal testing). Thus, there might be period of
time when driver and vpl rt builds will get broken right after merge to libva will happen.

Relates-to: oneapi-src/oneVPL-intel-gpu#237
Relates-to: intel/media-driver#1492

Signed-off-by: Cheng, Rachel <rachel.cheng@intel.com>

### DIFF
--- a/va/va_enc_av1.h
+++ b/va/va_enc_av1.h
@@ -249,7 +249,10 @@ typedef struct  _VAEncSequenceParameterBufferAV1 {
      */
     uint8_t     seq_tier;
 
-    uint8_t     reserved8b;
+    /** \brief Indicates whether or not the encoding is in dyadic hierarchical GOP structure.
+     *  value range [0..1].
+     */
+    uint8_t     hierarchical_flag;
 
     /** \brief Period between intra_only frames. */
     uint32_t    intra_period;
@@ -579,7 +582,11 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
      */
     uint8_t     ref_frame_idx[7];
 
-    uint8_t     reserved8bits0;
+    /** \brief When hierarchical_level_plus1 > 0, hierarchical_level_plus1-1 indicates
+     *  the current frame's level. If VAEncMiscParameterTemporalLayerStructure
+     *  is valid (number_of_layers >0), hierarchical_level_plus1 shouldn't larger than number_of_layers.
+     */
+    uint8_t     hierarchical_level_plus1;
 
     /** \brief primary reference frame.
      *  Index into reference_frames[]
@@ -595,6 +602,7 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
     /** \brief Corresponds to AV1 syntax element of the same name. */
     uint8_t     refresh_frame_flags;
 
+    /** \brief Reserved bytes for future use, must be zero. */
     uint8_t     reserved8bits1;
 
     /** \brief Suggest which frames to be used as references.
@@ -741,6 +749,7 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
         uint16_t    value;
     } qmatrix_flags;
 
+    /** \brief Reserved bytes for future use, must be zero. */
     uint16_t reserved16bits1;
 
     union {
@@ -797,6 +806,7 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
     /** \brief Number of tile rows. */
     uint8_t     tile_rows;
 
+    /** \brief Reserved bytes for future use, must be zero. */
     uint16_t    reserved16bits2;
 
     /** \brief The last tile column or row size needs to be derived. */
@@ -949,6 +959,7 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
      */
     uint8_t     number_skip_frames;
 
+    /** \brief Reserved bytes for future use, must be zero. */
     uint16_t    reserved16bits3;
 
     /** \brief Indicates the application forced frame size change in bytes.


Xuhenry1 (approved):
lgtm

dvrogozh (changes requested):
I am voting against this approach and suggest to return back to the API which we tried to introduce here and reverted later on: https://github.com/intel/libva/commit/97d750f45289867b1b75b947df1f181d9d7b6cd4

The reason I vote against is that doing as suggested in this PR we will introduce backward incompatibility in media-driver which relies on `reserved8` `reserved8bits0` fields. Relying on reserved is a wrong thing to do, no doubt. But that's not a reason to introduce backward incompatibility if we can avoid that.

The media-driver change which caused that is https://github.com/intel/media-driver/commit/4171a3e4fc8399fc3a372ae846ef6e9e25b2ac8d. It got included into intel-media-22.4.x and intel-media-22.3.x (thru cherry-pick) release series. I suggest that instead of merging in current PR, we return back to changing names of reserved fields used by media-driver. Surely, media-driver will need to use new definitions, plus it will need to fix 22.3.x and 22.4.x release series in the following way:
```
#if VA_CHECK_VERSION(1, 15, 0)
av1SeqParams->SeqFlags.fields.HierarchicalFlag = seqParams->HierarchicalFlag;
#else
av1SeqParams->SeqFlags.fields.HierarchicalFlag = seqParams->reserved8b;
#endif
```
That's assuming that the new field name got introduced in 1.15. Maybe we will need to update it to 1.16.

In this way we preserve backward compatibility and handle all build cases we have.

Xuhenry1 (approved):
LGTM

RachelChengRJ:
> I am voting against this approach and suggest to return back to the API which we tried to introduce here and reverted later on: [97d750f](https://github.com/intel/libva/commit/97d750f45289867b1b75b947df1f181d9d7b6cd4)
> 
> The reason I vote against is that doing as suggested in this PR we will introduce backward incompatibility in media-driver which relies on `reserved8` `reserved8bits0` fields. Relying on reserved is a wrong thing to do, no doubt. But that's not a reason to introduce backward incompatibility if we can avoid that.
> 
> The media-driver change which caused that is [intel/media-driver@4171a3e](https://github.com/intel/media-driver/commit/4171a3e4fc8399fc3a372ae846ef6e9e25b2ac8d). It got included into intel-media-22.4.x and intel-media-22.3.x (thru cherry-pick) release series. I suggest that instead of merging in current PR, we return back to changing names of reserved fields used by media-driver. Surely, media-driver will need to use new definitions, plus it will need to fix 22.3.x and 22.4.x release series in the following way:
> 
> ```
> #if VA_CHECK_VERSION(1, 15, 0)
> av1SeqParams->SeqFlags.fields.HierarchicalFlag = seqParams->HierarchicalFlag;
> #else
> av1SeqParams->SeqFlags.fields.HierarchicalFlag = seqParams->reserved8b;
> #endif
> ```
> 
> That's assuming that the new field name got introduced in 1.15. Maybe we will need to update it to 1.16.
> 
> In this way we preserve backward compatibility and handle all build cases we have.

Thanks Dmitry for the valuable comment, this pr has been changed accordingly, please help to review again, thanks very much!

dvrogozh (commented):
As we discussed, this will require immediate fix on media-driver and vpl rt sides. Do we have then already? Can you, please, provide draft PRs for these projects which we will promote on the moment this libva change will be merged?

RachelChengRJ:
> As we discussed, this will require immediate fix on media-driver and vpl rt sides. Do we have then already? Can you, please, provide draft PRs for these projects which we will promote on the moment this libva change will be merged?

Thanks. please check below prs:
msdk: https://github.com/intel-innersource/drivers.gpu.media.sdk.lib/pull/5927
driver:  https://github.com/intel-innersource/drivers.gpu.unified/pull/87490

XinfengZhang (approved):

RachelChengRJ:
> > As we discussed, this will require immediate fix on media-driver and vpl rt sides. Do we have then already? Can you, please, provide draft PRs for these projects which we will promote on the moment this libva change will be merged?
> 
> Thanks. please check below prs: msdk: [intel-innersource/drivers.gpu.media.vpl.public#3505](https://github.com/intel-innersource/drivers.gpu.media.vpl.public/pull/3505) driver: [intel-innersource/drivers.gpu.unified#87490](https://github.com/intel-innersource/drivers.gpu.unified/pull/87490)

VPL public repo's pr is attached, as for media driver, will start from inner source. will update comment after this change merged.

dvrogozh:
@RachelChengRJ : no, public repos are:
* https://github.com/oneapi-src/oneVPL-intel-gpu
* https://github.com/intel/media-driver

Your PRs are against intel internal repositories which should not be posted here at all.

RachelChengRJ:
Hi all, I've prepared below 2 PRs for open source VPL & Media driver, please help to review:
VPL: https://github.com/oneapi-src/oneVPL-intel-gpu/pull/237 
Media: https://github.com/intel/media-driver/pull/1492

thanks in advance,
Rachel

dvrogozh (approved):
looks good to me

va/va_enc_av1.h
```diff
@@ -249,7 +249,10 @@ typedef struct  _VAEncSequenceParameterBufferAV1 {
      */
     uint8_t     seq_tier;
 
-    uint8_t     reserved8b;
+    /** \brief Indicates whether or not the encoding is in dyadic hierarchical GOP structure.
+     *  value range [0..1].
+     */
+    uint8_t     hierarchicalFlag;
```
dvrogozh, line 255: Why it's named in camel case if everything else is not? This should be:
```
uint8_t hierarchical_flag;
```

dvrogozh (changes requested):
Well, sorry, just realized that member name style is wrong. Please, fix.

va/va_enc_av1.h
RachelChengRJ: > Why it's named in camel case if everything else is not? This should be:
> 
> ```
> uint8_t hierarchical_flag;
> ```

Thanks your carefully review, fixed.

XinfengZhang (approved):

dvrogozh (approved):
Looks good

rosetta-jpn:
I consider libva absolutely keeps backward compatibility.

However, this patch causes a backward-incompatibility. 
Why do we allow to do it in this patch?

XinfengZhang:
yes, libva absolutely keeps backward compatibility,
but this patch does not causes a backward incompatibility,
anyone should not use reserved byte and keep it as 0. 

with this patch , reserved byte was renamed to valid one. 
0 is telling driver there are no hierarchical information . ..

the usage is similar as using reserved byte in the tail of structure.

XinfengZhang:
potential risk: if application does not zero the structure, but using the field directly. sps->reserved8b = 0;
it will break the compilation , 
so, still should encourage  using the tail of the structure for future change

PR #613, opened 2022-07-07
Android fixes

evelikov:
Bunch of Android fixes while browsing through. Another stable release worthy material.

### DIFF
--- a/va/android/va_android.cpp
+++ b/va/android/va_android.cpp
@@ -28,7 +28,6 @@
 #include "va_backend.h"
 #include "va_internal.h"
 #include "va_trace.h"
-#include "va_fool.h"
 #include "va_android.h"
 #include "va_drmcommon.h"
 #include "va_drm_utils.h"
--- a/va/android/va_android.cpp
+++ b/va/android/va_android.cpp
@@ -43,33 +43,6 @@
 #define CHECK_SYMBOL(func) { if (!func) printf("func %s not found\n", #func); return VA_STATUS_ERROR_UNKNOWN; }
 #define DEVICE_NAME "/dev/dri/renderD128"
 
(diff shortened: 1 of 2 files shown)


dvrogozh (approved):
This does look good to me. However, are these blind changes or you had a chance to verify build and maybe test?

evelikov:
Build tested with hacked up the meson build.

va/android/va_android.cpp
```diff
@@ -28,14 +28,12 @@
...
 #include "va_internal.h"
 #include "va_trace.h"
-#include "va_fool.h"
 #include "va_android.h"
 #include "va_drmcommon.h"
 #include "va_drm_utils.h"
 #include <stdarg.h>
 #include <unistd.h>
 #include <sys/types.h>
-#include <sys/stat.h>
```
dvrogozh, line 38: Drop this include is the only thing which bother me a little. For example, https://manpages.ubuntu.com/manpages/xenial/man2/open.2.html:
```
       #include <sys/types.h>
       #include <sys/stat.h>
       #include <fcntl.h>
```

Though there are other mans over network which just list `#include <fcntl.h>`. Maybe safer to leave this include here? I am not sure how android includes would be structured.

evelikov: AFAICT there is lots of magic about these includes - both C runtime (glibc vs bionic) and version specific. So it's safer to keep it as-is. Pushed an update with the include preserved.

dvrogozh (approved):
looks good

xmao4-intel (approved):

TianmiChen (approved):

evelikov:
Happy one month anniversary \o/

Thanks for merging.

PR #340, opened 2019-10-09
drm: Add mapping of hantro kernel driver

kuhanh91:
This patch allows hantro kernel driver to be recognized
by va_getDriverName and for vaInitialize to select the
correct VA-API driver for hantro.

Signed-off-by: Murugasen Krishnan, Kuhanh <kuhanh.murugasen.krishnan@intel.com>

PR #615, opened 2022-07-08
Tweak the va driver-name mappings

evelikov:
- x11: drop the key_len from the mapping table
 - drm: drop the key_len from the mapping table
 - drm: fallback to drm driver-name, like we do in the x11/dri2 codebase
Thus we no longer need PRs like https://github.com/intel/libva/pull/340
 - drm: drop the no longer needed mappings \o/

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -48,27 +48,54 @@ static const struct driver_name_map g_driver_name_map[] = {
     { NULL,         NULL }
 };
 
+static char *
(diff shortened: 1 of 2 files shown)


dvrogozh (approved):
looks good

XinfengZhang (approved):

evelikov:
@XinfengZhang @dvrogozh I think we can close this PR: https://github.com/intel/libva/pull/340

XinfengZhang:
agree, now, default driver name (prefix) should be same as KMD

PR #616, opened 2022-07-12
Misc meson fixes

evelikov:
Nothing overly fancy here, small cleanup proper docs/doxygen handling and a fancy summary()

@ceyusa as the original author of the meson build, do you mind having a look?


meson.build
```diff
@@ -135,8 +135,8 @@ endif
 subdir('va')
 subdir('pkgconfig')
 
-doxygen = find_program('doxygen', required: false)
+doxygen = find_program('doxygen', required : get_option('enable_docs') == true)
```
eli-schwartz, line 151: If enable_docs is true, then this evaluates to `true == true`, which in turn evaluates to true

If enable_docs is false, then this evaluates to `false == true`, which in turn evaluates to false.

You can just use the option directly. :)

meson.build
```diff
@@ -140,3 +153,13 @@ doxygen = find_program('doxygen', required : get_option('enable_docs') == true)
 if get_option('enable_docs')
   subdir('doc')
 endif
+
+if meson.version().version_compare('>= 0.53')
```
eli-schwartz, line 157: The minimum meson version is already this.

eli-schwartz (changes requested):
Honestly, I'd use feature options more here.

meson.build
evelikov: Thanks for the help. Updated PR should have them addressed.

eli-schwartz:
There is also https://mesonbuild.com/Build-options.html#deprecated-options but then you'd have to bump the minimum meson to something very new. :(

eli-schwartz:
What would be the advantage of the synonym, other than the infrequent case that one migrates from a custom combo option that uses those strings?

va/meson.build
```diff
@@ -101,12 +101,12 @@ if WITH_DRM
...
 
   libva_drm_args = []
-  if get_option('with_x11') != 'no' and x11_dep.found()
+  if WITH_X11
     libva_drm_sources += [ 'drm/va_drm_auth_x11.c' ]
     libva_drm_args += [
       '-DLIBVA_MAJOR_VERSION=@0@'.format(libva_major_version)
     ]
-    deps += [ x11_dep ]
+    deps += [ x11_deps ]
```
ceyusa, line 109: I'm not sure if makes sens to link with all x11-related dependencies for drm

meson.build
```diff
@@ -127,8 +138,16 @@ endif
 subdir('va')
 subdir('pkgconfig')
 
-doxygen = find_program('doxygen', required: false)
+doxygen = find_program('doxygen', required : get_option('enable_docs'))
 
-if get_option('enable_docs') and doxygen.found()
+if get_option('enable_docs')
```
ceyusa, line 143: what if enable_docs is auto but doxygen is not installed?

ceyusa:
It's a nice clean up. An initial step to modernize the meson build.

meson.build
```diff
@@ -75,60 +75,50 @@ configinc = include_directories('.')
...
 
+# The libdrm package is required by all backends
 libdrm_dep = dependency('libdrm', version : '>= 2.4.60')
 
 backends = ''
-WITH_DRM = not get_option('disable_drm')
-if WITH_DRM
-  backends += 'drm'
+WITH_DRM = false
+if libdrm_dep.found() and not get_option('with_drm').disabled()
```
eli-schwartz, line 83: libdrm_dep cannot ever be not-found, so this one is a false choice

meson.build
eli-schwartz: This is currently broken and should be switched to use `if doxygen.found()`.

eli-schwartz (changes requested):

va/meson.build
evelikov: By default meson will pass `-Wl,--as-needed`, thus the linker will discard any unneeded DSOs.
Plus it makes little sense to have a separate dependency, just for this.

meson.build
evelikov: The enable_docs can never be auto - it's a bool.

evelikov: > This is currently broken and should be switched to use `if doxygen.found()`.

```
$ meson -D enable_docs=true

Program doxygen found: NO

meson.build:141:0: ERROR: Program 'doxygen' not found or not executable
```

Seems to be working as expected. Note the option is a bool and not a tristate (combo or feature).

meson.build
evelikov: Correct libdrm_dep.found is always true. I've added it for symmetry with the other code-paths and it's not causing any issues - be that confusing meson or adding any measurable overhead.

dvrogozh (changes requested):

va/meson.build
evelikov: I'm actually more tempted at removing the xlib dependency in the first place.

meson.build
```diff
@@ -81,11 +81,19 @@ WITH_DRM = not get_option('disable_drm')
 
 WITH_X11 = false
 if get_option('with_x11') != 'no'
-  x11_dep = dependency('x11', required : get_option('with_x11') == 'yes')
-  xext_dep = dependency('xext', required : get_option('with_x11') == 'yes')
-  xfixes_dep = dependency('xfixes', required : get_option('with_x11') == 'yes')
-
-  WITH_X11 = (x11_dep.found() and xext_dep.found() and xfixes_dep.found())
+  require_x11 =  get_option('with_x11') == 'yes'
```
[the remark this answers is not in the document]
evelikov: Having a simple foreach loop is idiomatic approach when using meson. Having it open-coded is repetitive and error-prone. Things will be even neater with the DRI3/Xwayland MR.

va/meson.build
evelikov: Dependency removed. This hunk is no longer relevant.

meson.build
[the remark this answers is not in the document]
evelikov: If you're interested at shorter code, then you'll love the last commit - using meson features. If you haven't used them before the [documentation](https://mesonbuild.com/Build-options.html#features) describes them well better than I could.

XinfengZhang (approved):

va/drm/va_drm_auth_x11.c
```diff
@@ -25,9 +25,16 @@
 #define _GNU_SOURCE 1
 #include "sysdeps.h"
 #include <dlfcn.h>
-#include <X11/Xlib.h>
+#include <stdint.h>
 #include "va_drm_auth_x11.h"
 
+/* A few ABI compabible defines and typedefs to avoid pulling Xlib.h */
```
dvrogozh, line 31: If we need Xlib.h, then we need to include it rather than mimic it. If we don't need to link against the library, then don't link, but add paths to include files - this should be managed on build system level.

dvrogozh (changes requested):

va/drm/va_drm_auth_x11.c
```diff
@@ -115,7 +122,7 @@ drm_auth_x11_init(DRMAuthX11 *auth)
     if (!auth->display)
         return false;
 
-    auth->window = DefaultRootWindow(auth->display);
+    auth->window = None;
```
dvrogozh, line 125: Why this is equivalent to what was here before?

va/drm/va_drm_auth_x11.c
evelikov: I would greatly encourage you to take a dive in the the Xlib.h headers. They are one of the worst one that exist. In practice this is the least crazy thing to do here.

va/drm/va_drm_auth_x11.c
evelikov: It is unused, so for all intents and purposes we can make it anything that we want.

evelikov:
@XinfengZhang any objections if we merge this for the next release? Obviously I would need to rebase it on top of the va-win32 addition that just landed.

evelikov:
Rebased the the PR with couple of small changes:
 - pushed the "use meson features" earlier in the series
 - make win32 a feature like the rest cc @sivileri

@XinfengZhang @dvrogozh as mentioned earlier, I would love it this can make it for 2.17.

evelikov:
@XinfengZhang @dvrogozh humble ping? Is there anything I can do to move this forward?

evelikov:
Bear in mind that the last commit will conflict, albeit trivially, with https://github.com/intel/libva/pull/656

evelikov:
@XinfengZhang @dvrogozh humble ping?

dvrogozh (changes requested):

va/drm/va_drm_auth_x11.c
evelikov: Sounds like throwing adjectives, instead of describing the details was a little misleading. Let me try again:

Various parts of X, predate standard C89/C99 and as such they often had to invent them to a varying degree. In some cases the X implementations would ignore/violate C standards be compatible on the API level but vary in ABI. Furthermore often the whole internal X ABI would be leaked and thus embedded into unsuspecting programs.

In this case here, we need a few very well defined and stable (from CPU arch or *nix flavour POV) types. We simply do not need the chaos that `#include <X11/Xlib.h>` brings.

For the same reasons, we've done similar changes in Mesa, so I believe we could and should really do the same in here. Does that make sense?

evelikov: @dvrogozh ping?

meson.build
```diff
@@ -75,44 +75,62 @@ configinc = include_directories('.')
 cc = meson.get_compiler('c')
 dl_dep = cc.find_library('dl', required : false)
 
-WITH_DRM = not get_option('disable_drm') and (host_machine.system() != 'windows')
-libdrm_dep = dependency('libdrm', version : '>= 2.4.60', required : (host_machine.system() != 'windows'))
+backends = ''
+
+WITH_DRM = false
+if not get_option('with_drm').disabled()
+  libdrm_dep = dependency('libdrm', version : '>= 2.4.60', required : (host_machine.system() != 'windows'))
```
xclaesse, line 82: If with_drm is set to auto, the libdrm dependency should be optional. Here you make it required on non-windows.

I would suggest doing instead:
```
libdrm_dep = dependency('libdrm', version : '>= 2.4.60', required : get_option('with_drm'))
WITH_DRM = libdrm_dep.found()
if WITH_DRM
  backends += ' drm'
endif
```

I don't the it matters to check the platform, if someone build with -Dwith_drm=enabled on Windows, they'll get an error because libdrm is not found and that's fine IMHO.

meson.build
```diff
@@ -75,44 +75,62 @@ configinc = include_directories('.')
...
+  x11_deps = [
+    dependency('x11', required : get_option('with_x11')),
+    dependency('xext', required : get_option('with_x11')),
+    dependency('xfixes', required : get_option('with_x11')),
+    dependency('x11-xcb', required : get_option('with_x11')),
+    dependency('xcb', required : get_option('with_x11')),
+    dependency('xcb-dri3', required : get_option('with_x11')),
+  ]
+  foreach d : x11_deps
+    WITH_X11 = WITH_X11 and d.found()
```
xclaesse, line 100: WITH_X11 is initially set to false, I think this will always be false then.

xclaesse:
Other than those small comments, the meson part looks good to me, it's a nice cleanup.

meson.build
evelikov: > If with_drm is set to auto, the libdrm dependency should be optional. 

Not quite: libdrm is a hard requirement for all non-win32 backends - android, x11 (dri3) and by extension glx, wayland (wl_drm) or drm itself - they all use the `va_drm_utils.c` helper which needs libdrm.

>I don't the it matters to check the platform, if someone build with -Dwith_drm=enabled on Windows, they'll get an error because libdrm is not found and that's fine IMHO.

There is a word missing above, it seems. Regardless - as long as the author @sivileri is fine it works for me.

meson.build
evelikov: Thanks fixed :+1:

evelikov:
@dvrogozh @XinfengZhang humble monthly ping?

meson.build
sivileri: For Windows it should still be fine, as long as nobody does `-Dwith_drm` without `libdrm`. I don't have any objections with making `libdrm` optional, as long as we are aware, as @evelikov points out, that doing the proposed change will make `libdrm` optional in all non-windows environments with have a hard dependency to it.

evelikov: Sounds like we're fine as-is them.

xclaesse: @evelikov I'm not libva expert, but it still feels wrong. With this PR, on linux, you can do `-Dwith_drm=disabled` and it will not use libdrm at all, so that dep is optional even on non-Windows platforms? If it makes sense to have libdrm disabled for a linux build, then `with_drm=auto` should allow libdrm dependency to be not-found. Only `-Dwith_drm=enabled` should force having libdrm on Linux IMHO. Otherwise a tristate enabled/disabled/auto does not make much sense, in your current implementation auto has the same effect as enabled. Unless there is something I missed?

xclaesse: Possibly with_drm should not be an option at all and we can remove the ` if not get_option('with_drm').disabled()` condition, to not allow disabling libdrm?

evelikov: Your wording threw me off there for a moment but you're spot on.

In particular `-D with_drm=disabled` is borked - the backend is correctly not selected for building, yet building the x11, wayland... ones fail, since meson sees the libdrm dependency as unresolved.

Will update in a moment to drop the explicit drm toggle - it makes zero sense to have it. Thanks for the suggestion and patience.

meson.build
```diff
@@ -75,50 +75,70 @@ configinc = include_directories('.')
...
+# NOTE: libdrm is required by all non-win32 backends x11 via dri3, glx via x11,
+# wayland via wl_drm and drm. All of these use va_drm_util.c which need libdrm.
+libdrm_dep = dependency('libdrm', version : '>= 2.4.60', required : (host_machine.system() != 'windows'))
 
-  x11_xcb_dep = dependency('x11-xcb', required : get_option('with_x11') == 'yes')
-  xcb_dep = dependency('xcb', required : get_option('with_x11') == 'yes')
-  xcb_dri3_dep = dependency('xcb-dri3', required : get_option('with_x11') == 'yes')
-  WITH_X11 = (WITH_X11 and x11_xcb_dep.found() and xcb_dep.found() and xcb_dri3_dep.found())
+WITH_X11 = true
+if not get_option('with_x11').disabled()
```
xclaesse, line 89: You have the opposite problem now, WITH_X11 is going to be true if with_x11=disabled AFAIK.

evelikov: thanks - third time's the charm :crossed_fingers:

va/drm/va_drm_auth_x11.c
evelikov: @XinfengZhang @dvrogozh humble poke

PR #617, opened 2022-07-12
Deprecate legacy emgd, nvctrl and fglrx codebase

evelikov:
With this MR we move all the emgd, nvctrl and fglrx code into the respective files.
Followed by a way to disable building it - disabled by default with meson, enabled (no switch) with autotools.

This code has not been relevant for distributions for a decade or so.

### DIFF
--- a/va/x11/va_fglrx.c
+++ b/va/x11/va_fglrx.c
@@ -23,6 +23,9 @@
  */
 
 #include "sysdeps.h"
+
+#ifdef HAVE_FGLRX
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -268,3 +271,5 @@ VAStatus va_FGLRX_GetDriverName(
 
     return VA_STATUS_SUCCESS;
 }
+
+#endif
--- a/va/x11/va_fglrx.c
+++ b/va/x11/va_fglrx.c
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <dlfcn.h>
 #include <X11/Xlib.h>
+#include "va_fglrx.h"
 
 #define ADL_OK 0
 #define ADL_MAX_PATH 256
@@ -121,9 +122,9 @@ static int match_display_name(Display *x11_dpy, const char *display_name)
     return m;
 }
 
-Bool VA_FGLRXGetClientDriverName(Display *dpy, int screen,
-                                 int *ddxDriverMajorVersion, int *ddxDriverMinorVersion,
-                                 int *ddxDriverPatchVersion, char **clientDriverName)
+static Bool VA_FGLRXGetClientDriverName(Display *dpy, int screen,
+                                        int *ddxDriverMajorVersion, int *ddxDriverMinorVersion,
+                                        int *ddxDriverPatchVersion, char **clientDriverName)
 {
     ADL_MAIN_CONTROL_CREATE          ADL_Main_Control_Create;
     ADL_MAIN_CONTROL_DESTROY         ADL_Main_Control_Destroy;
@@ -244,4 +245,26 @@ end:
     if (libadl_handle)
         dlclose(libadl_handle);
     return success;
+}
+
+VAStatus va_FGLRX_GetDriverName(
+    VADisplayContextP pDisplayContext,
+    char **driver_name,
+    int candidate_index
+)
+{
+    VADriverContextP ctx = pDisplayContext->pDriverContext;
+    int driver_major, driver_minor, driver_patch;
+    Bool result;
+
+    if (candidate_index != 0)
+        return VA_STATUS_ERROR_INVALID_PARAMETER;
+
+    result = VA_FGLRXGetClientDriverName(ctx->native_dpy, ctx->x11_screen,
+                                         &driver_major, &driver_minor,
+                                         &driver_patch, driver_name);
+    if (!result)
+        return VA_STATUS_ERROR_UNKNOWN;
+
+    return VA_STATUS_SUCCESS;
 }
--- a/configure.ac
+++ b/configure.ac
@@ -154,6 +154,21 @@ AC_ARG_ENABLE([va-messaging],
                     [build with va info and error messaging @<:@default=yes@:>@])],
     [], [enable_va_messaging="yes"])
 
+AC_ARG_WITH(legacy,
+    [AC_HELP_STRING([--with-legacy=[[components]]],
+                    [build with legacy components @<:@default=emgd,nvctrl,fglrx@:>@])],
+    [], [with_legacy="emgd,nvctrl,fglrx"])
+
+if test "$with_legacy" = *emgd*; then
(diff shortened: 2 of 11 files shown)


va/x11/va_fglrx.c
```diff
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <dlfcn.h>
 #include <X11/Xlib.h>
+#include "va_fglrx.h"
```
ceyusa, line 34: shall we remove this include in `va_x11.c`, perhaps also "va_nvctrl.h"

ceyusa: oh, no... nevermind.

XinfengZhang:
@xhaihao  , suppose you know more background of this? could you help to comments,  LGTM if no one is using it

evelikov:
@XinfengZhang what background are you interested in - I'm happy to provide.

EMGD is a former(?) Intel stack, which was never fully upstreamed - I believed it covered one of those Intel products paired with PowerVR GPU. The nvctrl is an Nvidia X extension (and library), which has never been relevant since there's no actual vaapi driver provided by them. While FGLRX is another X extension, this one for ATI GPUs - before AMD acquired ATI, back in 2006.

XinfengZhang:
thanks @evelikov , it is clear to me, if so , the PR LGTM

evelikov:
> thanks @evelikov , it is clear to me, if so , the PR LGTM

Glad to hear.

Rather noob question what does LGTM mean from a practical point of view - are we waiting for review from other people, who are we expecting to merge this, roughly at what point in time?

In general I've opened https://github.com/intel/libva/issues/622 since the contribution process to the project is really unclear to me.

dvrogozh (commented):
> AFAICT most distros are using meson

I am pretty sure that this is NOT the case for libva.

libva is a legacy long living project packages for which were prepared really long ago for major distributions and they were based on autoconf (since meson did not yet appear or was not widely adopted). Then, meson was introduced to libva, but autoconf was not wiped out. As a result, distros did not actually switched from autoconf to meson for libva which is pretty understandable: why change something which already works? For other projects the move from autoconf to meson was forced since autoconf was removed at some point. You can see that for yourself reviewing (mind no mentioning of meson and usage of aautoconf specific flags like ``--disable-wayland``):
1. https://salsa.debian.org/multimedia-team/libva/-/blob/master/debian/rules
2. https://src.fedoraproject.org/rpms/libva/blob/rawhide/f/libva.spec

You can compare with libdrm where you'll see meson usage:
1. https://salsa.debian.org/xorg-team/lib/libdrm/-/blob/debian-unstable/debian/rules
2. https://src.fedoraproject.org/rpms/libdrm/blob/rawhide/f/libdrm.spec

So, libva is kind of special case. My understanding is that Fedora and Debian are still on autoconf for libva. I did not check SUSE. And I believe that BSD is really using meson.

I don not think that meson build path is something which we are focusing on for libva. We are trying to keep it alive especially for BSD, but it's not a focus point. Autoconf is a focus.  In practice this means that meson path is not widely on radar. Take a look at CI, https://github.com/intel/libva/tree/master/.github/workflows, it's autoconf focused except for freebsd where we use meson.

evelikov:
I stand corrected - could swear I checked Fedora seemingly not. Arch has been using the meson build for at well over 2 years.

That said - not sure what tests are you looking for. The meson build (for people who want to use it) checks one of the permutations, autoconf the other. If people want to use autoconf + remove this code - they get to add the support. Alternatively they can use the meson build.

Personally I'm deliberately staying away from the autoconf build - is has many bugs and considering the rate at which PRs are merged, it will take well over 1 year to fix it. Not the amount of time I can spend - sorry.

evelikov:
Instead of "most distros" I really should have said "Arch and Gentoo" :sweat_smile:

dvrogozh (changes requested):
Summarizing my opinion on this PR.
1. I am totally fine with first 2 commits which are moving stuff for nv and fglrx drivers. It's good cleanup.
2. Last commit is a problem for me - for meson it's totally fine, but it lacks similar change to autotools build path. And I really don't want to diverged autotools/meson experience and create to parallel universes for libva developers and users.

@evelikov : unless you changed your mind to provide similar patch for autotools, I can suggest to decouple first 2 commits into separate PR which are all fine to be merged right away. For the build system part - I will be waiting for someone to do similar change to autotools.

evelikov:
I'm inclined to add autotools support on the premise that we actually deprecate and remove it in the near future. Will update the MR tomorrow + add a deprecation warning.

You hint shows zero interest in keeping it up-to date and adding that maintenance onto others is wrong on a number of levels.

evelikov:
As high-lighted in https://github.com/intel/libva/pull/655 - more distributions are opting for the meson build and it's only a matter of time for Debian, Ubuntu and OpenSuse to jump ship.

While I sympatise with the maintainers, it seems that in the mid/long run meson will be used across the board.

That said, I'll shortly respin this PR to wire-in the autotools build for the sake of completeness.

evelikov:
v2 adds the toggle in autotools, building the legacy components by default.

Thus ultimately preserving legacy behaviour for legacy users, while also testing both code-paths in the CI... Without adding yet another permutation in the matrix.

evelikov:
The style-check job caught some is small issues. Fixed.

evelikov:
@dvrogozh afaict the updated MR addresses all the points raised. Can you have another look? Thanks

dvrogozh (changes requested):
Update commit message for last commit: it's not only meson now and I think you implemented other behavior for autotools.

I am somewhat lost in behavior which was changed. Is this correc:
1. Autotools: introduced `--with-legacy=emgd,nvctrl,fglrx` option. Default build did not change.
2. meson: introduced `--enable-legacy=emgd|nvctrl|fglrx` option. Default build did change and above paths switched off by default?
This needs to be part of commit message and PR description.

By the way, I do not see that meson prints a summary. I thought you've added it?

evelikov:
Sorry about the with vs enable mix-up will fix with v2 (hopefully later today).

Wrt the missing meson summary: it's pending review in https://github.com/intel/libva/pull/616, where all the comments should be addressed AFAICT.

evelikov:
@dvrogozh PR updated

dvrogozh (approved):
looks good

XinfengZhang (approved):

mirh:
> The nvctrl is an Nvidia X extension (and library), which has never been relevant since there's no actual vaapi driver provided by them.

As I hinted in the previously referenced issue this assumption is wrong. There's the semi-official unmaintained-but-still-[hacked-up](https://github.com/xuanruiqi/vdpau-va-driver-vp9) original FD driver, and the [modern one](https://github.com/elFarto/nvidia-vaapi-driver) based on NVDEC. 
I originally thought you just had some trivial bug in the DRI/DRM code, but I just found an nvidia engineer [confirming](https://github.com/NVIDIA/open-gpu-kernel-modules/discussions/321#discussioncomment-3141883) the right (and only?) way to do checks under X11 is with NV-CONTROL.

PR #618, opened 2022-07-13
CI: upgrade FreeBSD version to avoid future breakage

jbeich:
FreeBSD Project currently doesn't keep recent binary packages for EOL versions. `/release_0` packages (non-default) frozen at 13.0 release (2021-04-13) will remain after 13.0 EOL (2022-08-31) for 13.* branch lifetime (until 2026-01-31) but maybe too old for libva CI.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -12,11 +12,11 @@ on:
 
 jobs:
   freebsd:
-    runs-on: macos-10.15
+    runs-on: macos-12
     steps:
     - uses: actions/checkout@v2
     - name: test
-      uses: vmactions/freebsd-vm@v0.1.5
+      uses: vmactions/freebsd-vm@v0
       with:
         prepare: |
           pkg install -y meson pkgconf libdrm libXext libXfixes wayland


jbeich:
Copypasta typo: s/epoll-shim/libva/

.github/workflows/freebsd.yml
```diff
@@ -12,11 +12,11 @@ on:
 
 jobs:
   freebsd:
-    runs-on: macos-10.15
+    runs-on: macos-12
     steps:
     - uses: actions/checkout@v2
     - name: test
-      uses: vmactions/freebsd-vm@v0.1.5
+      uses: vmactions/freebsd-vm@v0
```
dvrogozh, line 19: `v0` seems to be a moving tag currently pointing to v0.2.3, right?

dvrogozh (approved):

.github/workflows/freebsd.yml
dvrogozh: Found answer at https://github.com/vmactions/freebsd-vm:
```
The latest major version is: v0, which is the most recommended to use. (You can also use the latest full version: v0.2.3)
```

PR #621, opened 2022-08-26
va: Add Win32 Display/node and meson Windows build

sivileri:
This PR adds a new VA node for Win32 and adds support for Windows building with meson. 
- Some of the VA (mostly va.c and va_trace.c) runtime code was modified to build under Windows by adding the Win32 alternatives to platform specific libraries (ie. dlopen, dlclose, dlsym, pthreads, etc)
- Win32 Node highlights:
    -  vaGetDisplayWin32 receives LUID* to specify the desired adapter on Win32 platforms (NULL for adapter autoselection is allowed)
    - Surfaces can be imported/exported using either NT handles or ID3D12Resource object pointers
    - A default VA driver name (vaon12) will be provided for some scenarios like for example when no specific LUID is specified in vaGetDisplayWin32
    - When an adapter LUID is specified in vaGetDisplayWin32, the VA driver name will be dynamically loaded from a registry. If no entry is found in the registry for the specified adapter, the default VA driver will be used instead.

Related resources:
- https://github.com/intel/libva-utils/pull/283
- Samples using libva-win32 and D3D12 draft [HelloVAEncode](https://github.com/sivileri/DirectX-Graphics-Samples/tree/video_sample/Samples/Desktop/D3D12HelloWorld/src/HelloVAEncode) and [HelloVADecode](https://github.com/sivileri/DirectX-Graphics-Samples/tree/video_sample/Samples/Desktop/D3D12HelloWorld/src/HelloVADecode)
     - Adapter enumeration and selection with vaGetDisplayWin32
     - Importing existing D3D12 resources with vaCreateSurfaces, exporting to D3D12 with vaExportSurfaceHandle.
     - Usage of vaDeriveImage, vaAcquireBufferHandle, vaReleaseBufferHandle
     - HelloWorldVAEncode
         - VideoProcessing to alpha composite/scale different regions, RGB to YUV color conversion, render to screen and H264 encode
     - HelloWorldVADecode
         - Based on [llibva-utils sample](https://github.com/intel/libva-utils/blob/master/vendor/intel/sfcsample/DecodeParamBuffer.h)
         - Decodes H264 frame into NV12 texture, uses video processing to convert to RGB and renders to screen with D3D12
- Sample VA driver draft [here](https://gitlab.freedesktop.org/sivileri/mesa/-/commits/user/sivileri/va_win_build) `Acked-by: Emil Velikov <emil.l.velikov@gmail.com>`


evelikov:
Greetings, while I'm no maintainer here are bunch of high-level notes:
 - please split this up in meaningful commits - add abstraction X, Y, meson refactor, Windows support (code and build), CI?
 - move the _WIN32 abstraction to a standalone header, avoid abstracting the common POSIX/*nix use-case - use a local Windows abstraction for dlopen and friends
 - avoid duplication - seems like there's a bunch of that in the meson build
 - avoid using different naming scheme and/or location for .conf and drivers
 - be careful around meson dependencies - at a quick glance the new libdrm_dep handing seems off (will take another look once it's split)
 - add a new Windows CI step (ideally both mingw64 and msvc) to the CI - *nix hackers should not need to manually test the WIN32 code-paths and vice-versa


HTH o/

va/va_trace.c
```diff
@@ -855,6 +884,10 @@ void va_TraceEnd(VADisplay dpy)
     if (!pva_trace)
         return;
 
+#if defined(_WIN32)
+    DeleteCriticalSection(&pva_trace->resource_mutex);
+    DeleteCriticalSection(&pva_trace->context_mutex);
+#endif
```
XinfengZhang, line 890: looks no pthread_mutex_destroy called before?

sivileri: You're right. I'll add it now so there are no leaks on any of the platforms

sivileri:
Hi @evelikov , thank you for pointing this out. Please see inline below.

> Greetings, while I'm no maintainer here are bunch of high-level notes:
> 
> * please split this up in meaningful commits - add abstraction X, Y, meson refactor, Windows support (code and build), CI?

[sivileri] I went ahead and spitted the commits, please let me know if this breakdown makes more sense now.

> * move the _WIN32 abstraction to a standalone header, avoid abstracting the common POSIX/*nix use-case - use a local Windows abstraction for dlopen and friends

[sivileri] Do you you mean creating a header with the inline function and specific includes for windows and include that in main files like va.c instead of having it defined at the top of the C file directly?
For the POSIX/*unix specific things I've abstracted into common functions I did that to avoid branching with ifdefs in the bodies of different functions. Could you please give me an example of how a dlopen local Windows abstraction would look like?

> * avoid duplication - seems like there's a bunch of that in the meson build

[sivileri] Could you please point me to the duplicated part you mentioned? I found a small section that was duplicated but usually the win paths that are branched with ifs are not common to the existing code.

> * avoid using different naming scheme and/or location for .conf and drivers

[sivileri] Noted. Reverted the change for the .conf file to the usual path. For the VA drivers, we'd like to avoid having the suffix "drv_video.so" and have a dynamic loader that resolves the driver library name to load for each unique GPU/Adapter from the windows registry where other driver filenames are also usually stored. Does that make sense?

> * be careful around meson dependencies - at a quick glance the new libdrm_dep handing seems off (will take another look once it's split)

[sivileri] I think the libdrm dependency is correct now, it should only be mandatory on non-windows platforms and when get-option(disable-drm) is not set. 

> * add a new Windows CI step (ideally both mingw64 and msvc) to the CI - *nix hackers should not need to manually test the WIN32 code-paths and vice-versa

[sivileri] Added the CI for msvc and mingw on a separate commit.
> 
> HTH o/

va/meson.build
```diff
@@ -55,30 +55,43 @@ libva_headers_priv = [
   'va_trace.h',
 ]
 
-libva_sym = 'libva.syms'
-libva_sym_path = '@0@/@1@'.format(meson.current_source_dir(), libva_sym)
+if host_machine.system() == 'windows'
+  libva = static_library(
```
evelikov, line 59: Why are we creating a static library on Windows where all other platforms use a shared one?
AFAICT the project aims to provide some API/ABI guarantees and using a static library bypasses all of those.

Unless there is a very compelling reason, I would stick with a shared_library on Windows.

va/meson.build
```diff
@@ -269,6 +282,52 @@ if WITH_WAYLAND
...
+  # Make the main va target a static_library on Windows and have the win32 target generate the final DLL
+  libva_win32 = shared_library(
+    'va',
+    sources : libva_win32_sources +
+              libva_win32_headers,
+    soversion : libva_lt_current,
+    version : libva_lt_version,
+    install : true,
+    c_args : libva_win32_args,
+    vs_module_defs : 'libva.def',
```
evelikov, line 314: Building on the above (shared library) comment: libva should expose the core functionality, while libva-platform the platform specific one. As such the libva.def list and handling would need to be tweaked.

va/va.c
```diff
@@ -36,8 +36,35 @@
...
+inline char* strtok_impl(char *s, const char *delim, char **save_ptr) { return strtok_s(s, delim, save_ptr); }
+inline void* open_library(const char *file, int mode) { return LoadLibrary(file); }
+inline int close_library (void *handle) { return FreeLibrary(handle); }
+inline void *get_symbol_address (void *handle, const char *name) { return GetProcAddress(handle, name); }
+inline __uid_t geteuid() { return 0; };
+inline __uid_t getuid() { return 0; };
+#define ACCESS_OK 0
+#define DRIVER_EXTENSION    ".dll"
+#define DRIVER_PATH_STRING  "%s\\%s%s"
+#define ENV_VAR_SEPARATOR ";"
```
evelikov, line 53: Bit of an illustrative example on my earlier suggestion:
- create compat_win32.h or alike
- in there create a win32 "variant" of API respective APIs - `inline void *dlopen(...); inline char *strtok_r(...);....`
- use a `DRIVER_EXTENSION` similar to the *nix one - `#define DRIVER_EXTENSION "_drv_video.dll"

Keep the above a separate preparatory commit.

va/va_trace.c
```diff
@@ -44,16 +44,34 @@
...
 #include <string.h>
+#if defined(_WIN32)
+#include <windows.h>
+#include <io.h>
+#include <time.h>
+typedef CRITICAL_SECTION mutex_t;
+inline void mutex_lock(mutex_t* a) { EnterCriticalSection(a); }
+inline void mutex_unlock(mutex_t* a) { LeaveCriticalSection(a); }
+inline void mutex_init(mutex_t* a) { InitializeCriticalSection(a); }
+inline void mutex_destroy(mutex_t* a) { DeleteCriticalSection(a); }
```
evelikov, line 55: Similar to the `compat_win32.h` example above - these are better moved out, where a `pthread_foo` interface is provided or emulated.

The train of thought is that the onus of lacking standard API is on the platform, not on everyone else.

va/va_trace.c
```diff
@@ -538,6 +556,19 @@ static void FILE_NAME_SUFFIX(
     if (left < (size + 8 + 10))
         return;
 
+#if defined(_WIN32)
+    SYSTEMTIME st;
+    GetSystemTime(&st);
```
evelikov, line 561: The `compat_win32.h` should probably provide a `gettimeofday()` compatible implementation. Adding multiple WIN32 snippets through the code doesn't look nor scale well.

va/win32/Makefile.am
```diff
@@ -0,0 +1,23 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+ 
+# Extra clean files so that maintainer-clean removes *everything*
+MAINTAINERCLEANFILES = Makefile.in
```
evelikov, line 23: AFAICT the autoconf build isn't complete and this is a loose file which can be removed.
Technically one can add autoconf support, but I don't see it panning out in the long run - especially if there's no CI.

So let's remove this file.

va/win32/va_win32.c
```diff
@@ -0,0 +1,329 @@
...
+}
+
+static VAStatus va_DisplayContextGetNumCandidates(
+    VADisplayContextP pDisplayContext,
+    int *num_candidates
+)
+{
+    LUID* adapter = pDisplayContext->pDriverContext->native_dpy;
+    VADisplayContextWin32* pWin32Ctx = (VADisplayContextWin32*) pDisplayContext->opaque;
+    if(adapter && pWin32Ctx->registry_driver_available_flag)
```
evelikov, line 222: Can you please add a brief comment behind the magic numbers 2 and 1.

va/win32/va_win32.c
```diff
@@ -0,0 +1,329 @@
...
+)
+{
+    LUID* adapter = pDisplayContext->pDriverContext->native_dpy;
+    VADisplayContextWin32* pWin32Ctx = (VADisplayContextWin32*) pDisplayContext->opaque;
+    *driver_name = calloc(sizeof(pWin32Ctx->default_driver_name), sizeof(char));
+    if(adapter && pWin32Ctx->registry_driver_available_flag)
+    {
+        if(candidate_index == 0)
+            memcpy(*driver_name, pWin32Ctx->registry_driver_name, sizeof(pWin32Ctx->registry_driver_name));
+        if(candidate_index == 1)
```
evelikov, line 243: Similarly - a one like comment here behind the 0/1 index would be great.

va/win32/va_win32.c
```diff
@@ -0,0 +1,329 @@
...
+        return NULL;
+
+    pDisplayContext->vaIsValid       = va_DisplayContextIsValid;
+    pDisplayContext->vaDestroy       = va_DisplayContextDestroy;
+    pDisplayContext->vaGetDriverNameByIndex = va_DisplayContextGetDriverNameByIndex;
+    pDisplayContext->vaGetNumCandidates = va_DisplayContextGetNumCandidates;
+    pDisplayContext->opaque = calloc(1, sizeof(VADisplayContextWin32));
+
+    VADisplayContextWin32* pWin32Ctx = (VADisplayContextWin32*) pDisplayContext->opaque;
+    sprintf(pWin32Ctx->default_driver_name, "%s", "vaon12");
```
evelikov, line 273: Please document what "vaon12" is and why it makes sense to have it as default here.

va/win32/va_win32.c
```diff
@@ -0,0 +1,329 @@
...
+    unsigned short destw,
+    unsigned short desth,
+    VARectangle *cliprects, /* client supplied clip list */
+    unsigned int number_cliprects, /* number of clip rects in the clip list */
+    unsigned int flags /* de-interlacing flags */
+)
+{
+    return ctx->vtable->vaPutSurface(ctx, surface, (void *)draw, srcx, srcy, srcw, srch,
+                                     destx, desty, destw, desth,
+                                     cliprects, number_cliprects, flags);
```
evelikov, line 328: I would encourage you to implement vaExportSurfaceHandle and drop this PutSurface together.

The former should (implementation permitting) provide zero copy path, while does is always a copy. In addition, the API is defined only on va-x11 + the only project that I could find using it (libva-utils) simply does not work.

Aka it doesn't seem like a good choice from efficiency, portability and acceptance (aka actual users) POV.

va/win32/va_win32.h
```diff
@@ -0,0 +1,95 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct _VADisplayContextWin32 {
+    LUID adapter_luid;
+    char default_driver_name[MAX_PATH];
+    char registry_driver_name[MAX_PATH];
+    uint8_t registry_driver_available_flag;
+} VADisplayContextWin32;
```
evelikov, line 41: Used solely in `va_win32.c` so might as well move it in there.

va/win32/va_win32.h
```diff
@@ -0,0 +1,95 @@
...
+typedef struct _VAWin32SurfaceDescriptor {
+   /* Surface handle in the memory type specified */
+   void* handle;
+   /* Surface Width */
+   uint32_t width;
+   /* Surface Height */
+   uint32_t height;
+   /* Surface DXGI_FORMAT */
+   DXGI_FORMAT format;
+} VAWin32SurfaceDescriptor;
```
evelikov, line 64: Seemingly unused - remove to move next-to its user?

va/win32/va_win32.h
```diff
@@ -0,0 +1,95 @@
...
+
+/*
+ * Returns a suitable VADisplay for VA API
+ */
+VADisplay vaGetDisplayWin32(
+    /* Can be null for adapter autoselection in the VA driver */
+    const LUID* adapter_luid
+);
```
evelikov, line 72: This is a public header - so it should contain only this declaration + the respective includes (`va.h` for `VADisplay` and `windows.h`? for `LUID`).

pkgconfig/meson.build
```diff
@@ -2,49 +2,58 @@ pkg = import('pkgconfig')
...
-    filebase: 'libva-drm',
-    requires: 'libva',
-    variables: drm_vars,
+if WITH_WIN32
+  va_vars = vars + ['driverdir=' + driverdir] + ['display=win32']
+  pkg.generate(libva,
+    description: 'Userspace Video Acceleration (VA) Win32 interface',
+    filebase: 'libva',
+    variables: va_vars,
     version: va_api_version)
```
evelikov, line 11: If I'm reading this correctly, this provides only a `libva.pc` which contains `display=win32` where every else is a `libva.pc`  + `libva-$platform.pc` where the latter has `display=$platform`.

Unless there is a very compelling reason to deviate I would suggest using the pre-existing approach. This means that:
 - there will be no if/else spaghetti here
 - there will be less if/else spaghetti in the user-code - ffmpeg or others

meson.build
```diff
@@ -75,9 +75,8 @@ configinc = include_directories('.')
 cc = meson.get_compiler('c')
 dl_dep = cc.find_library('dl', required : false)
 
-libdrm_dep = dependency('libdrm', version : '>= 2.4.60')
-
-WITH_DRM = not get_option('disable_drm')
+WITH_DRM = not get_option('disable_drm') and (host_machine.system() != 'windows')
+libdrm_dep = dependency('libdrm', version : '>= 2.4.60', required : WITH_DRM)
```
evelikov, line 79: This one is off: libdrm_dep is required by all the existing backends - va-drm, va-x11 (+ by extension va-glx) and va-wayland.

meson_options.txt
```diff
@@ -3,5 +3,6 @@ option('disable_drm', type : 'boolean', value : false)
 option('with_x11', type : 'combo', choices : ['yes', 'no', 'auto'], value : 'auto')
 option('with_glx', type : 'combo', choices : ['yes', 'no', 'auto'], value : 'auto')
 option('with_wayland', type : 'combo', choices : ['yes', 'no', 'auto'], value : 'auto')
+option('with_win32', type : 'combo', choices : ['yes', 'no', 'auto'], value : 'auto')
 option('enable_docs', type : 'boolean', value : false)
-option('enable_va_messaging', type : 'boolean', value : true)
+option('enable_va_messaging', type : 'boolean', value : true)
```
evelikov, line 8: Please tweak your editor settings and/or git config so these "No newline at end of file" are resolved.

meson.build
```diff
@@ -120,8 +119,14 @@ if cc.has_argument('-fvisibility=hidden')
   va_c_args += ['-DHAVE_GNUC_VISIBILITY_ATTRIBUTE']
 endif
 
-if (not WITH_DRM and not WITH_X11 and not WITH_WAYLAND)
-  error('Please install at least one backend dev files (DRM, X11, Wayland)')
+WITH_WIN32 = false
+libwin32_dep = []
+if get_option('with_win32') != 'no'
+    WITH_WIN32 = (host_machine.system() == 'windows')
+endif
```
evelikov, line 126: I must be missing something here:
Is this the 3rd patch in the MR - as such one cannot build the project with just the first two commits?

That sounds odd to me - might as well squash this with the first patch in the PR.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  push:
+    paths-ignore:
+    - '.github/workflows/**'
+    - '!.github/workflows/windows.yml'
+  pull_request:
+    paths-ignore:
+    - '.github/workflows/**'
+    - '!.github/workflows/windows.yml'
```
evelikov, line 11: TIL paths-ignore, thank you :100: 

Out of curiosity: shouldn't `EnterDevShell.ps1` be something provided by the WIndows/MSVC SDK/DDKs or alternatively the Windows container/image provided by Github workflow?

The CI coverage looks great, thanks o/

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  windows-msvc:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
```
evelikov, line 18: Nit: copy/past the step names (like mingw below).

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install Meson and Ninja
+      run: pip install meson ninja
+    - name: Enter DevShell
+      run: '.github\workflows\EnterDevShell.ps1 ${{ inputs.architecture }}'
+      shell: pwsh
+    - name: Configure with meson for WIN32 backend only
+      run: meson build -Ddisable_drm=true -Dwith_x11=no -Dwith_glx=no -Dwith_wayland=no -Dwith_win32=yes
```
evelikov, line 27: Nit: drop the explicit drm/x11/glx/wayland here and below.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+jobs:
+  windows-msvc:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install Meson and Ninja
+      run: pip install meson ninja
```
evelikov, line 22: ninja is a c++ project, so it should not be available in pip. My guess is that it's available in the default image? if so, it can be also dropped in the mingw case below

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+    steps:
+    - name: 'Checkout'
+      uses: actions/checkout@v2
+    - name: 'Setup MSYS2'
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: mingw64
+        update: false
+        install: >-
+          git
```
evelikov, line 45: Nit: is git needed here or is this a left-over from earlier debugging?

meson.build
evelikov: Got confused there by the UI - this is the second commit, so it's fine as-is. Although it might be worth moving the CI commit to the end of the PR.

va/win32/va_win32.h
sivileri: Not used by the VA runtime itself, but by API clients/drivers (similar to `VADRMPRIMESurfaceDescriptor`)

va/win32/va_win32.h
sivileri: In addition to the ones you mention, I added va_backend.h for `VADriverContextP`, dxgiformat.h for `DXGI_FORMAT`

va/win32/va_win32.c
sivileri: Thanks for pointing this out. Sounds better/simpler to me too that apps use vaExportSurfaceHandle (ideally with no copies) and implement their own/defer to other present codepaths. If for some reason in the future a customer wants vaPutSurface for a good reason, we can extend the win32 API at that point.

va/va.c
sivileri: Sounds good, thanks for providing the example. Will look like a less intrusive code change to common code done this way with the win32 compat header. For the DRIVER_EXTENSION, I might be missing out on the historical reasons for the "_drv_video" suffix, but if possible, we'd like to have more flexibility (ie. no name suffix) in the driver name, especially for dynamically loaded driver names (ie. map LUID to driver name from the registry) that could follow different conventions/case naming for each vendor.

.github/workflows/windows.yml
sivileri: Unfortunately the install step for meson/ninja is needed in both msvc/mingw

meson.build
sivileri: Thanks for pointing this out. Now I've made libdrm not required only for Windows platforms

dvrogozh:
@sivileri Can you, please, clarify the intent: in which applications or services this will be used? Overall there are the following questions which arise:
1. Who will provide a libva driver? at which timeframe? will it be open sourced? (you mention something about veon12)
2. What kind of driver that will be? (I am guessing that will be a proxy over dx12 real driver, right?)
3. What are applications or services which will be using libva on Windows? any open source applications?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
+name: windows
+
+on:
+  push:
+    paths-ignore:
+    - '.github/workflows/**'
+    - '!.github/workflows/windows.yml'
```
dvrogozh, line 7: add:
```
- '!.github/workflows/EnterDevShell.ps1'
```

That's `paths-ignore`, idea is that if there is change only to other .yml workflows, avoid running this one. But you want to run it if there is a change to EnterDevShell.ps1.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+    - '.github/workflows/**'
+    - '!.github/workflows/windows.yml'
+  pull_request:
+    paths-ignore:
+    - '.github/workflows/**'
+    - '!.github/workflows/windows.yml'
+
+jobs:
+  windows-msvc:
+    runs-on: windows-latest
```
dvrogozh, line 15: Change `windows-latest` to explicit windows version. We don't need sudden break in the ci due to version bump up which we don't track.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Enter DevShell
+      run: '.github\workflows\EnterDevShell.ps1 ${{ inputs.architecture }}'
+      shell: pwsh
+    - name: Configure with meson for WIN32 backend only
+      run: meson build
+    - name: Build and Install with ninja
+      run: ninja -C build install
+
+  windows-mingw:
+    runs-on: windows-latest
```
dvrogozh, line 34: same

dvrogozh (changes requested):

va/va.c
evelikov: Many things on *nix follow naming patterns. Don't know if there's a particular reason for "drv_video" over any other, but as you have seen with the mesa vaon12 - the suffix is there. Unless there is a compelling reason, we should not.

Vendors, assuming we get any, still have flexibility - they can name the driver "i-like_my-bikeShed_InPurple_drv_video.dll" if they choose to do so.

va/compat_win32.h
```diff
@@ -0,0 +1,116 @@
...
+    return strtok_s(s, delim, save_ptr);
+}
+#endif
+
+inline void* dlopen(const char *file, int mode)
+{
+    return LoadLibrary(file);
+}
+
+inline int dlclose (void *handle)
```
evelikov, line 52: While the CONTRIBUTING.md does not state codying style, we do have "style-unify" in-tree. Please give that a try and squash with respective commits... Don't bother with changes to any other files.

va/compat_win32.h
```diff
@@ -0,0 +1,116 @@
...
+{
+    return FreeLibrary(handle);
+}
+
+inline void *dlsym (void *handle, const char *name)
+{
+    return GetProcAddress(handle, name);
+}
+
+static char last_err_string[512];
```
evelikov, line 62: Scope this within its user. The lifespan will be the same, without making the variable part of the API.

va/libva.def
```diff
@@ -0,0 +1,74 @@
...
+    vaUnlockSurface
+    vaCreateMFContext
+    vaMFAddContext
+    vaMFReleaseContext
+    vaMFSubmit
+    vaCreateBuffer2
+    vaQueryProcessingRate
+    vaExportSurfaceHandle
+    va_newDisplayContext
+    va_newDriverContext
```
evelikov, line 74: Pretty sure that "va_new*" are not supposed to be here. Please double-check against the public header `va.h`

va/va.c
```diff
@@ -451,8 +459,10 @@ static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
         va_infoMessage(dpy, "Trying to open %s\n", driver_path);
 #if defined(RTLD_NODELETE) && !defined(ANDROID)
         handle = dlopen(driver_path, RTLD_NOW | RTLD_GLOBAL | RTLD_NODELETE);
-#else
+#elif defined(RTLD_NOW) && defined(RTLD_GLOBAL)
         handle = dlopen(driver_path, RTLD_NOW | RTLD_GLOBAL);
+#else
+    handle = dlopen(driver_path, 0);
```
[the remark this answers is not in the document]
evelikov: Better yet - provide dummy defines in the `compat_win32.h` header and drop this extra "else"

va/win32/va_win32.c
```diff
@@ -0,0 +1,323 @@
...
+                fprintf(stderr, "VA_Win32: D3DKMTEnumAdapters2 status != SUCCESS\n");
+                goto out;
+            }
+            break;
+        }
+        EnumAdapters.adapters = malloc(sizeof(*EnumAdapters.adapters)*(EnumAdapters.adapter_count));
+        if (EnumAdapters.adapters == NULL)
+        {
+            fprintf(stderr, "VA_Win32: Allocation failure for adapters buffer\n");
+            goto out;
```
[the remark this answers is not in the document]
evelikov: Neither of these `return false` suggestions seem wise - they will cause a handle leak.

va/win32/va_win32.c
```diff
@@ -0,0 +1,323 @@
...
+
+    if (pDisplayContext->pDriverContext
+        && pDisplayContext->pDriverContext->native_dpy)
+        free(pDisplayContext->pDriverContext->native_dpy);
+
+    if (pDisplayContext->pDriverContext)
+        free(pDisplayContext->pDriverContext);
+
+    VADisplayContextWin32* pWin32Ctx = (VADisplayContextWin32*) pDisplayContext->opaque;
+    if (pWin32Ctx)
```
evelikov, line 216: Drop the null checks - `free(NULL)` is fine and the casting of `opaque`

va/win32/va_win32.c
```diff
@@ -0,0 +1,323 @@
...
+    pDisplayContext->vaIsValid       = va_DisplayContextIsValid;
+    pDisplayContext->vaDestroy       = va_DisplayContextDestroy;
+    pDisplayContext->vaGetDriverNameByIndex = va_DisplayContextGetDriverNameByIndex;
+    pDisplayContext->vaGetNumCandidates = va_DisplayContextGetNumCandidates;
+    pDisplayContext->opaque = calloc(1, sizeof(VADisplayContextWin32));
+
+    VADisplayContextWin32* pWin32Ctx = (VADisplayContextWin32*) pDisplayContext->opaque;
+
+    /*  Initialize default_driver_name to the DX12 VA driver implementation
+        which will be provided when the adapter LUID doesn't have an associated VA driver name
```
evelikov, line 287: Please use consistent comment format - the license blurb one is seems the defacto standard.

va/win32/va_win32.h
```diff
@@ -0,0 +1,71 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#ifndef _VA_WIN32_H_
+#define _VA_WIN32_H_
+
+#include <va/va.h>
+#include <va/va_backend.h>
```
evelikov, line 28: This is a frontend header, no backend specifics must be included here.

va/win32/va_win32.c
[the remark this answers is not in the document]
dvrogozh: Ok, I forgot `LoadLibrary`. Ignore then, let's use `goto`,

va/meson.build
```diff
@@ -269,6 +270,50 @@ if WITH_WAYLAND
...
+    vs_module_defs : 'libva_win32.def',
+    dependencies : deps)
+
+  libva_win32_dep = declare_dependency(
+    link_with : libva_win32,
+    include_directories : configinc,
+    dependencies : deps)
+
+  if fs.is_file(libva.full_path()) and fs.is_file(libva_win32.full_path())
+    install_data(
```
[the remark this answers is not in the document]
sivileri: This is to remove the -<version> from the dlls so they are installed as va.dll and va_win32.dll. I don't think meson has a way to directly rename the install name without doing this.

va/compat_win32.h
```diff
@@ -0,0 +1,116 @@
...
+#include <windows.h>
+#include <winsock.h>
+#include <io.h>
+#include <time.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define F_OK 0
```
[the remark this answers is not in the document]
sivileri: It's used in va.c

va/libva.def
sivileri: I added these two to be exported in va.dll so they can be linked/called from va_win32.dll

va/compat_win32.h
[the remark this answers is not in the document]
dvrogozh: Ah, in a code which was there before and did not show up in github diff, here:
https://github.com/intel/libva/blob/f5639d3916a52d34eb0584ec2789d8496f540e7c/va/va.c#L459

va/win32/va_win32_adapter.h
```diff
@@ -0,0 +1,130 @@
...
+#endif
+
+#ifndef NTSTATUS
+typedef LONG NTSTATUS;
+#define STATUS_SUCCESS  ((NTSTATUS)0x00000000L)
+#define STATUS_BUFFER_TOO_SMALL ((NTSTATUS)0xC0000023)
+#define NT_SUCCESS(status) (((NTSTATUS)(status)) >= 0)
+#endif
+
+typedef struct LoaderEnumAdapters2 {
```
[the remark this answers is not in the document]
sivileri: They come from the WDK d3dkmthk.h, d3dukmdt.h. I've added them here for compat with the mingw CI that doesn't seem to be picking them up (They're installed in the CI machine as MSVC works fine with them)

va/libva.def
evelikov: Right, the libva.so ABI is a mess. I should probably stop complaining and send some patches instead :sweat_smile: 

Please ignore my suggestion.

.github/workflows/windows.yml
evelikov: meson is one thing ninja is another.

`pip install ninja` does not work here, which is to be expected since it's not a python module. Have you tried changing the above to "pip install meson" and dropping "ninja" from the mingw/pacboy section - what kind of error do you see in the CI?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,58 @@
...
+          toolchain:p
+          meson:p
+          ninja:p
+    - name: Enter DevShell
+      run: '.github\workflows\EnterDevShell.ps1 ${{ inputs.architecture }}'
+      shell: pwsh
+    - name: Configure with meson for WIN32 backend only
+      run: CC=gcc CXX=g++ meson build
+    - name: Build and Install with ninja
+      run: CC=gcc CXX=g++ ninja -C build install
```
evelikov, line 58: CXX is unused and can be dropped. There's no C++ code in tree (ignoring the android one) even if the configure.ac lists `AC_PROG_CXX` - it should be removed ... Separate MR if you feel like fixing it - thanks o/

va/win32/va_win32.c
```diff
@@ -0,0 +1,310 @@
...
+    pDriverContext = va_newDriverContext(pDisplayContext);
+    if (!pDriverContext) {
+        va_DisplayContextDestroy(pDisplayContext);
+        return NULL;
+    }
+
+    pDriverContext->display_type = VA_DISPLAY_WIN32;
+
+    if (adapter_luid) {
+        // Copy LUID information to driver context
```
evelikov, line 304: Nit: we still have a zoo of comment styles

.github/workflows/windows.yml
sivileri: You're correct, I misread the first comment and removed both meson/ninja and it failed. But just leaving meson works correctly, did that change now.

sivileri:
@dvrogozh 
> @sivileri Can you, please, clarify the intent: in which applications or services this will be used? Overall there are the following questions which arise:
> 
> 1. Who will provide a libva driver? at which timeframe? will it be open sourced? (you mention something about veon12)
> 2. What kind of driver that will be? (I am guessing that will be a proxy over dx12 real driver, right?)

[sivileri] This change provides the option for IHVs to write and register VA drivers for their own adapters. When vaGetDisplayWin32 is provided with a LUID* for an adapter, the registered driver name is resolved (ie. va_win32.c/TryLoadDriverNameFromRegistry). When a registered driver/LUID is not available, we will provide an open source VA driver based on D3D12 (ie. vaon12), which is being developed and will map commands from the VA driver interface into D3D12 (and finally go through the D3D12 driver). 

> 3. What are applications or services which will be using libva on Windows? any open source applications?

[sivileri] From the applications/services side, the intent is to add new options for media APIs for ISVs and developers in the community. Any application will have the option to initialize with `vaGetDisplayWin32` and perform generic VA workloads, having interop/import/export capabilities with textures/buffers using `vaExportSurfaceHandle`/`vaCreateSurfaces`/`vaAcquireBufferHandle`/`vaReleaseBufferHandle` with either  `ID3D12Resource` or `HANDLE` memory types.

va/va_trace.c
```diff
@@ -1140,6 +1141,14 @@ static void va_TraceSurfaceAttributes(
     if (!attrib_list || !num_attribs)
         return;
 
+    int32_t memtype = 0;
+    for (i = 0; i < *num_attribs; i++) {
+        if(attrib_list[i].type == VASurfaceAttribMemoryType) {
+            memtype = attrib_list[i].value.value.i;
```
evelikov, line 1147: Overall this looks like a good fix, which should be a separate MR.

Few small suggestions:
 - seems like the memtype assignment should happen in the loop below, where it's being checked
 - why do we handle mem_type 0, the same as TYPE_DRM_PRIME - that doesn't seem right
 - can we bribe you to add support for DRM_PRIME_2 - thanks o/

sivileri: Thanks. I'd prefer if possible, to keep the fix in this MR, as a separate commit, which I think will be rebase-merged into master, keeping history clean too eventually. By having the fix commit in this MR, I can also then build upon it and add the win32 buffer types in a subsequent commit, and I can avoid the crash when I test `vaExportSurfaceHandle`. As this MR targets `va_win` branch, it'd be easier to avoid rebasing from multiple branches once the fix goes into master in a separate PR.

Regarding your other points I've added some comment in the code to clarify:
1) I'm preferred iterating over the attributes first, in case the API client sends `VASurfaceAttribMemoryType`  before `VASurfaceAttribExternalBufferDescriptor` in `attrib_list` as I didn't find any order specification in the docs, does this make sense ?
2) I found a note in va.h stating that `VASurfaceAttribExternalBuffers` type should be assumed for `VASurfaceAttribExternalBufferDescriptor` unless the memory type is provided.
3) I've added support for printing `VADRMPRIMESurfaceDescriptor` in the commit along with the fix, and then added a new commit with win32 types trace support too.

sivileri:
@evelikov @dvrogozh I'm seeing a failure in the freebsd build in the last couple of days but doesn't seem to be related to the code itself. For example, [the latest pipeline](https://github.com/intel/libva/runs/8272559848?check_suite_focus=true) seems to be failing around provisioning the VM and installing the packages defined in freebsd.yml (ie. pkgconf libdrm libXext libXfixes wayland and '^mesa($|-libs)')

```
Installing pkg-1.18.3...
Newer FreeBSD version for package pkg:
To ignore this error set IGNORE_OSVERSION=yes
- package: 1301000
- running kernel: 1300139
Ignore the mismatch and continue? [y/N]: 
Failed to install the following 1 package(s): /tmp//pkg.txz.f0CPd7
Bootstrapping pkg from pkg+http://pkg.freebsd.org/FreeBSD:13:amd64/quarterly, please wait...
Verifying signature with trusted certificate pkg.freebsd.org.2013102301... done
Error: The process '/usr/bin/ssh' failed with exit code 1
```

The url from where the pkg is being bootstrapped http://pkg.freebsd.org/FreeBSD:13:amd64/quarterly/ was updated two days ago. Could this be related to https://github.com/intel/libva/pull/618 ?

dvrogozh:
Well, the only thing I can tell right now is that you are incredibly lucky considering that #625 which I merged just few minutes before you rebased and hit this issue got built fine on freebsd :). We have issues w/ freebsd builds from time to time. We will handle this separately.

sivileri:
> Well, the only thing I can tell right now is that you are incredibly lucky considering that #625 which I merged just few minutes before you rebased and hit this issue got built fine on freebsd :). ~We have issues w/ freebsd builds from time to time. We will handle this separately.~
> 
> UPDATE; @sivileri : you did mistake in git-rebase:
> 
> ```
> $ git fetch origin pull/621/head:pr_621
> $ git checkout pr_621
> $ git branch -u origin/master
> Branch pr_621 set up to track remote branch master from origin.
> $ git status
> # On branch pr_621
> # Your branch and 'origin/master' have diverged,
> # and have 7 and 14 different commits each, respectively.
> #   (use "git pull" to merge the remote branch into yours)
> #
> nothing to commit, working directory clean
> ```
> 
> As a result you have outdated ci config for freebsd.yml which points to macos-10.5 instead of macos-12 and likely miss some other fixes. Please, rebase your PR on top of the latest master.

Thanks @dvrogozh. Rebased now to latest `intel:master` and also changed the PR to target `intel:master`. CC @XinfengZhang this PR now targets intel:master and no longer the intel:va_win branch you created a couple months ago.

dvrogozh:
> CC @XinfengZhang this PR now targets intel:master and no longer the intel:va_win branch you created a couple months ago.

Looks like you folks had some discussions/arrangements which I am not aware of. Will talk to @XinfengZhang offline to get up to speed.

va/va_trace.c
evelikov: Makes sense, thanks for going the extra mile with DRM_PRIME_2 :heart:

evelikov:
I would encourage maintainers to pick this for the next release. There has been a driver in mesa for a number of months now, props to @sivileri et al. So this PR allows end-users to have upstream stack.

Similar to the DRI3/Xwayland PR, it'll be nice to have this in master for at least a week before actual release is made. HTH o/

dvrogozh:
While I am fully positive of the PR, I would like to understand that we have open source driver and open source application(s) which are actually using libva on Windows. This will significantly encourage the merge in. Saying that "we have" above 2 things I am talking about real project already hosted on github or gitlab. For the application, I personally would be fine seeing functional vainfo and couple of basic test/sample apps for decoding, encoding and video processing.

> There has been a driver in mesa for a number of months now

I do not follow mesa development and not aware of that. Provide a link, please. If this can be used as a justification for driver part - this would be wonderful.

sivileri:
@evelikov @dvrogozh (cc @XinfengZhang)

Along with this pull request, there are projects hosted for [an open source driver being developed for Windows](https://gitlab.freedesktop.org/sivileri/mesa/-/commits/va_win_testing/) as well as [an open source sample app using libva on Windows](https://github.com/sivileri/DirectX-Graphics-Samples/tree/video_sample/Samples/Desktop/D3D12HelloWorld/src/HelloVAEncode) and [ported vainfo ](https://github.com/sivileri/libva-utils/tree/va_win_dev). As all those [depend on libva-win32](https://gitlab.freedesktop.org/sivileri/mesa/-/blob/41f2bad2c2f42adb2a33d9ab682a81a6ebeedd3d/meson.build#L832), I posted this pull request first. Once the libva runtime libva-win32 support is released/merged, the driver and the apps with the libva-win32 dependency can move forward.

dvrogozh:
@sivileri : you should have been brought this up front with the references in PR description. Not seeing that part and not knowing it even exists was the major reason I stopped actively reviewing your PR at some point. I suggest you should go ahead and submit draft PRs for the named projects referencing the libva PR. Something like:
```
Requires: https://github.com/intel/libva/pull/621
```
At least for libva-utils, please, do so - submit draft PR. For the mesa driver can you, please, receive "Acked-by:" from mesa maintainers which would signify that they are OK to accept the contribution to mesa project (eventually after cleanup, reviews, etc.)?

As for this PR, please, update PR description with the link to libva-utils draft PR, mesa draft PR (if you will submit it) and Acked-by from mesa?

evelikov:
Seems like a typical chicken and egg problem :egg:

Although the earlier links to the driver in mesa plus a casual search might have helped https://gitlab.freedesktop.org/mesa/mesa/-/commits/main?author=Sil%20Vilerino /me thinks :stuck_out_tongue_winking_eye:

evelikov:
@dvrogozh @sivileri with my mesa dev hat on (in case 2.5k commits count for something) the mesa va/win32 changes are preemptively:
```
Acked-by: Emil Velikov <emil.l.velikov@gmail.com>
```

va/win32/va_win32.c
```diff
@@ -0,0 +1,310 @@
...
+const char VAAPI_DEFAULT_DRIVER_NAME[] = "vaon12";
+const char VAAPI_REG_SUB_KEY[] = "VAAPIDriverName";
+const char VAAPI_REG_SUB_KEY_WOW[] = "VAAPIDriverNameWow";
+
+inline const char* GetVAAPIRegKeyName()
+{
+#ifdef _WIN64
+    return VAAPI_REG_SUB_KEY;
+#else
+    BOOL is_wow64;
```
dvrogozh, line 46: Please, initialize variable.
```
    BOOL is_wow64 = FALSE;
```

va/win32/va_win32.h
```diff
@@ -0,0 +1,70 @@
...
+#endif
+
+/** \brief HANDLE memory type
+ * 
+ * The HANDLE type can be used with D3D12 sharing resource functions such as:
+ * https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-opensharedhandle
+ * https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-createsharedhandle
+ *
+ * When used with VASurfaceAttribExternalBufferDescriptor for vaCreateSurfaces(...,num_surfaces,...),
+ * the associated VAGenericValueTypePointer is of type HANDLE[num_surfaces]
```
dvrogozh, line 41: I believe that other kinds of external surfaces follow the pattern that there is single handle in the attribute and functions should be called as `vaCreateSurfaces(...,1,...)`. Do we actually need to step out from this pattern for Windows?

sivileri: I think this is true for `VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2`, however for `VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME` when the passed structure is `VASurfaceAttribExternalBuffers`, there are two params `buffers` and `num_buffers` for passing individual handles for each of the `num_surfaces` to be created.

I found two driver implementations that implement vaCreateSurfaces**2** like that (not to confuse with the older vaCreateSurfaces function signature that doesn't receive the attribs as params). 

- https://github.com/intel/media-driver/blob/1faefc4110cb701b88a4942804b31ab19f3e1a3a/media_driver/linux/common/ddi/media_libva.cpp#L2918
    - Which I think ends up opening the each handle in https://github.com/intel/media-driver/blob/69235cbdfa375aefeded9afe1cc86b826ebbcf55/media_driver/linux/common/ddi/media_libva_util.cpp#L350

- https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/gallium/frontends/va/surface.c#L1024

I think it's useful to be able to import N handles at once into N created surfaces in Windows following the existing pattern for `VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME`.

evelikov: Personally I believe that having the `VA_SURFACE_ATTRIB_MEM_TYPE_` spread out across headers is a bad idea. Although I also think that we should adapt the vaGetDisplay interface after the EGL_PLATFORM extensions, since basically that's what the libva design was after.

Regardless - these are orthogonal to the feature added here and can be looked into at a later date.

dvrogozh: Ok, till that's documented (and it is for win surfaces) I am fine with that. Leave comment unresolved, please, for other people to provide feedback.

dvrogozh (approved):
Looks good to me.

sivileri:
Thanks for approving @dvrogozh , could you please (or any other maintainer) merge this to master? I don't seem to have permission to do so myself.

image

sivileri:
Exported two more functions in libva.def, that were used by the google tests in https://github.com/intel/libva-utils/pull/283, also ported the tests to windows in addition to vainfo now.

vaMaxNumConfigAttributes
vaBufferTypeStr

evelikov:
Just compared the va.def against the exported symbols on my Arch machine, here is the diff with some inline notes:

```diff
 vaAcquireBufferHandle
-VA_API_0.32.0 ## version tag, POSIX specific
-VA_API_0.33.0 ## version tag, POSIX specific
 vaAssociateSubpicture
-vaAttachProtectedSession ## va_prot.h API, should be added for win32?
 vaBeginPicture
 vaBufferInfo
 vaBufferSetNumElements
 vaBufferTypeStr
 vaConfigAttribTypeStr
-vaCopy ## Maybe not great from efficiency POV, but should be added for win32?
 vaCreateBuffer
 vaCreateBuffer2
 vaCreateConfig
 vaCreateContext
 vaCreateImage
 vaCreateMFContext
-vaCreateProtectedSession ## more va_prot.h API
 vaCreateSubpicture
-vaCreateSurfaces@@VA_API_0.33.0 ## symbol with version tag POSIX only
+vaCreateSurfaces
 vaDeassociateSubpicture
 vaDeriveImage
 vaDestroyBuffer
 vaDestroyConfig
 vaDestroyContext
 vaDestroyImage
-vaDestroyProtectedSession ## even more va_prot.h API
 vaDestroySubpicture
 vaDestroySurfaces
-vaDetachProtectedSession ## va_prot.h again
-vaDisplayIsValid ## internal API, used by va_wayland, va_glx ... not needed for win32
 vaEndPicture
 vaEntrypointStr
-va_errorMessage ## internal API, will send a PR to hide from the Linux builds
 vaErrorStr
 vaExportSurfaceHandle
 vaGetConfigAttributes
 vaGetDisplayAttributes
 vaGetImage
-vaGetLibFunc ## used solely by the Intel driver, to fetch a single function ... should really be removed see https://github.com/intel/compute-runtime/issues/569
-va_infoMessage ## internal API, will send a PR to hide from the Linux builds
 vaInitialize
 vaLockSurface
 vaMapBuffer
@@ -53,9 +42,7 @@
 vaMFSubmit
 va_newDisplayContext
 va_newDriverContext
-va_parseConfig ## internal API, will send a PR to hide from the Linux builds
 vaProfileStr
-vaProtectedSessionExecute ## final va_prot.h API
 vaPutImage
 vaQueryConfigAttributes
 vaQueryConfigEntrypoints
@@ -80,14 +67,8 @@
 vaSetInfoCallback
 vaSetSubpictureChromakey
 vaSetSubpictureGlobalAlpha
-vaSetSubpictureImage ## should be added for win32?
-vaStatusStr ## should be added for win32?
 vaSyncBuffer
 vaSyncSurface
-vaSyncSurface2 ## should be added for win32?
 vaTerminate
-va_trace_flag ## used by va_x11
-va_TracePutSurface ## used by va_x11
-va_TraceStatus ## internal API, will send a PR to hide from the Linux builds
 vaUnlockSurface
 vaUnmapBuffer
 ```

sivileri:
> Just compared the va.def against the exported symbols on my Arch machine, here is the diff with some inline notes:
> 
> ```diff
>  vaAcquireBufferHandle
> -VA_API_0.32.0 ## version tag, POSIX specific
> -VA_API_0.33.0 ## version tag, POSIX specific
>  vaAssociateSubpicture
> -vaAttachProtectedSession ## va_prot.h API, should be added for win32?
>  vaBeginPicture
>  vaBufferInfo
>  vaBufferSetNumElements
>  vaBufferTypeStr
>  vaConfigAttribTypeStr
> -vaCopy ## Maybe not great from efficiency POV, but should be added for win32?
>  vaCreateBuffer
>  vaCreateBuffer2
>  vaCreateConfig
>  vaCreateContext
>  vaCreateImage
>  vaCreateMFContext
> -vaCreateProtectedSession ## more va_prot.h API
>  vaCreateSubpicture
> -vaCreateSurfaces@@VA_API_0.33.0 ## symbol with version tag POSIX only
> +vaCreateSurfaces
>  vaDeassociateSubpicture
>  vaDeriveImage
>  vaDestroyBuffer
>  vaDestroyConfig
>  vaDestroyContext
>  vaDestroyImage
> -vaDestroyProtectedSession ## even more va_prot.h API
>  vaDestroySubpicture
>  vaDestroySurfaces
> -vaDetachProtectedSession ## va_prot.h again
> -vaDisplayIsValid ## internal API, used by va_wayland, va_glx ... not needed for win32
>  vaEndPicture
>  vaEntrypointStr
> -va_errorMessage ## internal API, to be hidden with https://github.com/intel/libva/pull/637
>  vaErrorStr
>  vaExportSurfaceHandle
>  vaGetConfigAttributes
>  vaGetDisplayAttributes
>  vaGetImage
> -vaGetLibFunc ## used solely by the Intel driver, to fetch a single function ... should really be removed see https://github.com/intel/compute-runtime/issues/569
> -va_infoMessage ## internal API, to be hidden with https://github.com/intel/libva/pull/637
>  vaInitialize
>  vaLockSurface
>  vaMapBuffer
> @@ -53,9 +42,7 @@
>  vaMFSubmit
>  va_newDisplayContext
>  va_newDriverContext
> -va_parseConfig ## internal API, to be hidden with https://github.com/intel/libva/pull/637
>  vaProfileStr
> -vaProtectedSessionExecute ## final va_prot.h API
>  vaPutImage
>  vaQueryConfigAttributes
>  vaQueryConfigEntrypoints
> @@ -80,14 +67,8 @@
>  vaSetInfoCallback
>  vaSetSubpictureChromakey
>  vaSetSubpictureGlobalAlpha
> -vaSetSubpictureImage ## should be added for win32?
> -vaStatusStr ## should be added for win32?
>  vaSyncBuffer
>  vaSyncSurface
> -vaSyncSurface2 ## should be added for win32?
>  vaTerminate
> -va_trace_flag ## used by va_x11
> -va_TracePutSurface ## used by va_x11
> -va_TraceStatus ## internal API, to be hidden with https://github.com/intel/libva/pull/637
>  vaUnlockSurface
>  vaUnmapBuffer
> ```

@evelikov Thanks for pointing these out, I've added the couple missing public API ones

image

XinfengZhang (approved):

PR #542, opened 2021-09-30
build: Wayland scanner should generate private code.

ceyusa:
wayland-scanner emits a warning:

Using "code" is deprecated - use private-code or public-code.
See the help page for details.

This patch updates argument to private-code.

PR #627, opened 2022-09-11
Misc wayland-scanner fixes

evelikov:
While skimming at the wayland-scanner code/private-code aka a few more issues showed up in the existing handling:
 - the wayland-client.pc prefix is used to look for wayland-scanner -> while wayland-scanner.pc wayland_scanner var has been a thing for almost a decade
 - meson will not pick the native program, so cross compilation can fail
 - we need a version check 1.15 for wayland-scanner, as well as removing of the existing export.c workaround

This PR supersedes https://github.com/intel/libva/pull/542

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -327,7 +327,7 @@ if test "x$enable_wayland" != "xno"; then
 
         AC_PATH_PROG([WAYLAND_SCANNER], [wayland-scanner])
         if test "x$WAYLAND_SCANNER" = "x"; then
-            PKG_CHECK_MODULES([WL_SCANNER], [wayland-scanner],
+            PKG_CHECK_MODULES([WL_SCANNER], [wayland-scanner >= 1.15],
                 [USE_WAYLAND="yes"], [:])
 
             if test "x$USE_WAYLAND" = "xno" -a "x$enable_wayland" = "xyes"; then
--- a/configure.ac
+++ b/configure.ac
@@ -325,11 +325,15 @@ if test "x$enable_wayland" != "xno"; then
 
     if test "$USE_WAYLAND" = "yes"; then
 
-        WAYLAND_PREFIX=`$PKG_CONFIG --variable=prefix wayland-client`
-        AC_PATH_PROG([WAYLAND_SCANNER], [wayland-scanner],,
-                     [${WAYLAND_PREFIX}/bin$PATH_SEPARATOR$PATH])
+        AC_PATH_PROG([WAYLAND_SCANNER], [wayland-scanner])
         if test "x$WAYLAND_SCANNER" = "x"; then
-            AC_MSG_ERROR([wayland-scanner not found: Install it or use --disable-wayland])
+            PKG_CHECK_MODULES([WL_SCANNER], [wayland-scanner],
+                [USE_WAYLAND="yes"], [:])
+
+            if test "x$USE_WAYLAND" = "xno" -a "x$enable_wayland" = "xyes"; then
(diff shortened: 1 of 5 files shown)


evelikov:
The exact meson and autotools logic has been used in mesa and Xorg for ~5 years, so I think it's been battle tested more than enough.

dvrogozh:
@ceyusa : are you ok with these changes?

meson.build
```diff
@@ -103,7 +103,8 @@ if get_option('with_wayland') != 'no'
   wayland_dep = dependency('wayland-client', version : '>= 1.11.0',
 			   required : get_option('with_wayland') == 'yes')
   wayland_scanner_dep = dependency('wayland-scanner',
-                                   required : get_option('with_wayland') == 'yes')
+                                   required : get_option('with_wayland') == 'yes',
+                                   native : true)
```
dvrogozh, line 107: what does "native" mean here?

meson.build
```diff
@@ -102,7 +102,7 @@ WITH_WAYLAND = false
 if get_option('with_wayland') != 'no'
   wayland_dep = dependency('wayland-client', version : '>= 1.11.0',
 			   required : get_option('with_wayland') == 'yes')
-  wayland_scanner_dep = dependency('wayland-scanner',
+  wayland_scanner_dep = dependency('wayland-scanner', version : '>= 1.15',
```
dvrogozh, line 105: From commit message:
```
The "code" option was deprecated by yours truly with wayland-scanner 1.15.
```
I don't understand this sentence, can you, please, reformulate?

configure.ac
```diff
@@ -327,7 +327,7 @@ if test "x$enable_wayland" != "xno"; then
 
         AC_PATH_PROG([WAYLAND_SCANNER], [wayland-scanner])
         if test "x$WAYLAND_SCANNER" = "x"; then
-            PKG_CHECK_MODULES([WL_SCANNER], [wayland-scanner],
+            PKG_CHECK_MODULES([WL_SCANNER], [wayland-scanner >= 1.15],
```
dvrogozh, line 330:
```
The "code" option was deprecated by yours truly with wayland-scanner 1.15.
```
Please, reformulate this sentence.

meson.build
ceyusa: before wayland-scanner 1.15 there was not `code-private` only `code`

meson.build
ceyusa: that is going to be used by the build machine, not the host system. It's important for cross-compilation.

meson.build
evelikov: Aside: @dvrogozh is there a IRC/Slack/other channel that you're on - could not see you on the official ones.

dvrogozh: >  "by yours truly" as bringing confusion

right:). I really would appreciate simpler commit message.

> @dvrogozh is there a IRC/Slack/other channel that you're on - could not see you on the official ones.

I afraid I am not using those.

dvrogozh:
I would like to merge in the change this week. @ceyusa, based on your comments I assume you are ok with the change. @evelikov can you, please, simplify those commit messages we discussed above?

ceyusa:
>  @ceyusa, based on your comments I assume you are ok with the change. 

I'm OK.

evelikov:
Didn't see any examples what's too complex so I've only dropped the "by yours truly" phrase. If you want to bikeshed them more, feel free to do so. All contributors have push access to the branch.

dvrogozh (approved):
looks good

PR #636, opened 2022-09-30
add attributes to report dynamic resolution support

XinfengZhang:

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1028,6 +1028,13 @@ typedef enum {
      * The value returned uses the VAConfigAttribValEncPerBlockControl type.
      */
     VAConfigAttribEncPerBlockControl    = 55,
+    /** \brief dynamic resolution change support. Read-only.
+     *
+     * This attributes describes whether backend driver support dynamic resolution change
+     * It could be used for encode/decode profile to indicate whether need to reset context.
+     * The value returned uses the /c VAConfigAttribValDynamicResolutionChange.
+     */
+    VAConfigAttribDynamicResolutionChange = 56,
     /**@}*/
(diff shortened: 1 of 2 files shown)


va/va.h
```diff
@@ -1028,6 +1028,13 @@ typedef enum {
      * The value returned uses the VAConfigAttribValEncPerBlockControl type.
      */
     VAConfigAttribEncPerBlockControl    = 55,
+    /** \brief dynamic resolution change support. Read-only.
+     *
+     * This attributes describes whether backend driver support dynamic resolution change
```
FurongZhang, line 1033: Correct the grammar: This attribute(-s) describes whether the backend driver supports(+s)...indicate whether needing to reset context.

va/va.h
```diff
@@ -1423,6 +1430,23 @@ typedef union _VAConfigAttribValEncPerBlockControl {
...
 
+typedef union _VAConfigAttribValDynamicResolutionChange {
+    struct {
+        /** \brief whether to support dynamic resolution change */
+        uint32_t drc_support             : 1;
+        /** \brief whether need to reset context for drc
+         *
+         *  0: invalid value
+         *  1: need to reset context for drc
+         *  2: could not reset context for drc
```
feiwan1, line 1441: It would be better to indicate "In this case, context resolution will not limit runtime surface resolution."

va/va.h
```diff
@@ -1423,6 +1430,23 @@ typedef union _VAConfigAttribValEncPerBlockControl {
...
+         */
+        uint32_t drc_context_reset       : 4;
```
feiwan1, line 1443: 2 bits are enough.

PR #552, opened 2021-11-18
add Linux ftrace mode for va trace, offload trace data format to trace tool

lindongw:
Signed-off-by: Lindong Wu <lindong.wu@intel.com>

PR #637, opened 2022-10-02
va_trace: annotate internal functions with DLL_HIDDEN

evelikov:
Annotate the following functions with DLL_HIDDEN:
 - va_errorMessage
 - va_infoMessage
 - va_parseConfig
 - va_TraceEvent
 - va_TraceEventBuffers
 - va_TraceStatus

The APIs have been used solely within libva.so ever since they were introduced. As such we can safely hide it was never used externally and it's not part of the public API.

--
Couple of the above (aka `va_TraceEvent*`) were added recently with https://github.com/intel/libva/pull/552, others like `va_TraceStatus` have been around since 2019, while the rest have been a thing for even longer.

@lindongw to double-check the `va_TraceEvent` symbols that you've added are meant to be internal-only, correct?

### DIFF
--- a/va/va_internal.h
+++ b/va/va_internal.h
@@ -32,9 +32,13 @@ extern "C" {
 #define CTX(dpy) (((VADisplayContextP)dpy)->pDriverContext)
 #define CHECK_DISPLAY(dpy) if( !vaDisplayIsValid(dpy) ) { return VA_STATUS_ERROR_INVALID_DISPLAY; }
 
+DLL_HIDDEN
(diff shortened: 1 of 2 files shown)


evelikov:
@XinfengZhang @dvrogozh these are safe to backport to 2.16

lindongw:
yes, va_TraceEvent is internal funcation

evelikov:
@XinfengZhang @dvrogozh ping?

XinfengZhang (approved):

evelikov:
@XinfengZhang can we merge this?

Issue #635, opened 2022-09-28
build failure without libxcb-x11


XinfengZhang:
if we are using a container to build libva, user have to update the image after PR #614  merged. if they still want libva-x11.  
so , suggest to separate DRI3 support with a new define macro, such as -HAVE_VA_X11_XCB. then give the time to users to add the correlated dependencies in build image. 

@evelikov @dvrogozh , what's your opinions on it?

evelikov:
We don't have a DRI2 option, so a DRI3 one makes little sense. In addition, having this configurable makes it easier for people to build things "wrong" and then report bugs when Xwayland "does not work".

I'm happy to send a PR that adds the dependencies - just point me to the project.

dvrogozh:
I also think we don't need to add a DRI3 specific option. Projects do pull in new dependencies - that's normal. Master branch and new release (2.17?) will be affected, previous release - will not. Users need to adopt.

PR #638, opened 2022-10-05
ci: add libx11-xcb-dev libxcb-dri3-dev to trigger X11 backend build

dvrogozh:
We forgot to add new packages required by recently merged DRI3 support into our ci install steps. As a result we were building w/o X11 actually. Fixing this now an re-enabling X11 backend build.

See: #635
See: ef1df02 ("x11: add basic DRI3 support")

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -40,6 +40,8 @@ jobs:
           libegl1-mesa-dev \
           libgl1-mesa-dev \
           libx11-dev \
+          libx11-xcb-dev \
+          libxcb-dri3-dev \
           libxext-dev \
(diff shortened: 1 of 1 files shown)


evelikov:
Ouch sorry for missing these. The commit is
`Reviewed-by: Emil Velikov <emil.l.velikov@gmail.com>`

The meson/freebsd build is ok, since it pulls the packages indirectly via mesa.
Also one could reduce the duplication in the yaml, by using a `matrix`. But that's for another day.

dvrogozh:
> Also one could reduce the duplication in the yaml, by using a matrix. But that's for another day.

That's what I tell myself each time I look into or modify this yaml :).

PR #640, opened 2022-10-11
fix resouce check null

Dylan-debug:
resource need check null before use it. 

Signed-off-by: caij <jianxing.cai@intel.com>


evelikov:
Out of curiosity: how did you spot this - via manual observation or via a third-party tool?

In case of the latter, it's nice to mention/credit them in the commit message. Which in itself is very non-descriptive. How about something like:

```
x11: avoid null pointer deference in va_DisplayContextDestroy

Issue detected by XXX tool.
```

dvrogozh (approved):
looks good

PR #642, opened 2022-10-17
ci: add deploy docs

jangsc0000:
Deploy release documentation automatically.

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -1,15 +1,19 @@
-
 name: deploy docs
+
 on:
   release:
-    types: [published]
+    types: [released]
     
 jobs:
-  build_and_deploy:
+  build:
     runs-on: ubuntu-latest
-    steps:
+    steps: 
     - uses: actions/checkout@v3
-    - name: install prerequisites
+    - id: tag_name
+      name: Get latest tag_name
+      run: |        
+        echo "GIT_LATEST_TAG=`echo $(git tag | grep '^[0-9]\+\.[0-9]\+\?\.[0-9]\+\?$' | sort -V -r | head -n1)`" >> $GITHUB_ENV
+    - name: Install prerequisites
       run: |
         sudo apt-get update
         sudo apt-get install -y \
@@ -21,25 +25,20 @@ jobs:
           libxfixes-dev \
           libwayland-dev \
           libx11-xcb-dev \
-           libxcb-dri3-0 \
-          doxygen 
+          libxcb-dri3-0 \
+          doxygen \
+          ninja-build
     - uses: actions/setup-python@v1
       with:
         python-version: '3.10' 
-
-    - name: Build
+    - name: Build docs
       run: |
-        python -m pip install ninja
         python -m pip install meson
         meson _build -Denable_docs=true
         meson compile -C _build
-
-    - name: Build echo
-      run:  echo ${{ github.ref }}
-
     - name: Deploy
       uses: peaceiris/actions-gh-pages@v3
+      if: ${{ github.ref_name }} == ${{ env.GIT_LATEST_TAG }}
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: ./_build/doc/html-out
-      
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -1,0 +1,45 @@
+
+name: deploy docs
+on:
+  release:
+    types: [published]
+    
+jobs:
+  build_and_deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: install prerequisites
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y \
+          libdrm-dev \
+          libegl1-mesa-dev \
+          libgl1-mesa-dev \
+          libx11-dev \
+          libxext-dev \
+          libxfixes-dev \
+          libwayland-dev \
+          libx11-xcb-dev \
+           libxcb-dri3-0 \
+          doxygen 
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.10' 
+
+    - name: Build
+      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
...
+        meson compile -C _build
+
+    - name: Build echo
+      run:  echo ${{ github.ref }}
+
+    - name: Deploy
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: ./_build/doc/html-out
```
XinfengZhang, line 44: now sure about this step, could update the https://github.com/intel/libva/tree/gh-pages directly?
or just simulate it? I dont want to update the doc directly, 
will update it with releases.

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
+
+name: deploy docs
+on:
+  release:
+    types: [published]
```
jangsc0000, line 5: name: deploy docs
on:
  release:
    types: [published]

.github/workflows/deploy_docs.yml
jangsc0000: When you release, the documentation in the release commit is updated.

.github/workflows/deploy_docs.yml
jangsc0000, line 5: This workflow only works when released by this part.

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
...
+          libx11-xcb-dev \
+           libxcb-dri3-0 \
+          doxygen 
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.10' 
+
+    - name: Build
+      run: |
+        python -m pip install ninja
```
dvrogozh, line 32: why install via pip vs. from ubuntu packages?

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
...
+        sudo apt-get install -y \
+          libdrm-dev \
+          libegl1-mesa-dev \
+          libgl1-mesa-dev \
+          libx11-dev \
+          libxext-dev \
+          libxfixes-dev \
+          libwayland-dev \
+          libx11-xcb-dev \
+           libxcb-dri3-0 \
```
dvrogozh, line 24: fix alignment, please

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
...
+        python -m pip install ninja
+        python -m pip install meson
+        meson _build -Denable_docs=true
+        meson compile -C _build
+
+    - name: Build echo
+      run:  echo ${{ github.ref }}
+
+    - name: Deploy
+      uses: peaceiris/actions-gh-pages@v3
```
dvrogozh, line 41: I did not look closely into github pages, but there seems to be standard ways of doing that. Why use custom github action?

* https://github.com/actions/deploy-pages
* https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#creating-a-custom-github-actions-workflow-to-publish-your-site
* https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

.github/workflows/deploy_docs.yml
dvrogozh: what will happen if we release 2.17.0, ..., 2.17.4, 2.18.0 and then release 2.17.5? will 2.17.5 override 2.18.0 posting?fff

dvrogozh (changes requested):

.github/workflows/deploy_docs.yml
jangsc0000: I tried this, but I can't deploy with release tag.

.github/workflows/deploy_docs.yml
```diff
@@ -1,15 +1,19 @@
...
+  build:
     runs-on: ubuntu-latest
-    steps:
+    steps: 
     - uses: actions/checkout@v3
-    - name: install prerequisites
+    - id: tag_name
+      name: Get latest tag_name
+      run: |        
+        echo "GIT_LATEST_TAG=`echo $(git tag | grep '^[0-9]\+\.[0-9]\+\?\.[0-9]\+\?$' | sort -V -r | head -n1)`" >> $GITHUB_ENV
```
jangsc0000, line 15: By checking the tag, only the latest version of the document is updated.

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
...
+    
+jobs:
+  build_and_deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: install prerequisites
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y \
```
evelikov, line 19: A better option IMHO is to use:
 - apt-get build-dep libva
 - apt-get install extra-deps-not-installed-with-above

.github/workflows/deploy_docs.yml
evelikov: Ninja is not a python module so the above does not make sense. Ideally both ninja and meson (below) should be installed with "apt-get install ..." if needed

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: Build
+      run: |
+        python -m pip install ninja
+        python -m pip install meson
+        meson _build -Denable_docs=true
+        meson compile -C _build
+
+    - name: Build echo
+      run:  echo ${{ github.ref }}
```
evelikov: This is a no-op - drop?

jangsc0000: It was removed

evelikov: The change should be squashed appropriately

PR #644, opened 2022-10-20
meson: Set va_win32 soversion to '' and remove the install_data rename

sivileri:
Fixes installed DLLs and PDBs having mismatched names (-<version> suffix). Sets soversion to '' for windows platforms to avoid having suffixes in the resulting DLLs/PDBs. With this change we also simplify meson.build for libva-win32 by removing the install rename step.

### DIFF
--- a/va/meson.build
+++ b/va/meson.build
@@ -66,7 +66,7 @@ libva = shared_library(
             libva_headers +
             libva_headers_priv,
   vs_module_defs : 'libva.def',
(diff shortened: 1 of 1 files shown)


dvrogozh (approved):
looks good

sivileri:
@dvrogozh thanks for taking a look, could you please merge it?
CC @XinfengZhang

PR #646, opened 2022-10-24
ci: update versions for checkout and set-python

dvrogozh:
See: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

@sivileri : fyi


uartie (approved):
LGTM

Issue #645, opened 2022-10-24
libva-win32: warnings during compile/link 


dvrogozh:
See: https://github.com/intel/libva/actions/runs/3298661454
Commit: https://github.com/intel/libva/commit/da4a18fe814c9695f420d466ad3065cebdbe288a

* windows-msvc:
```
LINK : warning LNK4044: unrecognized option '/Wl,-version-script,D:\a\libva\libva\va/libva.syms'; ignored
```

* windows-mingw:
```
../va/va_trace.c:947:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
```

@sivileri : can you, please, help to fix these?

sivileri:
@dvrogozh https://github.com/intel/libva/pull/647 has some changes for these

PR #647, opened 2022-10-24
Fix Windows CI build warnings

sivileri:
Fixes https://github.com/intel/libva/issues/645

See: https://github.com/intel/libva/actions/runs/3298661454
Commit: https://github.com/intel/libva/commit/da4a18fe814c9695f420d466ad3065cebdbe288a

windows-msvc:
`LINK : warning LNK4044: unrecognized option '/Wl,-version-script,D:\a\libva\libva\va/libva.syms'; ignored
`

windows-mingw:
`../va/va_trace.c:947:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]`

### DIFF
--- a/va/meson.build
+++ b/va/meson.build
@@ -56,7 +56,14 @@ libva_headers_priv = [
 ]
 
 libva_sym = 'libva.syms'
-libva_sym_path = '@0@/@1@'.format(meson.current_source_dir(), libva_sym)
+libva_sym_arg = '-Wl,-version-script,' + '@0@/@1@'.format(meson.current_source_dir(), libva_sym)
+
+libva_link_args = []
+libva_link_depends = []
+if cc.links('', name: '-Wl,--version-script', args: ['-shared', libva_sym_arg])
+  libva_link_args = libva_sym_arg
+  libva_link_depends = libva_sym
+endif
 
 install_headers(libva_headers, subdir : 'va')
 
@@ -70,8 +77,8 @@ libva = shared_library(
   version : libva_lt_version,
   c_args : [ '-DSYSCONFDIR="' + sysconfdir + '"'] + ['-DVA_DRIVERS_PATH="' + driverdir + '"'] + va_c_args,
   include_directories : configinc,
-  link_args : '-Wl,-version-script,' + libva_sym_path,
-  link_depends : libva_sym,
+  link_args : libva_link_args,
+  link_depends : libva_link_depends,
   install : true,
   dependencies : [ dl_dep ])
 
--- a/va/sysdeps.h
+++ b/va/sysdeps.h
@@ -48,12 +48,19 @@
 # include <log/log.h>
 #endif
 
-#if defined __GNUC__ && defined HAVE_GNUC_VISIBILITY_ATTRIBUTE
+// Defines for visibility attribute
+// based on https://gcc.gnu.org/wiki/Visibility
+#if defined _WIN32 || defined __CYGWIN__
+# define DLL_EXPORT __declspec(dllexport)
+# define DLL_HIDDEN
+#else
+#if __GNUC__ >= 4
 # define DLL_HIDDEN __attribute__((visibility("hidden")))
 # define DLL_EXPORT __attribute__((visibility("default")))
 #else
 # define DLL_HIDDEN
 # define DLL_EXPORT
+#endif
 #endif
 
 #endif /* SYSDEPS_H */
--- a/configure.ac
+++ b/configure.ac
@@ -187,28 +187,6 @@ if test "$enable_va_messaging" = "yes"; then
     AC_DEFINE([ENABLE_VA_MESSAGING], [1], [Defined to 1 if va messaging is needed])
 fi
 
-# Check for __attribute__((visibility()))
-AC_CACHE_CHECK([whether __attribute__((visibility())) is supported],
-    ac_cv_have_gnuc_visibility_attribute,
-    [cat > conftest.c <<EOF
-int foo __attribute__ ((visibility ("hidden"))) = 1;
-int bar __attribute__ ((visibility ("protected"))) = 1;
-EOF
-    ac_cv_have_gnuc_visibility_attribute="no"
-    if ${CC-cc} -Werror -S conftest.c -o conftest.s >/dev/null 2>&1; then
-        if grep '\.hidden.*foo' conftest.s >/dev/null; then
-            if grep '\.protected.*bar' conftest.s >/dev/null; then
-                ac_cv_have_gnuc_visibility_attribute="yes"
-            fi
-        fi
-    fi
-    rm -f conftest.[cs]
-])
-if test "$ac_cv_have_gnuc_visibility_attribute" = "yes"; then
-    AC_DEFINE([HAVE_GNUC_VISIBILITY_ATTRIBUTE], [1],
-              [Defined to 1 if GCC visibility attribute is supported])
-fi
-
 # Check for -ldl (often not required)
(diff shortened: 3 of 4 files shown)


va/sysdeps.h
```diff
@@ -48,7 +48,7 @@
 # include <log/log.h>
 #endif
 
-#if defined __GNUC__ && defined HAVE_GNUC_VISIBILITY_ATTRIBUTE
+#if defined __GNUC__ && defined HAVE_GNUC_VISIBILITY_ATTRIBUTE && !defined(_WIN32)
```
dvrogozh, line 51: Hm. Actually we had this:
```
meson.build:125:  va_c_args += ['-DHAVE_GNUC_VISIBILITY_ATTRIBUTE']

# Symbol visibility
if cc.has_argument('-fvisibility=hidden')
  va_c_args += ['-DHAVE_GNUC_VISIBILITY_ATTRIBUTE']
endif
```

Why this did not do the job?

va/meson.build
```diff
@@ -70,8 +70,8 @@ libva = shared_library(
   version : libva_lt_version,
   c_args : [ '-DSYSCONFDIR="' + sysconfdir + '"'] + ['-DVA_DRIVERS_PATH="' + driverdir + '"'] + va_c_args,
   include_directories : configinc,
-  link_args : '-Wl,-version-script,' + libva_sym_path,
-  link_depends : libva_sym,
+  link_args : host_machine.system() == 'windows' ? [] : '-Wl,-version-script,' + libva_sym_path,
```
dvrogozh, line 73: what's the operation priority here? do we risk to end up with just `+ libva_sym_path`?

va/sysdeps.h
sivileri: The  `HAVE_GNUC_VISIBILITY_ATTRIBUTE` condition evaluates to false correctly for MSVC with `CC=cl`, but for MinGW with `CC=gcc` it evaluates to true, which causes these warnings

image

va/meson.build
sivileri: I checked on Ubuntu/MSVC and the build command line was correct. Just to make sure a future meson change doesn't break this, I added the parentheses around it to make it explicit

va/sysdeps.h
dvrogozh: So, we stepped into non 100% guarantee that has_argument() works correctly. I suggest to do this on meson.build level instead of in source.

sivileri: Added the condition in meson.build, validated no longer repro in mingw local build with the change

eli-schwartz (changes requested):

va/sysdeps.h
dvrogozh: > This is wrong, please do not do this. Meson natively handles visibility by detecting if it's appropriate for the compiler, and there is no reason to avoid defining this attribute arbitrarily -- if you have GNUC, set it.

@eli-schwartz : if meson supports that, why are we getting the following warning: See above discussion, meson detected that visibility is supported, passed the flag and we get the warning. Can you suggest what was done incorrectly? 
```
../va/va_trace.c:947:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
```

sivileri: I think the problem is MINGW gcc specifically, so I've now disabled the attributes specifically in that case, I'm open to adding a more comprehensive check if it's available in meson. I tried both ways below in meson and both ways are supported when building in gcc/mingw but then the warning shows up.

```
if cc.has_function_attribute('visibility:hidden')
if cc.has_argument('-fvisibility=hidden')
```

va/meson.build
```diff
@@ -57,6 +57,13 @@ libva_headers_priv = [
 
 libva_sym = 'libva.syms'
 libva_sym_path = '@0@/@1@'.format(meson.current_source_dir(), libva_sym)
+libva_link_args = '-Wl,-version-script,' + libva_sym_path
```
dvrogozh, line 60: Can we do that vice versa?
```
libva_link_args = []
libva_link_depends = []
if cc.links('', name: '-Wl,--version-script', args: ['-shared', libva_link_args])
  libva_link_args = '-Wl,-version-script,' + libva_sym_path
  libva_link_depends = libva_sym
endif
```

sivileri: libva_link_args needs to be` '-Wl,-version-script,' + libva_sym_path` when checking. I refactored to be almost like your suggestion with that small change

va/sysdeps.h
```diff
@@ -48,7 +48,7 @@
 # include <log/log.h>
 #endif
 
-#if defined __GNUC__ && defined HAVE_GNUC_VISIBILITY_ATTRIBUTE
+#if defined __GNUC__ && defined HAVE_GNUC_VISIBILITY_ATTRIBUTE && !defined(__MINGW32__)
```
dvrogozh, line 51: @eli-schwartz: can you, please, comment on this? are you ok with this change considering wrong detection of visibility attribute for mingw?

eli-schwartz: I object to the use of HAVE_GNUC_VISIBILITY_ATTRIBUTE.

I'm -0 with checking for `__MINGW32__`, it seems overly specific to the mingw case when really this is about Windows, but it probably doesn't matter if you don't support building with MSVC...

dvrogozh: > but it probably doesn't matter if you don't support building with MSVC

We do support building with MSVC. I afraid I don't understand what's your proposal to handle this case. We do have `cc.has_argument` already and it does not seem to work correctly:
```
# Symbol visibility
if cc.has_argument('-fvisibility=hidden')
  va_c_args += ['-DHAVE_GNUC_VISIBILITY_ATTRIBUTE']
endif
```
See: https://github.com/intel/libva/blob/da4a18fe814c9695f420d466ad3065cebdbe288a/meson.build#L124

Are you saying we should not pass `HAVE_GNUC_VISIBILITY_ATTRIBUTE` at all? But how to handle that then. Please, be specific.

eli-schwartz: > Instead of checking for i`f __GNUC__ >= 4`, we use `HAVE_GNUC_VISIBILITY_ATTRIBUTE` to check dynamically for compiler attribute support, which will also be set to false by meson when building with MSVC. 

This right here is the thing I'm suggesting to not do.

Just check for `__GNUC__` as the GCC wiki suggests. Compilers which are compatible with the GNUC visibility attribute announce that support by defining their `__GNUC__` compatibility level. You don't need to fork and exec the compiler during setup in order to check that -- the preprocessor can do it much more efficiently while the source code is being parsed.

This makes your source code more reliable (less context depends on the build system, less things you need to implement multiple times -- once in meson.build and once in configure.ac) and speeds up builds. It's a win-win.

Why do you need to run configure-time checks? What problem are you trying to solve?

The problem in question isn't Windows support, because not only does MSVC not define `__GNUC__`, but you can just check `_WIN32` anyway, which works on mingw too.

dvrogozh: Thank you for clarification. This makes sense to me.

eli-schwartz: If you implement my suggestion, you can drop 22 lines from configure.ac right here:

https://github.com/intel/libva/blob/7afc9aea5b350a0ed114969c98f6449dce49ded5/configure.ac#L190-L211

sivileri: @eli-schwartz thanks for the clarification, I agree it's a simpler and better way. I've just pushed changes with it, please let me know what you think.

eli-schwartz: It.looka very nice now. :D

eli-schwartz (approved):

PR #648, opened 2022-11-05
editorconfig: add initial config

evelikov:
Allows for consistent coding experience across editors.

Signed-off-by: Emil Velikov <emil.velikov@collabora.com>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# To use this config with your editor, follow the instructions at:
+# http://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+
+[*.{c,h}]
+indent_style = space
+indent_size = 4
+
+[{Makefile.*,*.mk}]
+indent_style = tab
+
+[{meson.build,meson_options.txt}]
+indent_style = space
+indent_size = 2
--- a/Android.mk
+++ b/Android.mk
@@ -1,4 +1,4 @@
 # Recursive call sub-folder Android.mk
 #
 
- include $(call all-subdir-makefiles)
+include $(call all-subdir-makefiles)


evelikov:
@XinfengZhang @dvrogozh humble ping? Is there anything I can do to move this forward?

.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+[*]
+charset = utf-8
+insert_final_newline = true
+
+[*.{c,h}]
+indent_style = space
+indent_size = 4
+
+[{Makefile.*,*.mk}]
+indent_style = tab
```
XinfengZhang, line 15: not sure the usage of this config
it could be used to unify the code auto align style? no matter they are using vim or other editor?

evelikov: Nearly every editor out there support this (either built-in or via a simple plugins) - vim, emacs, eclipse, visual studio code, etc see http://editorconfig.org

Since it's used by the editor it will leave "less work" for the `style_check` script. The two are effectively complementing each other - they are not exclusive.

evelikov:
v2 add `end_of_line = lf`

With the extra config token, commits like https://github.com/intel/libva/pull/646/commits/bd1b8459df67da0d7cb244e29919526ffe0d5864 should be far less likely - @dvrogozh @sivileri

.editorconfig
XinfengZhang: I quick checked the android.mk file, it just one space https://github.com/intel/libva/blob/master/Android.mk#L4, so, how about this take effect?

XinfengZhang (commented):
thanks, appreciate the motivation of this change, looks it will let contributors feel convenient

.editorconfig
evelikov: Seems like a spurious white-space that can/should be removed. It bares no change in functional behaviour.

evelikov:
v3: path-ignore the file in CI

So any changes to that file do not trigger a rebuild of everything. There is a similar change for the style-checking, which will be a separate MR.

.editorconfig
evelikov: @XinfengZhang humble ping. Is there anything I can do to move this forward?

XinfengZhang (approved):

evelikov:
@XinfengZhang @dvrogozh humble poke - anything I can do to move this forward?

evelikov:
@XinfengZhang @dvrogozh humble monthly poke

evelikov:
/me taps microphone - Hello, is thing on? @XinfengZhang @dvrogozh

evelikov:
AFAICT the PR has been reviewed and approved already. What can I do to actually have this merged?

evelikov:
> Just to avoid others stepping into the same as I did when I tried editorconfig, note this bug on Ubuntu: https://bugs.launchpad.net/ubuntu/+source/vim-editorconfig/+bug/1723800, i.e. you might need to configure it manually to use under vim after apt-get install: vim-addon-manager install editorconfig.

Thanks that was an interesting rocky road. It seems like with Debian and by extension Ubuntu have their own home-grown vim plugin manager, despite that there are 3+ other more popular and better maintained ones. Their solution requires you to _manually_ enable every plugin that you want to use.

On the other hand: Vim 8 circa 2016, has built-in support for plugins/packs. So if the distribution package manager [does the right thing](https://archlinux.org/packages/community/any/vim-editorconfig/) they will be picked automatically and just work out of the box. In a way this makes me more grateful about jumping out the Ubuntu ship years ago.

PR #649, opened 2022-11-07
ci: use matrix, add ubuntu 20.04 + clang-15

evelikov:
Currently we have a bit of duplication in the CI. Instead we can use a simple matrix and also gain ubuntu 20.04 + clang-15 combination.

To keep things sane, we've moved the clang-15 install into a standalone script.

... as promised in https://github.com/intel/libva/pull/638#issuecomment-1273771424

/cc @dvrogozh

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -5,33 +5,31 @@ on:
     paths-ignore:
     - '.github/workflows/**'
     - '!.github/workflows/ubuntu.yml'
+    - '!.github/workflows/install-clang.sh'
   pull_request:
     paths-ignore:
     - '.github/workflows/**'
     - '!.github/workflows/ubuntu.yml'
+    - '!.github/workflows/install-clang.sh'
 
 env:
   CFLAGS: -Wall -Werror
 
 jobs:
-  clang-15:
-    runs-on: ubuntu-22.04
+  test:
+    strategy:
+      matrix:
+        compiler: [clang-15, gcc]
+        os: [ubuntu-22.04, ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
     env:
-      CC: /usr/bin/clang-15
-      CXX: /usr/bin/clang++-15
-      ASM: /usr/bin/clang-15
+      CC: ${{ matrix.compiler }}
+      DISTRO: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
     - name: install toolchain
-      run: |
-        if [[ -e $CC && -e $CXX ]]; then \
-          echo "clang-15 already presents in the image"; \
-        else \
-          echo "clang-15 missed in the image, installing from llvm"; \
-          echo "deb [trusted=yes] http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main" | sudo tee -a /etc/apt/sources.list; \
-          sudo apt-get update; \
-          sudo apt-get install -y --no-install-recommends clang-15; \
-        fi
+      if: ${{ (matrix.compiler == 'clang-15') }}
+      run: .github/workflows/install-clang.sh 15
     - name: install prerequisites
       run: |
         sudo apt-get update
@@ -45,6 +43,8 @@ jobs:
           libxext-dev \
           libxfixes-dev \
           libwayland-dev
+    - name: print compiler version
+      run: ${{ matrix.compiler }} --version
     - name: configure
       run: ./autogen.sh --prefix=/usr
     - name: build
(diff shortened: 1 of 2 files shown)


.github/install-clang-15.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
dvrogozh, line 1: clang-15 is not printed in stone for our usage. Basically, once clang-16 will be released there will be big desire to switch to it. So, similar script will be needed. I suggest to generalize that and rename script to `install-clang.sh` and allow it accept the argument - version number. As such, we will be able to call:
```
./install-clang.sh 15
./install-clang.sh 16
```

.github/install-clang-15.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -euo pipefail
+
+if [[ $CC != "clang-15" ]]; then
```
dvrogozh, line 5: Can we, please, have this check explicitly on github actions level? As such, I do not think this check here is not needed. I also suggest to avoid using external `CC` variable in this script and construct the required expression from the script argument I suggested above: `clang-$1`.

.github/workflows/ubuntu.yml
```diff
@@ -5,59 +5,30 @@ on:
...
-    - name: check
-      run: make check
-    - name: install
-      run: sudo make install
-
-  ubuntu-22-04:
-    runs-on: ubuntu-22.04
-    steps:
-    - uses: actions/checkout@v3
+      run: .github/install-clang-15.sh
```
dvrogozh, line 31:
```
    steps:
    - uses: actions/checkout@v3
    - name: install toolchain
      if: ${{ matrix.compiler == clang-15 }} 
      run: .github/install-clang.sh 15
```

.github/workflows/ubuntu.yml
```diff
@@ -5,59 +5,30 @@ on:
...
+      matrix:
+        compiler: [clang-15, gcc]
+        os: [ubuntu-22.04, ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
     env:
-      CC: /usr/bin/clang-15
-      CXX: /usr/bin/clang++-15
-      ASM: /usr/bin/clang-15
+      CC: ${{ matrix.compiler }}
+      DISTRO: ${{ matrix.os }}
```
dvrogozh, line 27: Can you, please, add a step below which will print compiler version? For "clang-15" it's eventually obvious, but for gcc it's not. And I do not think it's getting printed anywhere in autoconf configuraiton.

dvrogozh (changes requested):

.github/install-clang-15.sh
evelikov: Existing code used hard-coded clang-15, so I followed suit. Making it more flexible sounds like a good idea though - hell one could even add few versions in the matrix.

.github/install-clang-15.sh
evelikov: Of course - your example below works perfectly.

.github/workflows/ubuntu.yml
evelikov: Didn't know about the "if:" syntax, thanks.

.github/workflows/ubuntu.yml
evelikov: Sure thing, added.

evelikov:
Hmm CI seems unhappy with the extra yaml changes. Will double-check and fix tomorrow

.github/workflows/ubuntu.yml
```diff
@@ -5,59 +5,31 @@ on:
...
-    - name: check
-      run: make check
-    - name: install
-      run: sudo make install
-
-  ubuntu-22-04:
-    runs-on: ubuntu-22.04
-    steps:
-    - uses: actions/checkout@v3
+      if: ${{ matrix.compiler == clang-15 }}
```
dvrogozh, line 31: It might need to be w/o `${{}}` + quoted for clang-15 to make it a string, https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif.
```
if: matrix.compiler == 'clang-15'
```

evelikov: An extra set of parenthesises + quoted string did the job.

Also changed the version print to use `${{ matrix.compiler }}`

dvrogozh (approved):
looks good

evelikov:
v3 - move install-clang.sh within .github/workflows/ ... so changes to it do not trigger windows/freebsd CI pipelines

dvrogozh:
> v3 - move install-clang.sh within .github/workflows/ ... so changes to it do not trigger windows/freebsd CI pipelines

That's a good catch. Thank you.

XinfengZhang (approved):

dvrogozh:
I will reconfigure ci required checked in project settings after the merge.

PR #650, opened 2022-11-07
drm: remove unnecessary dri2 version/extension query

evelikov:
We don't use any fancy DRI2 features - as seen we just fetch the version and don't check it. So there's no need for the query calls.

If we're on a server which lacks the extension, the auth call will fail.

NOTE: we cannot remove the API since it's used internally by the libva_x11 backend.

Signed-off-by: Emil Velikov <emil.l.velikov@gmail.com>


XinfengZhang (approved):

dvrogozh (approved):
looks good

PR #654, opened 2022-11-19
ci: add job to check for signed-off-by tags

evelikov:
As per the contributing documentation DCO/SoB is required.

Looking at the existing Github Actions, nearly all of them are based on
node12, which is deprecated and will be removed shortly. So opt for the
one that doesn't suffer that issue - cncf/dcochecker

### DIFF
--- a/.github/workflows/sob.yml
+++ b/.github/workflows/sob.yml
@@ -1,0 +1,9 @@
+name: DCO
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  dco-check:
+    uses: cncf/dcochecker/.github/workflows/dco.yml@main


evelikov:
@XinfengZhang @dvrogozh humble ping?

dvrogozh:
> Looking at the existing Github Actions, nearly all of them are based on
node12

I am not sure I understand what do you mean here. Do you mean that actions we have for libva in .github/workflows/ have a problem (I believe they do not since we've updated recently to avoid this issue)? Or you mean all actions for DCO/SoB use node12?

.github/workflows/sob.yml
```diff
@@ -0,0 +1,9 @@
+name: DCO
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  dco-check:
+    uses: cncf/dcochecker/.github/workflows/dco.yml@main
```
dvrogozh, line 9: I am paranoid on adding non-standard stuff into the project. Looking into https://github.com/cncf/dcochecker I am not sure how active it is supported. Besides, we can implement this with just something like `git -1 | grep Signed-off-by`, no need in 3rd party action.

dvrogozh (changes requested):

.github/workflows/sob.yml
evelikov: Just double-checking: By "non-standard" you mean, actions done by other parties - not by Github themselves. correct?

Sure it can be re-implemented locally - can give it a try unless someone beats be to it.

dvrogozh: > actions done by other parties - not by Github themselves. correct?

Yes. Everything outside of Github itself would need careful consideration. It's not impossible to take them, but we need a confidence these would be supported for a really long period of time.

evelikov:
Seems like there's an official github app to do this - open https://github.com/apps/dco and tap on "Configure".

Leaving this PR in place as a fall-back, until the maintainers have finalised the setup.

PR #655, opened 2022-11-19
ci: add meson permutation in the Ubuntu builds 

evelikov:
A number of Linux distributions have switched to the meson build. In
particular - Arch, Fedora, Gentoo and more are to follow in due time.

Even the libva builds that Chrome uses (aka instrumented libraries) is
using meson (on top of Ubuntu, currently aiming at switching to Focal).

Note: `meson compile` was introduced with meson 0.54, where we use 0.53
since that's available in Ubuntu 20.04. As such we fallback to calling
ninja directly

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -49,7 +49,5 @@ jobs:
       run: ./autogen.sh --prefix=/usr
     - name: build
       run: make
-    - name: check
-      run: make check
     - name: install
       run: sudo make install
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -21,6 +21,7 @@ jobs:
       matrix:
         compiler: [clang-15, gcc]
         os: [ubuntu-22.04, ubuntu-20.04]
+        build: [meson, autotools]
     runs-on: ${{ matrix.os }}
     env:
       CC: ${{ matrix.compiler }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ubuntu.yml
```diff
@@ -42,14 +43,25 @@ jobs:
           libxcb-dri3-dev \
           libxext-dev \
           libxfixes-dev \
-          libwayland-dev
+          libwayland-dev \
+          meson
```
dvrogozh, line 47: * https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md
* https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

Interesting, they indeed does not seem to include meson by default. At least according to readme description.

.github/workflows/ubuntu.yml
```diff
@@ -42,14 +43,25 @@ jobs:
...
     - name: print compiler version
       run: ${{ matrix.compiler }} --version
+    - name: meson setup
```
dvrogozh, line 50: Would grouping meson and autotools together be a little better?
```
    # meson path
    - name: meson setup
      run: meson setup ./builddir --prefix=/usr
    - name: meson compile
      run: meson compile -C ./builddir || ninja -C ./builddir
    - name: meson install
      run: sudo meson install -C ./builddir

    # autotools path
    - name: configure
      if: ${{ (matrix.build == 'autotools') }}
      run: ./autogen.sh --prefix=/usr
    - name: build
      if: ${{ (matrix.build == 'autotools') }}
      run: make
    - name: install
      if: ${{ (matrix.build == 'autotools') }}
      run: sudo make install
```

.github/workflows/ubuntu.yml
evelikov: Many distros do not include meson in the normal build tool-chain - not sure why.

.github/workflows/ubuntu.yml
evelikov: Thanks fixed

dvrogozh (approved):
looks good

XinfengZhang (approved):

PR #656, opened 2022-11-19
Always build with va-messaging API

evelikov:
Currently the API is enabled in meson, autotools and Android (non-user) builds.

We don't have any CI to check if the nothing is broken when building w/o the API. Plus overall disabling the API at build-time breaks the contract that the users (both end-users and API users) expect.

Namely: the API works and can be controlled via the config file and environment variable.


evelikov:
@XinfengZhang @dvrogozh any comments on this MR?

evelikov:
@XinfengZhang @dvrogozh monthly ping?

dvrogozh (approved):
This does make sense to me. I don't think such build granularity is needed. I would expect this feature to be always built. I would also expect it to be controlled via some environment variable which I think we don't have, but that would be a subject for other PR.

Change looks good to me.

evelikov:
As the note just above said - there is both a configuration file (see `va_parseConfig`) and an environment variable (see `va_MessagingInit` and `LIBVA_MESSAGING_LEVEL`).

Thanks o/

evelikov:
@XinfengZhang @dvrogozh humble poke - anything I can do to move this forward?

XinfengZhang:
@evelikov  I just want to wait to check whether there are any objection from user who are using --enable-va-messaging=no.
 seems no, the change is LGTM

XinfengZhang (approved):

evelikov:
@XinfengZhang do you have a user in mind - perhaps you can @ them?

Can I suggest using something like "unless we hear objections, I'm planning to merge this in 2 weeks"? Otherwise it's not clear a) what your stance is (for or against) and b) what time frames are we looking at.

PR #657, opened 2022-11-22
Address meson deprecation warnings

evelikov:
The meson build is throwing a couple of deprecation warnings - fix them.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,4 +1,8 @@
- headerdir = join_paths(meson.source_root(), 'va')
+if meson.version().version_compare('>= 0.56')
+  headerdir = join_paths(meson.project_source_root(), 'va')
+else
+  headerdir = join_paths(meson.source_root(), 'va')
+endif
 footer = join_paths(meson.current_source_dir(), 'va_footer.html')
 
 libva_headers_doc = [
--- a/va/meson.build
+++ b/va/meson.build
@@ -318,5 +318,9 @@ if WITH_WIN32
 endif
 
 foreach header : libva_headers_subproject
-  configure_file(output: fs.name(header), input: header, copy: true)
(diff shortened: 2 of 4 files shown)


evelikov:
@sivileri for the win32 fix.

Previously we were using `configure_file()`, which was copying the header early in the build process from `source/va/win32/va_win32.h` to `build/va/va_win32.h`. Now that happens at the very end (the replacement `fs.copyfile()` has different semantics) so the build fails.

va/meson.build
```diff
@@ -318,5 +318,9 @@ if WITH_WIN32
 endif
 
 foreach header : libva_headers_subproject
-  configure_file(output: fs.name(header), input: header, copy: true)
+  if meson.version().version_compare('>= 0.64')
+    fs.copyfile(header)
+  else
+    configure_file(output: fs.name(header), input: header, copy: true)
```
dvrogozh, line 324: Where I can find information on meson side that it got deprecated? I don't see that here: https://mesonbuild.com/Reference-manual_functions.html#configure_file

doc/meson.build
```diff
@@ -1,4 +1,8 @@
- headerdir = join_paths(meson.source_root(), 'va')
+if meson.version().version_compare('>= 0.56')
+  headerdir = join_paths(meson.project_source_root(), 'va')
+else
+  headerdir = join_paths(meson.source_root(), 'va')
```
dvrogozh, line 4: Briefly looking I don't see neither of these functions in meson documentation. And can't check deprecation status. Maybe I just don't know where to look.

dvrogozh (approved):
looks good

evelikov:
@dvrogozh Is there more I can do here to have this merged?

dvrogozh:
> @dvrogozh Is there more I can do here to have this merged?

Merged.

PR #658, opened 2022-11-22
ci: pretty and consistent messages, restrict better

evelikov:
This is a mixed bag of CI bits:
 - use consistent conventions - step names, commands - across the workflows
 - honour the version passed to install-clang.sh
 - restrict the scope of the workflows - before they would trigger on "README.md" "doc/*" and others
 
 The last commit adds a bit of duplication. If anyone has ideas how to avoid it, I'm all ears.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -22,5 +22,6 @@ jobs:
           pkg install -y meson pkgconf libdrm libXext libXfixes wayland
           pkg install -y -x '^mesa($|-libs)'
         run: |
-          meson _build
+          meson setup _build
           meson compile -C _build
+          meson install -C _build
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -14,8 +14,9 @@ jobs:
   freebsd:
     runs-on: macos-12
     steps:
-    - uses: actions/checkout@v3
-    - name: test
+    - name: 'Checkout'
+      uses: actions/checkout@v3
+    - name: 'Install prerequisites and build'
       uses: vmactions/freebsd-vm@v0
       with:
         prepare: |
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,14 +1,6 @@
 name: freebsd
 
-on:
-  push:
-    paths-ignore:
-    - '.github/workflows/**'
-    - '!.github/workflows/freebsd.yml'
-  pull_request:
-    paths-ignore:
-    - '.github/workflows/**'
-    - '!.github/workflows/freebsd.yml'
+on: [push, pull_request]
 
 jobs:
   freebsd:
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -16,14 +16,13 @@ jobs:
   style-check:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v3
-    - name: install prerequisites
+    - name: 'Checkout'
+      uses: actions/checkout@v3
+    - name: 'Install prerequisites'
       run: |
         sudo apt-get update
         sudo apt-get install -y \
           astyle
-    - name: style unify
-      run: ./style_unify
-    - name: check
-      run:  .github/workflows/style.sh
+    - name: 'Check for style changes'
+      run: ./style_unify && .github/workflows/style.sh
 
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -1,16 +1,6 @@
 name: style
 
-on:
-  push:
-    paths-ignore:
-    - '.github/workflows/**'
-    - '!.github/workflows/style.yml'
-    - '!.github/workflows/style.sh'
-  pull_request:
-    paths-ignore:
-    - '.github/workflows/**'
-    - '!.github/workflows/style.yml'
-    - '!.github/workflows/style.sh'
+on: [push, pull_request]
 
 jobs:
   style-check:
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -27,11 +27,12 @@ jobs:
       CC: ${{ matrix.compiler }}
       DISTRO: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v3
-    - name: install toolchain
+    - name: 'Checkout'
+      uses: actions/checkout@v3
+    - name: 'Install toolchain'
       if: ${{ (matrix.compiler == 'clang-15') }}
       run: .github/workflows/install-clang.sh 15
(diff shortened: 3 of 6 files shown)


evelikov:
@XinfengZhang @dvrogozh humble ping. Is there anything I can do to move this forward?

.github/workflows/freebsd.yml
```diff
@@ -5,22 +5,39 @@ on:
     paths-ignore:
     - '.github/workflows/**'
     - '!.github/workflows/freebsd.yml'
+    - 'doc/**'
```
dvrogozh, line 8: libva is pretty small project. I do not see the point to invest in such a list.

.github/workflows/freebsd.yml
```diff
@@ -5,22 +5,39 @@ on:
...
+    - name: 'Install prerequisites and build'
       uses: vmactions/freebsd-vm@v0
       with:
         prepare: |
           pkg install -y meson pkgconf libdrm libXext libXfixes wayland
           pkg install -y -x '^mesa($|-libs)'
         run: |
-          meson _build
-          meson compile -C _build
+          meson setup _build
```
dvrogozh, line 42: I need acked-by from @jbeich who helped to create this .yml and is maintainer for freebsd. Major question I have is: which way better reflects what freebsd is doing in its side? If older way is a better approximation, then I am not ok to do this change even if it follows the spirit of meson.

.github/workflows/style.yml
```diff
@@ -2,28 +2,29 @@ name: style
 
 on:
   push:
-    paths-ignore:
-    - '.github/workflows/**'
-    - '!.github/workflows/style.yml'
-    - '!.github/workflows/style.sh'
+    paths:
+    - '.github/workflows/style.*'
+    - 'style_unify'
```
dvrogozh, line 7: Same - no reason to invest.

.github/workflows/ubuntu.yml
```diff
@@ -6,11 +6,28 @@ on:
     - '.github/workflows/**'
     - '!.github/workflows/ubuntu.yml'
     - '!.github/workflows/install-clang.sh'
+    - 'doc/**'
```
dvrogozh, line 9: same

.github/workflows/ubuntu.yml
```diff
@@ -45,24 +63,18 @@ jobs:
...
-    - name: meson setup
-      if: ${{ (matrix.build == 'meson') }}
-      run: meson setup ./builddir --prefix=/usr
-    - name: meson compile
+    - name: 'Configure (meson)'
       if: ${{ (matrix.build == 'meson') }}
-      run: meson compile -C ./builddir || ninja -C ./builddir
-    - name: meson install
+      run: meson setup _build --prefix=/tmp/libva/usr
+    - name: 'Build and Install (meson)'
```
dvrogozh, line 71: I am not OK to collapse make and make install stages. I believe 2 stages approach better reflect what people are doing hence has more chances to catch issues.

.github/workflows/ubuntu.yml
```diff
@@ -45,24 +63,18 @@ jobs:
...
-    - name: configure
-      if: ${{ (matrix.build == 'autotools') }}
-      run: ./autogen.sh --prefix=/usr
-    - name: build
+    - name: 'Configure (autotools)'
       if: ${{ (matrix.build == 'autotools') }}
-      run: make
-    - name: install
+      run: ./autogen.sh --prefix=/tmp/libva/usr
+    - name: 'Build and Install (autotools)'
```
dvrogozh, line 78: same - don't collapse, please

dvrogozh (changes requested):

.github/workflows/freebsd.yml
evelikov: Can you be a little more specific here - are you concerned about the "meson" -> "meson setup" change, "meson compile" -> "meson install" or a mix?

In case you've missed it: running meson without "setup" will flag a warning.

.github/workflows/ubuntu.yml
evelikov: Sure thing - will split them with v2.

Out of curiosity: do you have any issues in mind, or this is more of a "just in case"?

.github/workflows/freebsd.yml
eli-schwartz: These changes aren't even relevant to freebsd specifically. They reflect running Meson on ***any*** operating system.

`meson _build` is implicitly rewritten to `meson setup _build` inside Meson's `main()` function, immediately before argument parsing. It is therefore obviously correct to not rely on that, but to use the official and documented command-line syntax of the `meson` command -- even without considering that Meson now warns you that implicit argument rewriting is deprecated.

Aside for that -- building with Meson is a 3-step process, roughly equivalent to the traditional autotools `./configure && make && make install`. Only for Meson, it is `meson setup _build/` instead of configure, and make invocations are replaced with `ninja`.

In both autotools and meson, running the "install" command will first make sure the build itself is up to date, so ***technically*** you can do this:

```
./configure && make install
```

But people usually don't. Just like they usually build with meson, before installing with meson.

Despite this, the freebsd tests did not test installing at all, only building. The commit message for this change states that the goal of this change is to test installing.

(It does that by collapsing both stages into one, which as I pointed out above, people do not generally do.)

evelikov: Thanks for the comprehensive reply Eli. It's quite reassuring hearing that from someone with deep knowledge of meson's internals.

jbeich: `meson setup` is fine. Likely caused by early examples predating `setup` support (before 0.42.0; `USES=meson` abstraction in FreeBSD appeared during 0.39.1) or based on outdated documentation e.g., https://github.com/mesonbuild/meson/commit/a0ce13d4e9bb

However, I don't like removing `meson compile` in favor of implicit dependency by `meson install`:
- build and install in separate phases match many downstream packaging systems and makes it easy to reason about in logs
- install phase is often a priveleged operation e.g., when installing system-wide for all users, so shouldn't happen by accident/bug during  build
- build may need to be adjusted with arguments not supported by install phase e.g., `--verbose` or `--jobs` (both are often used downstream)

.github/workflows/freebsd.yml
evelikov: Updated the PR dropping the `paths-ignore` section all together.

evelikov:
All the comment should be addressed now. Can you give it another look?

dvrogozh (approved):
looks good

dvrogozh:
@jbeich : this latest PR version has explicit `meson setup -> meson compile -> meson install` steps. I consider that's to be fine. If you think some adjustments are needed, let us know.

jbeich (approved):

XinfengZhang (approved):

PR #660, opened 2022-11-23
ci: Add documentation check workflow

evelikov:
We really don't build the docs in CI. Due to this a number of issues have crept in.

---
There is a bit of a caveat here - as-is the CI fails, although I don't have enough background knowledge to resolve the issues. Glancing around, some of the issues were introduced by the current maintainers - so I hope they can provide fixes.

### DIFF
--- a/.github/workflows/docs.exp
+++ b/.github/workflows/docs.exp
@@ -9,10 +9,6 @@ warning: Tag 'MSCGEN_PATH' at line 2178 of file 'doc/Doxyfile' has become obsole
 /home/runner/work/libva/libva/va/va.h:5270: warning: end of file with unbalanced grouping commands
 /home/runner/work/libva/libva/va/va_enc_hevc.h:795: warning: unbalanced grouping commands
 /home/runner/work/libva/libva/va/va_enc_hevc.h:893: warning: unbalanced grouping commands
-/home/runner/work/libva/libva/va/va_fei.h:169: warning: end of file while inside a group
-/home/runner/work/libva/libva/va/va_fei_h264.h:505: warning: end of file while inside a group
-/home/runner/work/libva/libva/va/va_fei_hevc.h:252: warning: end of file while inside a group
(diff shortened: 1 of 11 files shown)


dvrogozh (changes requested):
Well, we are not ready to check for warnings/errors from doxygen at the moment. There is really a bunch of problems and  we will need some time to fix.

So, let's break this into stages. I hope #661 will be the first one to add "fatal" docs build ci and addressing few high level trivial problems.

evelikov:
A slightly better idea came to mind, inspired by Mesa:
 - store the expectation in-tree
 - fail the pipeline if the output changes - fixes need to update the expectations, new changes must not add errors

Will update the MR, to include your fixes as an illustration.

evelikov:
If the documentation is going through refactor (like the last patch), the approach seems a little noisy. But overall something everyone can live with in the short to mid term.

evelikov:
@dvrogozh what is your view on the updated approach?

dvrogozh (changes requested):
> @dvrogozh what is your view on the updated approach?

We don't have that many errors from doxygen actually to invent the regression checker. I think we need to attempt fix them. As such I suggest to go step by step and first merge in #661 which adds minor fixes + adds a build path check.

evelikov:
> Overall I agree that fixing the issues would be the better approach. Then again it goes into the "perfect is the enemy of good" category.
> 
> Regardless, say that we fix have a fix for all of the issues - how can we ensure, in automated way, that new issues don't get introduced?
> 
> The proposed regression tracker trivially handles that, while also helping us to get there. Doesn't it?

@dvrogozh does the above make sense?

evelikov:
@dvrogozh @XinfengZhang humble poke - is there anything I can to do move this forward?

evelikov:
Ping?

XinfengZhang:
looks there are different opinions , could we merge #681  as first step, then continue the discussion of  the next step ? we could decouple them ...

PR #664, opened 2022-12-14
add VAProfileH264High10

gitgjogh:
Signed-off-by: jianfeng.zheng <jianfeng.zheng@mthreads.com>

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -530,7 +530,8 @@ typedef enum {
     VAProfileAV1Profile1                = 33,
     VAProfileHEVCSccMain444_10          = 34,
     /** \brief Profile ID used for protected video playback. */
-    VAProfileProtected                  = 35
+    VAProfileProtected                  = 35,
+    VAProfileH264High10                 = 36
 } VAProfile;
 
 /**
--- a/va/va_str.c
+++ b/va/va_str.c
@@ -37,6 +37,7 @@ const char *vaProfileStr(VAProfile profile)
         TOSTR(VAProfileMPEG4Main);
         TOSTR(VAProfileH264Main);
         TOSTR(VAProfileH264High);
+        TOSTR(VAProfileH264High10);
         TOSTR(VAProfileVC1Simple);
         TOSTR(VAProfileVC1Main);
         TOSTR(VAProfileVC1Advanced);
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -5749,6 +5749,7 @@ void va_TraceRenderPicture(
                 va_TraceMPEG4Buf(dpy, context, buffers[i], type, size, num_elements, pbuf + size * j);
             }
             break;
+        case VAProfileH264High10:
         case VAProfileH264Main:
         case VAProfileH264High:
         case VAProfileH264ConstrainedBaseline:


va/va.h
```diff
@@ -530,7 +530,14 @@ typedef enum {
...
     VAProfileHEVCSccMain444_10          = 34,
     /** \brief Profile ID used for protected video playback. */
-    VAProfileProtected                  = 35
+    VAProfileProtected                  = 35,
+
+    /**
+     * @note 'VA_PROFILE_H264_HIGH_10' is an conditional compiling option
+     * for 'VAProfileH264High10' support.
+     */
+#define VA_PROFILE_H264_HIGH_10
```
XinfengZhang, line 539: I dont think that libva need add such macro,  and it is always defined.  please remove it. if driver want it, it should be defined inside driver.

XinfengZhang (changes requested):
I believe we dont need the VA_PROFILE_H264_HIGH_10, please remove it, driver could determine whether report such profile.  
and , is it convenient  to share which driver will support high 10 profile, enc or dec?

va/va.h
gitgjogh: It was original wish to let mediadriver and applications such as ffmpeg support this new feature while still make backward compatibility with old version of libva.

Never mind，I can wait the new release and then merge corresponding code to FFmpeg by using 'VA_CHECK_VERSION'

evelikov:
Do we have an idea which drivers will support this profile - any of the Intel ones (i965, iHD), mesa (radeonsi, nouveau) or some other?

More importantly - I think we need [to update the tests](https://github.com/intel/libva-utils), right?

gitgjogh:
> Do we have an idea which drivers will support this profile - any of the Intel ones (i965, iHD), mesa (radeonsi, nouveau) or some other?

I'm now working for Moore Threads. [Moore Threads](https://www.mthreads.com) (short for Mthreads) is a Chinese GPU manufacturer. Our newest products, MTTS80/MTTS2000/MTT3000, all has H264High10 enc/dec capabilities at max 8k resolution.

Beside H264High10, our produncts also support [AVS+ and AVS2 video](http://www.avs.org.cn/avsStandard.asp) decoding that are now all out of VAAPI definition. And we also wish to add them to VAAPI.

> More importantly - I think we need [to update the tests](https://github.com/intel/libva-utils), right?

For H264High10, since only enum VAProfileHigh10 has been added, I think it may not need to test for LIBVA. Rather, MediaDriver and uplayer application like gstreamer or FFMpeg need to be test. We actually has  bring up both decoding and encoding wok flow under FFmpeg and test under vaapi-fits.

evelikov:
The test suite in libva-utils will also need some very small changes. See my comment in https://github.com/intel/libva-utils/pull/274 for some pointers.

Adding ffmpeg and vaapi-fits support sounds amazing, thank you \o/

gitgjogh:
hi evelikov,

[intel/libva-utils#298](https://github.com/intel/libva-utils/pull/298) has been created.

gitgjogh:
Hi their,

Corresponding ffmpeg patch has been commit, see ffmpeg-devel mail list

http://ffmpeg.org/pipermail/ffmpeg-devel/2022-December/305100.html

or

https://github.com/FFmpeg/FFmpeg/compare/master...gitgjogh:FFmpeg:VAProfileH264High10

Thanks

XinfengZhang (approved):
LGTM now, thanks

gitgjogh:
rebased

XinfengZhang ***@***.***> 于2023年1月5日周四 10:23写道：

PR #666, opened 2022-12-19
Rewrite driver registry query and fix some bugs/leaks/inefficiencies

jenatali:
Instead of enumerating all adapters, just open the one we want by LUID. Also close the adapter, and consolidate error handling.

### DIFF
--- a/va/win32/va_win32.c
+++ b/va/win32/va_win32.c
@@ -277,6 +277,10 @@ VADisplay vaGetDisplayWin32(
     pDisplayContext->vaGetDriverNameByIndex = va_DisplayContextGetDriverNameByIndex;
     pDisplayContext->vaGetNumCandidates = va_DisplayContextGetNumCandidates;
     pDisplayContext->opaque = calloc(1, sizeof(VADisplayContextWin32));
+    if (!pDisplayContext->opaque) {
+        va_DisplayContextDestroy(pDisplayContext);
+        return NULL;
+    }
 
     VADisplayContextWin32* pWin32Ctx = (VADisplayContextWin32*) pDisplayContext->opaque;
     if (adapter_luid) {
--- a/va/win32/va_win32.c
+++ b/va/win32/va_win32.c
@@ -35,20 +35,6 @@
  * does not have a registered VA driver
 */
 const char VAAPI_DEFAULT_DRIVER_NAME[] = "vaon12";
-const char VAAPI_REG_SUB_KEY[] = "VAAPIDriverName";
-const char VAAPI_REG_SUB_KEY_WOW[] = "VAAPIDriverNameWow";
-
-inline const char* GetVAAPIRegKeyName()
-{
-#ifdef _WIN64
-    return VAAPI_REG_SUB_KEY;
-#else
-    BOOL is_wow64 = FALSE;
-    if (IsWow64Process(GetCurrentProcess(), &is_wow64) && is_wow64)
-        return VAAPI_REG_SUB_KEY_WOW;
-    return VAAPI_REG_SUB_KEY;
-#endif
-}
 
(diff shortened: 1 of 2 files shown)


sivileri:
@XinfengZhang @dvrogozh could you please take a look?

jenatali:
Commits squashed.

dvrogozh (approved):
looks good

dvrogozh (commented):
Sorry, one more comment, can you, please:
1. Change commit title a little exposing that it's for win32. For example:
```
win32: rewrite driver registry query and fix some bugs/leaks/inefficiencies
```
2. Add a `Signed-off-by: name <email>`

Thank you.

jenatali:
> 

Done.

dvrogozh (approved):
LGTM

va/win32/va_win32.c
```diff
@@ -58,136 +44,76 @@ typedef struct _VADisplayContextWin32 {
 
 static bool TryLoadDriverNameFromRegistry(const LUID* adapter_luid, VADisplayContextWin32* pWin32Ctx)
 {
-    if (!adapter_luid)
+    if (!adapter_luid || !pWin32Ctx)
```
evelikov, line 47: Neither of of these can be NULL - so I would omit this change.

If anything, the former argument can be dropped. since it's accessible via the latter. Also I would s/Try// and fix the type handling - return boot, which gets stored as int masked bool.

Assuming you agree - the above can be a separate change, in this or separate PR.

va/win32/va_win32.c
```diff
@@ -58,136 +44,76 @@ typedef struct _VADisplayContextWin32 {
...
-                goto out;
-            }
-            free(pQueryBuffer);
-        }
-out:
-        free(EnumAdapters.pAdapters);
+    PFND3DKMT_OPENADAPTERFROMLUID pfnOpenAdapterFromLuid = (PFND3DKMT_OPENADAPTERFROMLUID)GetProcAddress(hGdi32, "D3DKMTOpenAdapterFromLuid");
+    PFND3DKMT_CLOSEADAPTER pfnCloseAdapter = (PFND3DKMT_CLOSEADAPTER)GetProcAddress(hGdi32, "D3DKMTCloseAdapter");
+    PFND3DKMT_QUERYADAPTERINFO pfnQueryAdapterInfo = (PFND3DKMT_QUERYADAPTERINFO)GetProcAddress(hGdi32, "D3DKMTQueryAdapterInfo");
+    if (!pfnOpenAdapterFromLuid || !pfnCloseAdapter)
```
evelikov, line 77: `... || !pfnQueryAdapterInfo)` right?

evelikov (commented):
Left you a couple of small suggestions - overall the updated code is much cleaner and saner.

HTH o/

va/win32/va_win32.c
jenatali: Yeah, good catch.

va/win32/va_win32.c
dvrogozh: Actually, `pWin32Ctx` can be NULL because we forgot to check result of `calloc` here:

https://github.com/intel/libva/blob/0d971e918da2af176f158cad3948b1c88db3d55b/va/win32/va_win32.c#L279

We need to add a check for `calloc` operation. After that I agreed that check here won't make sense anymore.

dvrogozh (changes requested):

va/win32/va_win32.c
jenatali: Added the null check in its own commit, and addressed the rest of the feedback from this comment in a followup commit.

dvrogozh (approved):
looks good

XinfengZhang (approved):

PR #671, opened 2022-12-20
ci: enable Werror across the build

evelikov:
The autotool build(s) already have that, alas the meson ones do not. Use the upstream recommended -D werror=true option.

Signed-off-by: Emil Velikov <emil.l.velikov@gmail.com>

### DIFF
--- a/va/win32/va_win32.c
+++ b/va/win32/va_win32.c
@@ -59,7 +59,7 @@ static void LoadDriverNameFromRegistry(VADisplayContextWin32* pWin32Ctx)
 #ifndef _WIN64
     BOOL isWowProcess = false;
     if (IsWow64Process(GetCurrentProcess(), &isWowProcess) && isWowProcess)
-        wcscpy(RegistryInfo.ValueName, "VAAPIDriverNameWow");
+        wcscpy(RegistryInfo.ValueName, L"VAAPIDriverNameWow");
 #endif
     D3DKMT_QUERYADAPTERINFO QAI = {
         .Type = KMTQAITYPE_QUERYREGISTRY,
--- a/va/win32/va_win32.c
+++ b/va/win32/va_win32.c
@@ -107,7 +107,10 @@ cleanup:
         free(pRegistryInfo);
     if (pfnCloseAdapter && OpenArgs.hAdapter) {
         D3DKMT_CLOSEADAPTER Close = { OpenArgs.hAdapter };
-        pfnCloseAdapter(&Close);
+        /* The explicit negation is a no-op, yet required to silence the
+         * Wunused-result warning.
+         */
+        (void) !pfnCloseAdapter(&Close);
     }
     FreeLibrary(hGdi32);
 }
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -15,6 +15,6 @@ jobs:
           pkg install -y meson pkgconf libdrm libXext libXfixes wayland
           pkg install -y -x '^mesa($|-libs)'
         run: |
-          meson setup _build
+          meson setup _build -D werror=true
           meson compile -C _build
           meson install -C _build
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,7 +18,7 @@ jobs:
       run: '.github\workflows\EnterDevShell.ps1 ${{ inputs.architecture }}'
       shell: pwsh
     - name: 'Configure with meson'
-      run: meson setup _build
+      run: meson setup _build -D werror=true
     - name: 'Build'
       run: meson compile -C _build
     - name: 'Install'
@@ -46,7 +46,7 @@ jobs:
       run: '.github\workflows\EnterDevShell.ps1 ${{ inputs.architecture }}'
       shell: pwsh
     - name: 'Configure'
-      run: meson setup _build
+      run: meson setup _build -D werror=true
     - name: 'Build'
       run: meson compile -C _build
     - name: 'Install'


evelikov:
@jenatali @sivileri seems like https://github.com/intel/libva/pull/666 has introduced a couple of small issues

va/win32/va_win32.c
```diff
@@ -59,7 +59,7 @@ static void LoadDriverNameFromRegistry(VADisplayContextWin32* pWin32Ctx)
 #ifndef _WIN64
     BOOL isWowProcess = false;
     if (IsWow64Process(GetCurrentProcess(), &isWowProcess) && isWowProcess)
-        wcscpy(RegistryInfo.ValueName, "VAAPIDriverNameWow");
+        wcscpy(RegistryInfo.ValueName, L"VAAPIDriverNameWow");
```
jenatali, line 62: Oof, yeah I didn't build x86, that's my bad.

evelikov: Heh - we've all been there. That's why the CI use werror=true

sivileri:
> @jenatali @sivileri seems like #666 has introduced a couple of small issues

CC @dvrogozh @XinfengZhang patched on this change with commits 75b6cca45cb4c1fe054e9946f9a1b8a7bb0b00d3 and fdcca6f020e9cf426cc354c9d6c7b70177309a38

dvrogozh (changes requested):

.github/workflows/ubuntu.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       run: ${{ matrix.compiler }} --version
     - name: 'Configure (meson)'
       if: ${{ (matrix.build == 'meson') }}
-      run: meson setup ./builddir --prefix=/usr
+      run: meson setup ./builddir --prefix=/usr -D werror=true
```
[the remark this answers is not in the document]
evelikov: AFAICT this does now work with MSVC since the `-Werror` style is gcc specific. In other words even if used with clang (in msvc mode) this won't trigger.

So I would rather stay consistent across the meson board and use the same flag. Considering the shift away from autotools and the (pre-existing bugs in autotools) I seriously won't bother, and would rather use the proper style.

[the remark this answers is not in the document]
eli-schwartz: On visual studio style compilers (MSVC, clang-cl) the Meson setting `-Dwerror=true` is translated to the compiler flag `/WX`. This is a good proof that Meson should indeed be in charge of this, but if you need further proof, there are a variety of other things it could be, depending both on the compiler, and the programming language:
- `--fatal-warnings`
- `--emit_warnings_as_errors`
- `-D warnings`
- `-w`
- `-Werror=cross-execution-space-call,deprecated-declarations,reorder`
- `-warnaserror`
- `-change_message=error`


I suppose however, that you're not using ccrx, ti, or xc16 branded C compilers in CI, so the only practical issue you run into for portability, is indeed MSVC and clang-cl.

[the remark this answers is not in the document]
dvrogozh: If meson has exposed a generalized option across a board to handle a setting which is different across different compilers - that is a good thing. However, I do not see why I can not use a practice to specify additional compiler options (thru `CFLAGS`) which I used for 2 decades and which so far has worked for me? Especially why I can't use it for the particular CI case with particular compilers (gcc/clang) running on particular OS (ubuntu)? Moreover, if that's the way how most people would specify additional compiler flags, it does make sense to me to check in CI that project correctly handles such external specification (and I did see far too often when projects do break this convention). That's the reason I personally prefer to leave CFLAGS for our ubuntu CI case with gcc/clang.

By the way, a question. What's meson "recommended" way to specify arbitrary compiler/linker flags? Is that thru CFLAGS or there is some other better way?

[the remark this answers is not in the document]
eli-schwartz: > However, I do not see why I can not use a practice to specify additional compiler options (thru `CFLAGS`) which I used for 2 decades and which so far has worked for me?

You can, and Meson will respect that by assuming you know exactly which flags you need.

But... the flags you specified do not work on MSVC. :)



> By the way, a question. What's meson "recommended" way to specify arbitrary compiler/linker flags? Is that thru CFLAGS or there is some other better way?

Meson respects the standard CFLAGS for arbitrary compiler/linker flags.

Some things, Meson has builtin support for via Meson options, which you can use as well -- and if Meson has a builtin option for it, that option will be cross-platform, rather than having to specify one set of CFLAGS for GCC and another set for MSVC.

However, you *absolutely* can specify `CFLAGS=-Werror` on GCC and `CFLAGS=/WX` on MSVC, if you prefer.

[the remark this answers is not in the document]
evelikov: Must admit I did not expect such a trivial change to cause that much of a discussion. Above all - thanks Eli for providing insights from meson's POV.

@dvrogozh onto your points:
 - meson will honour CFLAGS, so you can still use them but as mentioned that will not work for other people's use-cases
 - to meson or autotools - as hinted in the past, people and more importantly distributions want to move away from the latter. the fact that FreeBSD and Windows use the former is a clear indication, especially since autotools just works with them

The second is a yak shaving which should be left separate IMHO. Will address your other comment in a bit.

evelikov:
MR updated - the `CFLAGS` piece is dropped since the meson equivalent already covers much broader range (both code and platforms).

evelikov:
@dvrogozh @XinfengZhang team you take another look at this MR. Thanks o/

.github/workflows/ubuntu.yml
```diff
@@ -2,9 +2,6 @@ name: ubuntu
 
 on: [push, pull_request]
 
-env:
```
dvrogozh, line 5: I think this and next commit where you add `-D werror=true` need to be squashed. Otherwise current commit which drops CFLAGS basically regress experience.

Unfortunately combination of this and next commit still causes regression in experience for autotools build since you drop `-Werror` and don't add it back.

Also, I am highly unsure that `-Wall` is getting added by default in both autools/meson scripts. As a result you basically reduce set of warnings we monitor.

dvrogozh (changes requested):

.github/workflows/ubuntu.yml
evelikov: Sure can squash.

Don't fully understand your concern with "regression" - all the code that's being build with autoconf is already build with meson - on even more platforms.

Does that make sense or am I simply too tired and misread the code?

eli-schwartz: It's also safe to just reverse the order... If -Werror is passed twice, once by meson and once by CFLAGS, that's redundant but harmless and the follow-up commit can be marketed as a mere cleanup.

...

Meson definitely defaults to -Wall, unless you specify a non-default `warning_level=0` option.

evelikov: @dvrogozh for meson we use `warning_level=1`  which gives us `-Wall` plus a bit more. On the configure side - _no_ compiler options are passed at all. Even so, as I said earlier meson already compiles the autoconf covered code so we should be all good.

Happy to make changes though - just clearly put your suggestion.

dvrogozh: This PR started as a fix for Windows warning and Windows CI where we missed `-Werror`. I suggest to fix exactly what is being missed - Windows part plus FreeBSD part where we also miss the flag. If you want to use meson way of fixing that - I am all fine.

However, I am not ok with the change to Linux part of the ci where we already had `-Werror` and where you are dropping it for autotools part. I do not believe that's a right thing to do. Autotools build path is still a way to go for a number of people and I do want this to get tested. I am not aligned with the motivation that we already test this with meson. Maybe or maybe not - it's much easier to just test autotools with `Werror` and be sure w/o any time spent on arguing. Same goes with `-Wall` - I believe it's reasonable to get build tested with all the warnings enabled. Current way of testing explicitly enables that on CI level. Just leave that as is.

Summarizing Linux part - I would prefer to leave it as-is without any modifications and just do Windows & freebsd.

evelikov: Thanks for elaborating. Your earlier suggestion "we can have one but not both" was pointing at another direction.

Kept the `CFLAGS` piece and dropped the meson `-D error=true` from the Linux/Ubuntu build. Hope that's better

dvrogozh (approved):
looks good now

XinfengZhang (approved):

PR #672, opened 2022-12-21
Remove reference to non-existent symbol

tamird:
This fixes compilation in recent clang versions (tested against https://github.com/llvm/llvm-project/commit/6443c0e) which is apparently stricter with respect to undefined symbols. It appears that this version script reference was never correct.

### DIFF
--- a/va/libva.syms
+++ b/va/libva.syms
@@ -6,6 +6,4 @@ VA_API_0.32.0 {
 VA_API_0.33.0 {
     global:
         vaCreateSurfaces;
-    local:
-        vaCreateSurfaces_0_33_0;
 } VA_API_0.32.0;


dvrogozh (approved):
That was done in 6febbcad which introduced vaCreateSurfaces_0_32_0 (which has real definition and usage) and vaCreateSurfaces_0_33_0 (which is mentioned only in a .syms file). It does not make sense to handle a symbol which is not actually defined. This makes change LGTM to me. @XinfengZhang : anything I miss?

tamird:
A possible cause: https://reviews.llvm.org/D135402.

XinfengZhang (approved):

PR #680, opened 2023-02-01
win32: Add vaDisplayIsValid to .def export list

sivileri:
vaDisplayIsValid is used by apps calling into VAAPI, and we were missing it from the exports list.

Example of usage on [this app source](https://github.com/FFmpeg/FFmpeg/blob/7158f1e64d9b76afea78537a35c465447df0cff8/libavutil/hwcontext_opencl.c#L778): 

### DIFF
--- a/va/libva.def
+++ b/va/libva.def
@@ -83,3 +83,4 @@ EXPORTS
     vaSetSubpictureImage
     vaStatusStr
     vaSyncSurface2
+    vaDisplayIsValid


sivileri:
FYI @dvrogozh @XinfengZhang

va/libva.def
```diff
@@ -83,3 +83,4 @@ EXPORTS
     vaSetSubpictureImage
     vaStatusStr
     vaSyncSurface2
+    vaDisplayIsValid
```
dvrogozh, line 86: LGTM considering this function is part of VAAPI public API:
https://github.com/intel/libva/blob/b75818c092367723c76e716c56d3ef9765ce9e51/va/va.h#L445

dvrogozh (approved):

va/libva.def
sivileri: @dvrogozh thanks for taking a look! Could a maintainer please help merging it? I don't seem to have enough permissions to do it myself.

image

dvrogozh: I am just giving a chance for others to review. I will merge it in tomorrow if no concerns from others unless @XinfengZhang won't do that before me.

XinfengZhang (approved):

va/libva.def
sivileri: Sounds good thanks :)

PR #681, opened 2023-02-03
win32/loader: Allow backends to provide custom driver search paths

sivileri:

VA changes:

```
Add a new optional backend function vaGetDriverSearchPathByIndex.
The vaInitialize function will now query the backend (the vaGetDriverSearchPathByIndex not null)
for extra search paths and combine them with the previously usual 
driver_search_path that comes either from LIBVA_DRIVERS_PATH or the VA_DRIVERS_PATH constant.
```

Win32 changes:

```
- Implement vaGetDriverSearchPathByIndex to provide the
search driver absolute path for the registered driver

- Change va_DisplayContextGetDriverNameByIndex to extract the
driver name (minus the drv_video prefix) from the absolute path
in the registry key string.

- Provide a .\ search path for the default vaon12
driver for apps to avoid having to set LIBVA_DRIVERS_PATH
for the default.

```

### DIFF
--- a/va/win32/va_win32.c
+++ b/va/win32/va_win32.c
@@ -35,6 +35,7 @@
  * does not have a registered VA driver
 */
 const char VAAPI_DEFAULT_DRIVER_NAME[] = "vaon12";
+const char VAAPI_DEFAULT_DRIVER_SEARCH_PATH[] = ".";
 
 typedef struct _VADisplayContextWin32 {
     LUID adapter_luid;
@@ -165,7 +166,7 @@ static VAStatus va_DisplayContextGetDriverNameByIndex(
         /* Always prefer the adapter registered driver name as first option */
         if (candidate_index == 0) {
             *driver_name = calloc(sizeof(pWin32Ctx->registry_driver_name), sizeof(char));
-            memcpy(*driver_name, pWin32Ctx->registry_driver_name, sizeof(pWin32Ctx->registry_driver_name));
+            va_win32_get_drv_from_path(pWin32Ctx->registry_driver_name, *driver_name, sizeof(pWin32Ctx->registry_driver_name));
         }
         /* Provide the default driver name as a fallback option */
         else if (candidate_index == 1) {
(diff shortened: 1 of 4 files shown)


sivileri:
CC @jenatali @dvrogozh @XinfengZhang

va/win32/va_win32.c
```diff
@@ -181,6 +182,39 @@ static VAStatus va_DisplayContextGetDriverNameByIndex(
...
+        /* Only provide the registry search path if there is no driver name override */
+        VADriverContextP ctx = CTX((VADisplay)pDisplayContext);
+        if ((candidate_index == 0) && !getenv("LIBVA_DRIVER_NAME") && !ctx->override_driver_name) {
+            /* Always prefer the adapter registered driver search path as first option */
+            char* last_separator = strrchr(pWin32Ctx->registry_driver_name, '\\');
+            if (last_separator) {
+                /* Always prefer the adapter registered driver search path as first candidate */
+                *driver_search_path = calloc(sizeof(pWin32Ctx->registry_driver_name), sizeof(char));
+                memcpy(*driver_search_path, &pWin32Ctx->registry_driver_name[0], last_separator - &pWin32Ctx->registry_driver_name[0]);
+            }
```
jenatali, line 203: If there's no separator, nothing gets returned?

jenatali (commented):
Personally, I think this is a bit overly complex. Getting an absolute path, only to break it up into path, part of the driver name, and then an assumed suffix, seems... wrong. Can't we just have a path in the loader where we support loading a full path instead of doing this dance?

evelikov:
I cannot find any mention of `VAAPIDriverName` or the absolute path in the link provided. Is it the correct one? Who is responsible for ensuring there are no issues like path traversal attacks - some Windows component or libva itself? More importantly - is it possible to change the restrictions to follow the Linux path/suffix style?

If absolute paths is the only way forward, I'm inclined a tiny `#ifdef _WIN32` hunk is the cleaner way forward. Adding the extra code seems like unnecessary over complication to me.

jenatali:
> I cannot find any mention of `VAAPIDriverName` or the absolute path in the link provided. Is it the correct one?

Right, that link is about OpenGL, but the same principle should apply for VA. Back in the day, drivers used to put their DLLs in System32 (i.e. automatic DLL search path) and just put a filename in the registry, but beginning with early versions of Win10, the DLLs no longer go into System32, and the registry now stores a full path, a design choice called "run from driver store."

> Who is responsible for ensuring there are no issues like path traversal attacks - some Windows component or libva itself?

Like with other ICD-based loaders, the registry is treated as trusted. Once a path is pulled out of it, it's assumed that the path is one that's actually meant to be used.

> More importantly - is it possible to change the restrictions to follow the Linux path/suffix style?

No, that'd be going backwards. When a driver is installed on Windows, the amount of global state they're allowed to apply is incredibly minimal, with a goal from the Windows drivers team to push that toward zero. Instead, all state is isolated to a per-driver store, both on disk and in the registry. Since the driver DLLs cannot end up in a place that is easily searched, the DLL lookup procedure is to go from device -> registry -> DLL path.

> If absolute paths is the only way forward, I'm inclined a tiny `#ifdef _WIN32` hunk is the cleaner way forward. Adding the extra code seems like unnecessary over complication to me.

I think Sil's got an alternate changeset he's working on after some offline discussion last week, but I'll let him go into more details.

evelikov:
@jenatali thanks for the explanation. Would it make sense to add at least some of that into the official MS website? Having all this, fairly important stuff on a random Github MR seems a bit weird.

sivileri:
Closing in favor of new change https://github.com/intel/libva/pull/684

Issue #677, opened 2023-01-18
2.17 brakes hardware video decoding in Chromium


gladykov:
libva 2.17 breaks hw video decode with or without vulkan on Chromium

See https://bbs.archlinux.org/viewtopic.php?pid=2080054#p2080054

Accoring to author this commit broke it: https://github.com/intel/libva/commit/ef1df02f3ad45ac98b1fa57c373176d7c14dcc57

shtirlic:
Also here https://bugs.chromium.org/p/chromium/issues/detail?id=1407223#c10

mtak-misc:
There will be a difference in the output of vainfo in my environment. 

If https://github.com/intel/libva/commit/ef1df02f3ad45ac98b1fa57c373176d7c14dcc57  is enabled, "libva error: /usr/lib/dri/iHD_drv_video.so init failed" will be displayed in the following way.
```
Trying display: wayland
Trying display: x11
libva error: /usr/lib/dri/iHD_drv_video.so init failed
vainfo: VA-API version: 1.17 (libva 2.17.1)
vainfo: Driver version: Intel i965 driver for Intel(R) Sandybridge Mobile - 2.4.1
vainfo: Supported profile and entrypoints
      VAProfileMPEG2Simple            : VAEntrypointVLD
      VAProfileMPEG2Main              : VAEntrypointVLD
      VAProfileH264ConstrainedBaseline: VAEntrypointVLD
      VAProfileH264ConstrainedBaseline: VAEntrypointEncSlice
      VAProfileH264Main               : VAEntrypointVLD
      VAProfileH264Main               : VAEntrypointEncSlice
      VAProfileH264High               : VAEntrypointVLD
      VAProfileH264High               : VAEntrypointEncSlice
      VAProfileH264StereoHigh         : VAEntrypointVLD
      VAProfileVC1Simple              : VAEntrypointVLD
      VAProfileVC1Main                : VAEntrypointVLD
      VAProfileVC1Advanced            : VAEntrypointVLD
      VAProfileNone                   : VAEntrypointVideoProc
```

If  https://github.com/intel/libva/commit/ef1df02f3ad45ac98b1fa57c373176d7c14dcc57  is disabled, "libva error: /usr/lib/dri/iHD_drv_video.so init failed" will not be displayed in the following way.
Even when using libva 2.16.0, this message will not be displayed.
```
Trying display: wayland
Trying display: x11
vainfo: VA-API version: 1.17 (libva 2.17.1)
[...]

PR #682, opened 2023-02-06
Rework device names

evelikov:
The current GetDriver API is rather fragile and bonkers.

In particular it does a two-step pass - number vs names. That in itself
is not great, since it assumes the devices do not change in between.
Additionally there is the assumption of index X will match device (name)
X, which does not hold true.

The most prominent issue is the inability to provide accurate number and
list of names.

Instead, we can have a single API, that provides both pieces of data in
one do - shorter and simpler code - in reliable manner.

---
Note: I haven't fixed the reserved size with [va: add vaGetDriverNames() internal ABI](https://github.com/intel/libva/commit/d9c4d4c449d5d965f62cb58566e3b9a6b47d890a) - not 100% sure 

This is the first step towards fixing https://github.com/intel/libva/issues/677


evelikov:
@sivileri this MR reworks the driver-name API, conflicting with your other work. Would be great if we can merge this first, but at the least can you please check this doesn't explode on Windows. Thanks o/

@XinfengZhang @dvrogozh this is the first big step towards addressing the DRI3 regressions some people are seeing. Would be great if we can get to it sooner than later.

ceyusa:
fwiw, it looks cleaner and robust.

What worries me is that drivers lists are scattered among different backend implementation files, and, iiuc, some drivers can be used by several backends. I wonder if we could centralized all of them in a single file so we could update that single structure for all at once. Anyway, this change can be done later.

va/win32/va_win32.c
```diff
@@ -59,7 +60,7 @@ static void LoadDriverNameFromRegistry(VADisplayContextWin32* pWin32Ctx)
 #ifndef _WIN64
     BOOL isWowProcess = false;
     if (IsWow64Process(GetCurrentProcess(), &isWowProcess) && isWowProcess)
-        wcscpy(RegistryInfo.ValueName, "VAAPIDriverNameWow");
+        wcscpy(RegistryInfo.ValueName, L"VAAPIDriverNameWow");
```
ceyusa, line 63: afaik, this is already merged in https://github.com/intel/libva/pull/671/files#diff-d8b70b73203726f1dcdd2908e543118f1353b61a50515196eb294e8c63689ba9 right?

evelikov:
> fwiw, it looks cleaner and robust.
> 
> What worries me is that drivers lists are scattered among different backend implementation files, and, iiuc, some drivers can be used by several backends. I wonder if we could centralized all of them in a single file so we could update that single structure for all at once. Anyway, this change can be done later.

IMHO we should probably spend our efforts into a proper mechanism (say something like the Vulkan ICD/json) to remove this hard-coding + add some order/priority guarantees. Which in itself is an important yet orthogonal task.

va/win32/va_win32.c
evelikov: Yup fixed. Also added the WSL bits that landed recently.

evelikov:
@XinfengZhang @dvrogozh any comments?

dvrogozh (changes requested):
So, proposal is to replace 2 functions currently used to enumerate drivers:

evelikov:
Rebased, fixed the reserved bits and added one line summary about the API change.

@dvrogozh can you have another look?

dvrogozh (commented):
Change looks good to me. Basically I am trying to imagine the case where something might go wrong, but without a success so far. I think this change needs to be extensively tried before the merge. Recommend to pick it up for validation. @XinfengZhang - please, take a look.

XinfengZhang:
the change itself looks good, but from my perspective, we do need test it.
also, we could merge it after 2.18 release immediately, to keep it in master about 3 month to be tried by different users.

evelikov:
@XinfengZhang when you say "we need to test it" can you elaborate who is "we" and what testing do you expect to be done?

There has been zero comments here for over 3 months and 2.18 was released 2 month ago. This makes me wonder if as-is libva is even considered maintained?

va/va.c
```diff
@@ -741,6 +743,72 @@ static VAStatus va_legacy_opendriver(VADisplay dpy)
...
 }
 
+static VAStatus va_new_opendriver(VADisplay dpy)
+{
+    VADisplayContextP pDisplayContext = (VADisplayContextP)dpy;
+    /* In the extreme case we can get up-to 5ish names. Pad that out to be on
+     * the safe side. In the worst case, the DRIVER BUG below will print and
+     * we'll be capped to the current selection.
+     */
+    char *drivers[20] = { 0, };
```
XinfengZhang, line 657: 20 should be enough now, but does we need a check?

evelikov: What check are you looking for - can you provide a concrete example?

XinfengZhang: no need check now, it is impossible that {i915, {iHD, i965, i966 .... iHD1, iHD2...} exceed 20 backend UMD drivers for one KMD.

evelikov:
@XinfengZhang do you mind replying to my earlier question https://github.com/intel/libva/pull/682#issuecomment-1564281748, thanks.

XinfengZhang:
> @XinfengZhang do you mind replying to my earlier question [#682 (comment)](https://github.com/intel/libva/pull/682#issuecomment-1564281748), thanks.

sorry for late response firstly, just like my previous comments, the change itself looks good to me 
https://github.com/intel/libva/pull/682#issuecomment-1467608032
but that time , we are closing 2.18 release. so I want to keep it there and let more people try it. 
and merge it after 2.18 release. but there are no more comments after that, 
I got the notification from your comments of last week. and revisit it and try it again.
I apologize for this , I understand it is bad practice to a respected contributor, will avoid such bad practice in future

about why we did not include it in 2.18. 
the code make the code more elegant ,  unify 2 steps to 1. it is very nice code,
but it is not urgent feature at that time, and it include more 19 commit, looks a little risky at that time. 
and the Q1 release validation cycle already been triggered. 

I tested it with intel media stack.

evelikov:
While I appreciate the heartfelt apology, I am a firm believer that actions speak louder than words. [Here's to](https://dictionary.cambridge.org/dictionary/english/here-s-to) this being the last instance.

MR has been updated and lightly tested - basically a simple s/getenv/secure_getenv/ change was required.

XinfengZhang:
hi @dvrogozh , could you help to dismiss your changes request if your concern was addressed already and no further concerns. 
I will merge it , thanks

dvrogozh (approved):
Looks good.

va/va.c
```diff
@@ -719,15 +661,78 @@ VAStatus vaSetDriverName(
...
+    }
+
+    ctx = CTX(dpy);
+    driver_name_env = secure_getenv("LIBVA_DRIVER_NAME");
+
+    if ((ctx->override_driver_name) || (driver_name_env && (geteuid() == getuid()))) {
+        const char *driver = ctx->override_driver_name ?
+                             ctx->override_driver_name : driver_name_env;
+
+        drivers[0] = strdup(driver);
```
XinfengZhang, line 701: should free previous driver[0] if there are already one

va/va.c
```diff
@@ -719,15 +661,78 @@ VAStatus vaSetDriverName(
...
+        num_drivers = 1;
```
XinfengZhang, line 702: also , maybe has the second driver name, so driver[1] may leak

evelikov: Indeed we are leaking all drivers - quite busy ATM, will open a MR in a few days unless someone beats me to it.

Note that as the XXX comment just above says, the current ordering is off. So there is a serious work to be done here and fixing things.

evelikov: Should be fixed with https://github.com/intel/libva/pull/719 ... plus a typo fix.

PR #683, opened 2023-02-06
va: remove dummy isValid() hook

evelikov:
Bunch of the backends are using a simple
  return pDisplayContext && pDisplayContext->pDriverContext

... where pDisplayContext is always true. A few others are checking the
display_type, which in itself is absolutely useless.

So move the pDriverContext check within core libva and drop the isValid
hook across all winsys libraries.

### DIFF
--- a/va/wayland/va_wayland.c
+++ b/va/wayland/va_wayland.c
@@ -54,15 +54,6 @@ va_wayland_error(const char *format, ...)
     va_end(args);
 }
 
-static int
-va_DisplayContextIsValid(VADisplayContextP pDisplayContext)
-{
-    VADriverContextP const pDriverContext = pDisplayContext->pDriverContext;
-
-    return (pDriverContext &&
-            pDriverContext->display_type == VA_DISPLAY_WAYLAND);
-}
-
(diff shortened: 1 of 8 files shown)


va/wayland/va_wayland.c
```diff
@@ -54,15 +54,6 @@ va_wayland_error(const char *format, ...)
     va_end(args);
 }
 
-static int
-va_DisplayContextIsValid(VADisplayContextP pDisplayContext)
-{
-    VADriverContextP const pDriverContext = pDisplayContext->pDriverContext;
-
-    return (pDriverContext &&
-            pDriverContext->display_type == VA_DISPLAY_WAYLAND);
```
XinfengZhang, line 63: this will not be checked?

evelikov: The display_type is set alongside the isValid hook, in the respective GetDisplay function. As such it will always be the same (barring very specific memory corruption, which is a straw man argument), is borderline nonsensical and can be removed.

XinfengZhang (approved):

PR #685, opened 2023-02-07
fix: Add libva.def into distribution package

XinfengZhang:
libva.def is missing in release tarball

### DIFF
--- a/va/Makefile.am
+++ b/va/Makefile.am
@@ -149,4 +149,5 @@ EXTRA_DIST = \
 	libva.syms		\
 	va_version.h.in		\
 	meson.build		\
+        libva.def               \
 	$(NULL)


dvrogozh (approved):
Ok, so missing this in sources tarball will cause build error on Windows (if someone builds from tarball rather than git-clone copy).

evelikov:
Distribution package usually refers to what Ubuntu/Fedora/etc ship. `s/package/tarball/` seems more suitable

Mind you the whole `win32`  folder + `compat_win32.h` are missing, so as-is this PR is ... not so helpful

dvrogozh:
Actually we have a zoo of dist source tarballs which can be produced:
* w/ meson on Windows
* w/ meson on Linux
* w/ autotools on Linux

@XinfengZhang : you did use autotools tarball to attach to https://github.com/intel/libva/releases/tag/2.17.0, right?

autotools most definitely produces different tarball than meson on Linux or Windows. Whether meson tarballs will be the same across Windows/Linux - I am not sure, I am concerned about WITH_WIN32 kind of protections in meson scripts which protect some of the sources. @evelikov : can you suggest what we should expect here?

Now, we never added Windows build support to autotools and I suggest we should not to. That being said, I think we need to stop posting autotools dist tarballs and either stop posting tarballs at all or use meson produced tarballs.

I also started https://github.com/intel/libva/pull/686 to add make dist into ci, but the weak point is that verification is done on the same OS where tarball is generated. So, there is a chance that on other OS it won't build - which is the case for autotools one. I am not sure that I want to add all that complexity into ci.

eli-schwartz:
- autotools:
  - by default adds any files used in build rules
  - can remove files with build rules of `nodist_*`
  - can add files with EXTRA_DIST
  - can customize tarball with `dist-hook`
- Meson:
  - by default adds any files checked into git
  - can remove files with .gitattributes export-ignore
  - can customize tarball with `meson.add_dist_script()`

Meson should produce the same tarballs on any OS, unless your `meson.add_dist_script()` does different things on different OSes.

...

You may want to run autogen.sh as a meson dist script to ensure Meson-produced tarballs have the generated configure scripts etc. to successfully build with autotools.

dvrogozh:
> Meson should produce the same tarballs on any OS, unless your meson.add_dist_script() does different things on different OSes.

We don't use add_dist_script as of now:
```
$ fgrep -rsn add_dist_script
$
```

> You may want to run autogen.sh as a meson dist script to ensure Meson-produced tarballs have the generated configure scripts etc. to successfully build with autotools.

Ok, so as of now meson will produce a dist package which will be different from what we produced with autotools in a sense that some targets autotools was adding (like ./configure) will be missing? and the way to fix that will be to use add_dist_script...

eli-schwartz:
Yes.

eli-schwartz:
The only files which autotools dist on Linux creates, that Meson does not, are:
- va/va_version.h
  - supposed to be created by ./configure
  - should not be distributed, but is because it's listed as a source file
  - harmless
- va/wayland/wayland-drm-client-protocol.c
  - supposed to be created by wayland-scanner at build time
  - you're supposed to regenerate this using the most recent version of the scanner tool

Arguably it is "more correct" for Meson to exclude those. So those differences between autotools and Meson are fine.

dvrogozh:
I personally would prefer to go with whatever meson produces right now, i.e. without those additional files from autotools. But I am not the one who uses dist tarballs, so we need someone else's opinion on that.

XinfengZhang:
actually, this is a simple fix for such issue when un-tar this tarball and build with meson
"../va/meson.build:70:0: ERROR: File libva.def does not exist." reported by others, but looks it should be fixed in another way : such as remove this requirement from meson.build for non-win32 build.

PR #687, opened 2023-02-10
compat_win32: Add getenv wrapper for WIN32

sivileri:
The Windows getenv function reads the snapshot of values of env. variables at va.dll load time, disregarding future putenv calls from the same process.

This causes issues for apps/tests that do putenv/setenv for LIBVA_DRIVER_NAME or LIBVA_DRIVERS_PATH before calling vaInitialize. For example, the vaInitialize_vaTerminate tests in libva-utils test suite.

Fortunately there's a GetEnvironmentVariableA function that does the proper env. var reading taking into account calls to putenv/setenv. Use that function instead for Windows.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-170 https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablea

### DIFF
--- a/va/compat_win32.h
+++ b/va/compat_win32.h
@@ -29,6 +29,11 @@
 #include <io.h>
 #include <time.h>
 
+// Include stdlib.h here to make sure we
+// always replace MSVC getenv definition
+// with the macro / _getenv inline below
+#include <stdlib.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -45,6 +50,15 @@ typedef LONG NTSTATUS;
 #endif
 
 typedef unsigned int __uid_t;
+
+#if _MSC_VER
+#define getenv _getenv
+inline char* _getenv(const char *varname)
+{
+    static char _getenv_buf[32767];
+    return GetEnvironmentVariableA(varname, &_getenv_buf[0], sizeof(_getenv_buf)) ? &_getenv_buf[0] : NULL;
+}
+#endif
 
 #ifdef _MSC_VER
 inline char* strtok_r(char *s, const char *delim, char **save_ptr)


va/compat_win32.h
```diff
@@ -46,6 +46,15 @@ typedef LONG NTSTATUS;
 
 typedef unsigned int __uid_t;
 
+#if _MSC_VER
+#define getenv _getenv
+char _getenv_buf[32768];
```
dvrogozh, line 51: This will define `_getenv_buf` in each .c file which includes this header. And that might lead to multiple definitions of same variable, i.e. link error. Can you move that to function implementation and make static?

dvrogozh (changes requested):

sivileri:
Related https://github.com/intel/libva-utils/pull/307

va/compat_win32.h
```diff
@@ -46,6 +46,15 @@ typedef LONG NTSTATUS;
 
 typedef unsigned int __uid_t;
 
+#if _MSC_VER
+#define getenv _getenv
```
dvrogozh, line 50: So, Windows does define getenv in `<stdlib.h>`: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-170. And since code did built without error we were fetching this definition.

It seems that's important in which order we include header files. If compat_win32.h will by any chance be included before stdlib.h, then your trick to replace getenv won't work. Maybe we need include stdlib.h here in compat_win32.h explicitly to make sure we always replace getenv definition?

sivileri: You're right. It is picking the include stdlib.h from va.c before compat_win32. For the Windows build path, I moved the stdlib.h inclusion out from va.c and into compat_win32.c instead as you suggested to replace the definition.

va/va.c
```diff
@@ -34,14 +34,14 @@
...
 #include <stdio.h>
-#include <stdlib.h>
 #include <string.h>
 #if defined(_WIN32)
 #include "compat_win32.h"
 #define DRIVER_EXTENSION    "_drv_video.dll"
 #define DRIVER_PATH_STRING  "%s\\%s%s"
 #define ENV_VAR_SEPARATOR ";"
 #else
+#include <stdlib.h>
```
dvrogozh, line 44: I don't think we need this change here. You might use it for testing that on Windows all is getting build and works as expected, but code here was good as is: stdlib.h was defined before compat_win32.h. I think there is no need to change this file. Revert.

va/compat_win32.h
```diff
@@ -28,6 +28,7 @@
 #include <winsock.h>
 #include <io.h>
 #include <time.h>
+#include <stdlib.h>
```
dvrogozh, line 31: leave in-source comment why we include it here - people most likely will miss the reason when reading the code

dvrogozh (approved):
looks good

evelikov (approved):
Looks great, thanks

sivileri:
@dvrogozh kindly pinging on possibly merging this PR and the related https://github.com/intel/libva-utils/pull/307 ? Thanks!

va/compat_win32.h
```diff
@@ -46,6 +51,15 @@ typedef LONG NTSTATUS;
 
 typedef unsigned int __uid_t;
 
+#if _MSC_VER
+#define getenv _getenv
+inline char* _getenv(const char *varname)
+{
+    static char _getenv_buf[32768];
```
XinfengZhang, line 58: I am not familiar with windows GetEnviomentVariableA, but looks it is too large?

sivileri: According to [the API documentation](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getenvironmentvariable) for the `lpBuffer` argument to `GetEnviomentVariableA`, "An environment variable has a maximum size limit of 32,767 characters, including the null-terminating character".

As [getenv](https://cplusplus.com/reference/cstdlib/getenv/) returning char* is not freed by the caller and points to an internal block of memory whose content or validity may be altered by further calls to getenv, I wanted to follow this behavior with the wrapper function here (as opposed as returning dynamically allocated memory the caller does not free).

dvrogozh: Sounds good to me. @XinfengZhang ?

XinfengZhang (approved):

PR #688, opened 2023-02-10
va_drm_utils: Map d3d12 gallium driver for Windows Subsytem for Linux

sivileri:
In Windows Subsystem for Linux we use VGEM to expose a dri device and load the mesa d3d12 gallium driver.

The d3d12 component will then load the actual GPU vendor user mode driver lib from /lib/wsl/lib or /lib/wsl/drivers

When libva initializes with DRM, add detection of the VGEM device in the WSL environment, and only then map to WSL/d3d12 driver name so that libva/drm loads d3d12_drv_video.so mesa gallium driver.

### DIFF
--- a/va/drm/va_drm_utils.c
+++ b/va/drm/va_drm_utils.c
@@ -27,6 +27,7 @@
 #include "sysdeps.h"
 #include <xf86drm.h>
 #include <sys/stat.h>
+#include <sys/utsname.h>
 #include "va_drm_utils.h"
 #include "va_drmcommon.h"
 
@@ -42,6 +43,7 @@ static const struct driver_name_map g_driver_name_map[] = {
     { "radeon",     "r600"     }, // Mesa Gallium driver
     { "radeon",     "radeonsi" }, // Mesa Gallium driver
     { "amdgpu",     "radeonsi" }, // Mesa Gallium driver
+    { "WSL",        "d3d12"    }, // Mesa Gallium driver
     { "nvidia-drm", "nvidia"   }, // NVIDIA driver
     { NULL,         NULL }
 };
@@ -110,6 +112,15 @@ VA_DRM_GetDriverName(VADriverContextP ctx, char **driver_name_ptr, int candidate
         return VA_STATUS_ERROR_INVALID_DISPLAY;
 
     *driver_name_ptr = va_DRM_GetDrmDriverName(drm_state->fd);
+
+    /* Map vgem to WSL2 for Windows subsystem for linux */
+    struct utsname sysinfo = {};
+    if (!strncmp(*driver_name_ptr, "vgem", 4) && uname(&sysinfo) >= 0 &&
+        strstr(sysinfo.release, "WSL")) {
+        free(*driver_name_ptr);
+        *driver_name_ptr = strdup("WSL");
+    }
+
     if (!*driver_name_ptr)
         return VA_STATUS_ERROR_UNKNOWN;
 


va/drm/va_drm_utils.c
```diff
@@ -110,6 +112,14 @@ VA_DRM_GetDriverName(VADriverContextP ctx, char **driver_name_ptr, int candidate
         return VA_STATUS_ERROR_INVALID_DISPLAY;
 
     *driver_name_ptr = va_DRM_GetDrmDriverName(drm_state->fd);
+
+    /* Map vgem to WSL2 for Windows subsystem for linux */
+    struct utsname sysinfo = {};
+    if (!strncmp(*driver_name_ptr, "vgem", 4)
+        && !uname(&sysinfo)
+        && strstr(sysinfo.release, "WSL"))
+        *driver_name_ptr = strdup("WSL");
```
dvrogozh, line 121: You are losing memory here since driver_name_ptr containerd "vgem" which needs to be freed:
```
if (...) {
  free(*driver_name_ptr);
  *driver_name_ptr = strdup("WSL");
}
```

va/drm/va_drm_utils.c
```diff
@@ -110,6 +112,14 @@ VA_DRM_GetDriverName(VADriverContextP ctx, char **driver_name_ptr, int candidate
         return VA_STATUS_ERROR_INVALID_DISPLAY;
 
     *driver_name_ptr = va_DRM_GetDrmDriverName(drm_state->fd);
+
+    /* Map vgem to WSL2 for Windows subsystem for linux */
+    struct utsname sysinfo = {};
+    if (!strncmp(*driver_name_ptr, "vgem", 4)
+        && !uname(&sysinfo)
```
dvrogozh, line 119: from `man 3 uname`:
> Upon successful completion, a non-negative value shall be returned.

I.e. it might necessarily return 0 on success depending on implementation.

Also, does VS support this function?

dvrogozh (changes requested):

va/drm/va_drm_utils.c
sivileri: Great catch, fixed it

va/drm/va_drm_utils.c
sivileri: Gotcha, I was reading from [here](https://man7.org/linux/man-pages/man2/uname.2.html) instead, to make sure I added `&& (uname(&sysinfo) >= 0)` which should work with both definitions.

This will not build for windows, it's using libva in [Windows Subsystem for linux](https://learn.microsoft.com/en-us/windows/wsl/about#what-is-wsl-2). The goal of this change is to automatically detect and load the d3d12_drv_video.so mesa gallium driver instead of having to always set `LIBVA_DRIVER_NAME=d3d12`

sivileri: More context: We enabled libva in WSL/Windows native last year by adding an d3d12-based implementation of the mesa VA driver. When using that driver on Windows, it is named `vaon12_drv_video.dll` and uses the libva-win32 backend. In WSL it's named `d3d12_drv_video.so` and uses the usual backends like DRM.

sivileri:
Related https://github.com/intel/libva-utils/pull/308

va/drm/va_drm_utils.c
```diff
@@ -29,6 +29,7 @@
 #include <sys/stat.h>
 #include "va_drm_utils.h"
 #include "va_drmcommon.h"
+#include <sys/utsname.h>
```
dvrogozh, line 32: minor: can you move this next to sys/stat.h above?

dvrogozh (approved):
looks good

va/drm/va_drm_utils.c
```diff
@@ -110,6 +112,16 @@ VA_DRM_GetDriverName(VADriverContextP ctx, char **driver_name_ptr, int candidate
         return VA_STATUS_ERROR_INVALID_DISPLAY;
 
     *driver_name_ptr = va_DRM_GetDrmDriverName(drm_state->fd);
+
+    /* Map vgem to WSL2 for Windows subsystem for linux */
+    struct utsname sysinfo = {};
+    if (!strncmp(*driver_name_ptr, "vgem", 4)
+        && (uname(&sysinfo) >= 0)
+        && strstr(sysinfo.release, "WSL")) {
```
evelikov, line 120: nit:
```
    if (!strncmp(*driver_name_ptr, "vgem", 4) && uname(&sysinfo) >= 0 &&
        strstr(sysinfo.release, "WSL")) {
...
```

But more importantly, can we have this combination (Windows/WSL) in the CI matrix.

dvrogozh: > can we have this combination (Windows/WSL) in the CI matrix

Can you elaborate what do you mean here? which CI matrix? on which side?

sivileri: WSL runs a virtualized Linux distro on top of Windows like [Ubuntu WSL](https://ubuntu.com/wsl) using the original linux libva binaries, no special build is necessary.

evelikov: Was under the impression that everything WSL is built within a different environment. Glad to hear that's not the case :+1:

sivileri:
@dvrogozh Kindly pinging on this, anything else I can help with to help merging this PR and the related https://github.com/intel/libva-utils/pull/308 ? Thanks!

XinfengZhang (approved):

PR #689, opened 2023-02-15
x11: use LIBVA_DRI3_DISABLE in GetNumCandidates

evelikov:
As reported by @zhmars, we need to check the override in both code paths.

That's because we get only one candidate, while for Intel GPUs we have 2 drivers - so this might work when the user needs iHD, but will fail if they rely on the older driver.


XinfengZhang (approved):

dvrogozh (approved):
looks good

evelikov:
Thanks team. Github was complaining the branch eing out of date, so I've rebased it.

PR #698, opened 2023-03-16
Enable support for license acquisition of multiple protected playbacks

ankitjain88:


va/va_prot.h
```diff
@@ -441,7 +441,18 @@ typedef struct _VAProtectedSessionExecuteBuffer {
     /** \brief Return the result of this function. The status result is IHV's
      * implementation specific */
     uint32_t status;
-    uint32_t va_reserved[VA_PADDING_LOW];
+    /** \brief vtag is to support provisioning and license acquisition of multiple playbacks
+     *
+     * If TEE interface was created using VAEntrypointProtectedTEEComm, vtag should
+     * be valid i.e. non zero in case of multiple playback and vtag should be
+     * invalid i.e. zero in case of single playback
```
XinfengZhang, line 448: zero should be invalid to keep the backward compatibility? if it is old application , which will treat it as reserved, it is zero

ankitjain88: Yes, zero is invalid vtag value and it will keep backward compatibility. If it is an old application, they can keep treating vtag as reserved bytes and send zero.

XinfengZhang (approved):

va/va_prot.h
kumar-gaurav-2008: brief - vtag represents TEE sandbox identification for multiple playback scenario

if TEE interface is created using VAEntrypointProtectedTEEComm then
   vtag - non-zero for multiple playback
              zero for single playback
if TEE interface is not created using VAEntrypointProtectedTEEComm then 
   vtag - reserved

kumar-gaurav-2008: Please use above text

PR #702, opened 2023-03-22
use secure_getenv instead of getenv

XinfengZhang:
replace the getenv with secure_getenv to avoid some potential risk feature test micro for this function is _GNU_SOURCE

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,9 @@ PKG_PROG_PKG_CONFIG
 AC_HEADER_STDC
 AC_SYS_LARGEFILE
 
+#check for secure_getenv
+AC_CHECK_FUNCS(secure_getenv)
+
 # Check for Doxygen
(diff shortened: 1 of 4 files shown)


dvrogozh (approved):
looks good

evelikov:
The secure variant honours certain capabilities like "CAP_NICE" IIRC.

Compositors tend to use that, so this commit will break the overrides for things the vast majority of desktop users - kde/plasma, gnome/shell, gamescope (as used by the Steam Deck) ...

evelikov:
For CAP_NICE I suggest reading [through the manual](https://man7.org/linux/man-pages/man7/capabilities.7.html) it describes things quite well.

I'm not a security engineer, but last time speaking with one they did not use "maybe" and "potentially". Instead they would outlining the attack vector and build from there.

PR #703, opened 2023-03-22
minor fix of readme for meson build

XinfengZhang:
the build folder already exist for gen_version.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ sudo make install
 ```
 or build using Meson
 ```
-mkdir build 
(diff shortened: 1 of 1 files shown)


evelikov:
Using the "build" folder will not cause any issues - meson does not care. AFACIT this does not fix anything.

XinfengZhang:
there are already build folder, and include some android file, dont want bring in confusion. 
and I always "delete everything under build , then run meson again" ... maybe not a good habit. ..

Issue #711, opened 2023-05-10
win32/msys: libva.pc causes app dependency on non-existing va.dll


dvrogozh:
Using msys2 to build libva and libva-utils:
```
pacman -S msys/bison msys/flex msys/git mingw64/mingw-w64-x86_64-meson mingw64/mingw-w64-x86_64-python-mako mingw-w64-x86_64-toolchain
git clone https://github.com/intel/libva.git && cd libva
meson setup --prefix=/c/install _build
meson compile -C _build
meson install -C _build
cd ..
export PKG_CONFIG_PATH=/c/install/lib/pkgconfig
git clone https://github.com/intel/libva-utils.git && cd libva-utils
meson setup --prefix=/c/install _build
meson compile -C _build
meson install -C _build
```
After that we are getting this:
```
$ ls
libva.dll  libva_win32.dll  vainfo.exe
$ ldd vainfo.exe
        ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x7ffe1eaa0000)
        KERNEL32.DLL => /c/windows/System32/KERNEL32.DLL (0x7ffe1ceb0000)
        KERNELBASE.dll => /c/windows/System32/KERNELBASE.dll (0x7ffe1c3e0000)
        msvcrt.dll => /c/windows/System32/msvcrt.dll (0x7ffe1e9b0000)
        va.dll => not found
        va_win32.dll => not found
```

I.e. vainfo depends on non-existing va.dll and va_win32.dll. @sivileri : fyi

dvrogozh:
@eli-schwartz can you advise on this issue? is there a way in meson to query library link name to put in .pc file?

eli-schwartz:
So clearly it should depend on libva.dll rather than va.dll, but due to Microsoft things, either one could be intended by `-lva`. This is somewhat more complex because actually it looks for an import library, not a dll, at link time.

Obviously this worked somehow when compiling libva-utils -- meson, and the compiler, found the intended libraries. This would be reported by the verbose compilation logs that were omitted from the bug report

So the question is, why does libva-utils link to an import library successfully, but then erroneously think that the associated DLL file name to embed into the vainfo.exe binary metadata should be different from what it actually is?

eli-schwartz:
```
  vs_module_defs : 'libva.def',
```

This file hardcodes the library name as "va", which is fine as long as you also conditionally set the name prefix in meson to empty on platforms where that def file is used.

dvrogozh:
[...]

PR #712, opened 2023-05-12
va/win32: rely on compiler to define link names

dvrogozh:
Fixes: #711

This removes hardcoding library names to link applications against. Hardcode was set to be va.dll while mingw produces library with lib prefix, i.e. libva.dll. This causes missing libraries in link dependencies with obvious workaround to `cp libva.dll va.dll`.

### DIFF
--- a/va/libva.def
+++ b/va/libva.def
@@ -1,4 +1,3 @@
-LIBRARY va
 EXPORTS
     vaInitialize
     vaErrorStr
--- a/va/libva_win32.def
+++ b/va/libva_win32.def
@@ -1,3 +1,2 @@
-LIBRARY va_win32
 EXPORTS
     vaGetDisplayWin32


dvrogozh:
@eli-schwartz : do you mind to review this PR?

eli-schwartz (approved):
This seems correct although I haven't tested it since I don't have a mingw system.

I do notice that even though it is built in CI for that platform, libva itself doesn't contain a test suite? I think it would be a good idea to create a simple test program that links against libva and tries to use a basic form of the API. It could then be run as part of `meson test -C builddir/`

This would be useful for doing rapid edit/build/test cycles when hacking on the code, and it would also constitute a linkability test which would fail in CI when there are bugs such as the one this PR fixes.

dvrogozh:
> I do notice that even though it is built in CI for that platform, libva itself doesn't contain a test suite? 

That is true. The history behind is that original libva repository was at some point split into 2: this repo containing the library + https://github.com/intel/libva-utils containing utils and tests. So, unfortunately tests migrated out of this repository which from my perspective was not a correct thing to do for the libva - that's not that big project to have a need for such separation.

dvrogozh:
I wonder whether there are tests in libva-utils which would run w/o GPU? @XinfengZhang - do we have some test subset in libva-utils which allows some basic checks for libva w/o any runtime driver?

sivileri:
@dvrogozh From the docs, I think the `LIBRARY` statement is mandatory in the .def file definition to indicate a library, but I also see some places where the definitions only have `EXPORTS` (ie. [here](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/gallium/targets/va/vaon12.def.in) then processed with [this](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/bin/gen_vs_module_defs.py) script) so this may be valid as well, and looks like a shorter solution than adding different `name_prefix` params to `shared_library` for different platforms (va(_win32).dll in Windows, libva(_win32).dll otherwise). 

The MSVC CI seems to be producing va.dll/va_win32.dll correctly in [the associated CI run](https://github.com/intel/libva/actions/runs/4962306314/jobs/8880190466?pr=712)

@jenatali do you know if without the `LIBRARY` entry, the compiler will take the library names 'va' / 'va_win32' from the meson `shared_library` statements in va/meson.build and this .def file would be it's valid ?

sivileri:
> I wonder whether there are tests in libva-utils which would run w/o GPU? @XinfengZhang - do we have some test subset in libva-utils which allows some basic checks for libva w/o any runtime driver?

For windows, I wonder if you can run `vainfo --display win32 --device 0` on PCs with no GPU and vaon12 driver. It won't support/list any decode/encode entrypoints but it should be able to at least query/load the mesa driver, create/initialize the VADisplay from it, and query/print an empty entrypoint list from it.

dvrogozh:
> From the docs, I think the LIBRARY statement is mandatory

Which documents you are referencing to? I just found this one: https://learn.microsoft.com/en-us/cpp/build/reference/library?view=msvc-170 describing LIBRARY statement in def file with the remark "You can also use the [/OUT](https://learn.microsoft.com/en-us/cpp/build/reference/out-output-file-name?view=msvc-170) linker option to specify the DLL's output name." which means that LIBRARY is not mandatory....

dvrogozh:
> I wonder if you can run vainfo --display win32 --device 0 on PCs with no GPU and vaon12 driver.

That is interesting behavior, I was not aware of it. Which device is meant by 'device 0'? does it by any chance signify MSFT basic adapter? And, by any chance is this device available under docker to allow us to use this cmdline directly in our Windows ci?

sivileri:
> > I wonder if you can run vainfo --display win32 --device 0 on PCs with no GPU and vaon12 driver.
> 
> That is interesting behavior, I was not aware of it. Which device is meant by 'device 0'? does it by any chance signify MSFT basic adapter? And, by any chance is this device available under docker to allow us to use this cmdline directly in our Windows ci?

`--device <i>` will select the adapter over all the GPUs available in your system, including `Microsoft Basic Adapter`. So if you have a hybrid GPU platform you will have --device 0, 1 the integrated,discrete GPUs and 2 being `Microsoft Basic Adapter`. I think you should be able to use it in the CI.

evelikov (approved):
In case it matters: all other projects I've seen either omit the LIBRARY like or user an @FOO@ placeholder and substitute it at build time.

So the MR looks good IMHO

XinfengZhang:
> I wonder whether there are tests in libva-utils which would run w/o GPU? @XinfengZhang - do we have some test subset in libva-utils which allows some basic checks for libva w/o any runtime driver?

unfortunately no, libva-utils include test, but it need libva + a driver.  because libva is  a thin layer,  the function/feature comes from backend driver

XinfengZhang (approved):

sivileri:
Looks good to me.

PR #713, opened 2023-05-25
va/win32: change default driver search path to bindir

dvrogozh:
On Windows dlls and drivers (vaon12 specifically) gets installed to bindir instead of libdir (or libdir/dri for va drivers). As a result libva fails to find driver at default location and user needs to setup LIBVA_DRIVERS_PATH explicitly pointing to driver location.

This patch changes default driver location for Windows to bindir.

@sivileri, @evelikov, @jenatali - please, provide comments.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -67,7 +67,16 @@ sysconfdir = join_paths(get_option('prefix'), get_option('sysconfdir'))
 
 driverdir = get_option('driverdir')
 if driverdir == ''
-  driverdir = join_paths(get_option('prefix'), get_option('libdir'), 'dri')
+  # "libdir" on Windows is essentially only for static and import libraries,
+  # while "bindir" is the actual runtime directory - containing both
+  # executable and dynamic libraries. During install meson uses correct install
+  # location depending on the type of library, requiring zero user intervention
+  # in the common case.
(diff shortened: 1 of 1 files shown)


jenatali (approved):
Yeah that makes sense to me.

sivileri (approved):
Looks good to me, thanks for submitting the change.

evelikov (approved):
Tl;Dr: The change is spot on.

Personally I would add an inline comment describing things (feel free to reuse any of my follow-up notes), since it's not obvious for non Windows devs.

Disclaimer: My Windows knowledge is 5+ years old, so please correct me at any point ;-)

IIRC "libdir" on Windows is essentially only for static and import libraries, while bindir is the actual runtime directory - containing both executable and dynamic libraries.

During install meson uses the corrects the install location depending on the type of library, requiring zero user intervention in the common case.

dvrogozh:
> Personally I would add an inline comment describing things<...>

I added inline comment. Let me know if something needs to be changed in a text. No functional changes vs. prev. version.

evelikov (approved):

XinfengZhang (approved):

PR #715, opened 2023-06-09
win32/driver loader: Allow backends to load drivers from registry path

sivileri:
Rework of https://github.com/intel/libva/pull/684 after rebasing for #682.

The different commits sequentially do:

- va: Extract the actual driver loading from va_OpenDriver into va_OpenDriverFromPath
     - No functional changes
- va: When driver loading fails with the provided name by the backend, try full path loading as fallback
    - This fixes loading of the registry driver full path for a given device
    - This fixes loading `LIBVA_DRIVER_NAME` pointing to an absolute path which we want to support.
    - No changes to existing loading mechanisms
- va/win32: Change default driver name to the full filename to the default vaon12_drv_video.dll filename shipped in [the VideoAccelerationCompatibilityPack](https://www.nuget.org/packages/Microsoft.Direct3D.VideoAccelerationCompatibilityPack/) package (also default binary name when building from mesa directly)
    - This allows the fallback to use `vaon12_drv_video` as a path and load it from the current directory (dlopen/LoadLibrary doesn't require the .dll extension on Windows)
- va: When driver loading fails with provided name, try without adding the `drv_video` suffix
    - This fixes loading `LIBVA_DRIVER_NAME` pointing to an absolute driver file name which we want to support.

CC @evelikov, @dvrogozh, @XinfengZhang

### DIFF
--- a/va/va.c
+++ b/va/va.c
@@ -370,9 +370,146 @@ static char *va_getDriverPath(const char *driver_dir, const char *driver_name)
     return driver_path;
 }
 
-static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
+static VAStatus va_openDriverFromPath(VADisplay dpy, char *driver_path)
 {
     VADriverContextP ctx = CTX(dpy);
+    VAStatus vaStatus = VA_STATUS_ERROR_UNKNOWN;
+    va_infoMessage(dpy, "Trying to open %s\n", driver_path);
+#if defined(RTLD_NODELETE) && !defined(ANDROID)
+    void* handle = dlopen(driver_path, RTLD_NOW | RTLD_GLOBAL | RTLD_NODELETE);
+#else
+    void* handle = dlopen(driver_path, RTLD_NOW | RTLD_GLOBAL);
+#endif
+    if (!handle) {
+        /* Don't give errors for non-existing files */
+        if (0 == access(driver_path, F_OK))
+            va_errorMessage(dpy, "dlopen of %s failed: %s\n", driver_path, dlerror());
+    } else {
+        VADriverInit init_func = NULL;
+        char init_func_s[256];
+        int i;
+
+        struct {
+            int major;
+            int minor;
+        } compatible_versions[VA_MINOR_VERSION + 2];
+        for (i = 0; i <= VA_MINOR_VERSION; i ++) {
+            compatible_versions[i].major = VA_MAJOR_VERSION;
+            compatible_versions[i].minor = VA_MINOR_VERSION - i;
+        }
+        compatible_versions[i].major = -1;
+        compatible_versions[i].minor = -1;
+
+        for (i = 0; compatible_versions[i].major >= 0; i++) {
+            if (va_getDriverInitName(init_func_s, sizeof(init_func_s),
+                                     compatible_versions[i].major,
+                                     compatible_versions[i].minor)) {
+                init_func = (VADriverInit)dlsym(handle, init_func_s);
+                if (init_func) {
+                    va_infoMessage(dpy, "Found init function %s\n", init_func_s);
+                    break;
+                }
+            }
+        }
+
+        if (compatible_versions[i].major < 0) {
+            va_errorMessage(dpy, "%s has no function %s\n",
+                            driver_path, init_func_s);
+            dlclose(handle);
+        } else {
+            struct VADriverVTable *vtable = ctx->vtable;
+            struct VADriverVTableVPP *vtable_vpp = ctx->vtable_vpp;
+            struct VADriverVTableProt *vtable_prot = ctx->vtable_prot;
+
+            vaStatus = VA_STATUS_SUCCESS;
+            if (!vtable) {
+                vtable = calloc(1, sizeof(*vtable));
+                if (!vtable)
+                    vaStatus = VA_STATUS_ERROR_ALLOCATION_FAILED;
+            }
+            ctx->vtable = vtable;
+
+            if (!vtable_vpp) {
+                vtable_vpp = calloc(1, sizeof(*vtable_vpp));
+                if (vtable_vpp)
+                    vtable_vpp->version = VA_DRIVER_VTABLE_VPP_VERSION;
+                else
+                    vaStatus = VA_STATUS_ERROR_ALLOCATION_FAILED;
+            }
+            ctx->vtable_vpp = vtable_vpp;
+
+            if (!vtable_prot) {
+                vtable_prot = calloc(1, sizeof(*vtable_prot));
+                if (vtable_prot)
+                    vtable_prot->version = VA_DRIVER_VTABLE_PROT_VERSION;
+                else
+                    vaStatus = VA_STATUS_ERROR_ALLOCATION_FAILED;
+            }
+            ctx->vtable_prot = vtable_prot;
+
+            if (init_func && VA_STATUS_SUCCESS == vaStatus)
+                vaStatus = (*init_func)(ctx);
+
+            if (VA_STATUS_SUCCESS == vaStatus) {
+                CHECK_MAXIMUM(vaStatus, ctx, profiles);
+                CHECK_MAXIMUM(vaStatus, ctx, entrypoints);
+                CHECK_MAXIMUM(vaStatus, ctx, attributes);
+                CHECK_MAXIMUM(vaStatus, ctx, image_formats);
+                CHECK_MAXIMUM(vaStatus, ctx, subpic_formats);
+                CHECK_STRING(vaStatus, ctx, vendor);
+                CHECK_VTABLE(vaStatus, ctx, Terminate);
+                CHECK_VTABLE(vaStatus, ctx, QueryConfigProfiles);
+                CHECK_VTABLE(vaStatus, ctx, QueryConfigEntrypoints);
+                CHECK_VTABLE(vaStatus, ctx, QueryConfigAttributes);
+                CHECK_VTABLE(vaStatus, ctx, CreateConfig);
+                CHECK_VTABLE(vaStatus, ctx, DestroyConfig);
+                CHECK_VTABLE(vaStatus, ctx, GetConfigAttributes);
+                CHECK_VTABLE(vaStatus, ctx, CreateSurfaces);
+                CHECK_VTABLE(vaStatus, ctx, DestroySurfaces);
+                CHECK_VTABLE(vaStatus, ctx, CreateContext);
+                CHECK_VTABLE(vaStatus, ctx, DestroyContext);
+                CHECK_VTABLE(vaStatus, ctx, CreateBuffer);
+                CHECK_VTABLE(vaStatus, ctx, BufferSetNumElements);
+                CHECK_VTABLE(vaStatus, ctx, MapBuffer);
+                CHECK_VTABLE(vaStatus, ctx, UnmapBuffer);
+                CHECK_VTABLE(vaStatus, ctx, DestroyBuffer);
+                CHECK_VTABLE(vaStatus, ctx, BeginPicture);
+                CHECK_VTABLE(vaStatus, ctx, RenderPicture);
+                CHECK_VTABLE(vaStatus, ctx, EndPicture);
+                CHECK_VTABLE(vaStatus, ctx, SyncSurface);
+                CHECK_VTABLE(vaStatus, ctx, QuerySurfaceStatus);
+                CHECK_VTABLE(vaStatus, ctx, QueryImageFormats);
+                CHECK_VTABLE(vaStatus, ctx, CreateImage);
+                CHECK_VTABLE(vaStatus, ctx, DeriveImage);
+                CHECK_VTABLE(vaStatus, ctx, DestroyImage);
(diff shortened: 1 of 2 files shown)


va/va.c
```diff
@@ -721,6 +721,10 @@ static VAStatus va_new_opendriver(VADisplay dpy)
         }
 
         vaStatus = va_openDriver(dpy, drivers[i]);
+        /* Try as path as fallback */
+        if (vaStatus != VA_STATUS_SUCCESS)
```
dvrogozh, line 724: Copying from PR description:
> va: When driver loading fails with the provided name by the backend, try full path loading as fallback
> This fixes loading of the registry driver full path for a given device
> This fixes loading LIBVA_DRIVER_NAME pointing to an absolute path which we want to support.
> No changes to existing loading mechanisms

This needs to be added to the commit message of 2nd commit (maybe paraphrased). Main thing to highlight is that LIBVA_DRIVER_NAME now allows full path specification.

va/win32/va_win32.c
```diff
@@ -35,7 +35,7 @@
  * which will be selected when provided with an adapter LUID which
  * does not have a registered VA driver
 */
-const char VAAPI_DEFAULT_DRIVER_NAME[] = "vaon12";
+const char VAAPI_DEFAULT_DRIVER_NAME[] = "vaon12_drv_video";
```
dvrogozh, line 38: emmm. But "vaon12_drv_video" is not a full path. Full path (on Windows) would be for example "C:\msys64\mingw64\bin\vaon12_drv_video.dll". You probably can name it "full driver name", but saying "full path" I personally understand different thing.

va/va.c
```diff
@@ -529,19 +527,20 @@ static VAStatus va_openDriver(VADisplay dpy, char *driver_name)
...
-            free(search_path);
-            return VA_STATUS_ERROR_ALLOCATION_FAILED;
-        }
-
-        vaStatus = va_openDriverFromPath(dpy, driver_path);
-        free(driver_path);
-        if (VA_STATUS_SUCCESS == vaStatus)
-            break;
+        for (int use_suffix = 1; use_suffix >= 0; use_suffix--) {
+            char *driver_path = va_getDriverPath(driver_dir, driver_name, use_suffix);
```
dvrogozh, line 531: will this construct `vaon12_drv_video_drv_video` with the changed default driver name for windows?

dvrogozh (changes requested):
Major problem which I have with this PR is: what are the VAAPI drivers which are/soon will be available from the device storage? As of now this PR is basically a placeholder for such drivers. If we will merge in this change, but such drivers won't appear we will need to support a code path which no one is actually using. Which does not make sense.

I suggest that we need to get an evidence that some VAAPI drivers in the device storage already exist or soon will appear to merge in this change.

va/va.c
```diff
@@ -370,9 +370,146 @@ static char *va_getDriverPath(const char *driver_dir, const char *driver_name)
...
     VADriverContextP ctx = CTX(dpy);
+    VAStatus vaStatus = VA_STATUS_ERROR_UNKNOWN;
+    va_infoMessage(dpy, "Trying to open %s\n", driver_path);
+#if defined(RTLD_NODELETE) && !defined(ANDROID)
+    void* handle = dlopen(driver_path, RTLD_NOW | RTLD_GLOBAL | RTLD_NODELETE);
+#else
+    void* handle = dlopen(driver_path, RTLD_NOW | RTLD_GLOBAL);
+#endif
+    if (!handle) {
+        /* Don't give errors for non-existing files */
```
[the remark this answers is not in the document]
sivileri: I'd happy to make that change to make it cleaner/simpler, but that is the already existing behavior in [va.c in master](https://github.com/intel/libva/blob/master/va/va.c) too, I didn't want to make any functional changes here other than extracting the `va_OpenDriverFromPath` function. Maybe this behaviour has a historical reason for being like this? The last change to this line was 16 years ago :/

va/va.c
sivileri: Yes, the first attempt will be `vaon12_drv_video_drv_video` and then fallback to `vaon12_drv_video` (no added suffix). This is done like this (and in this order) to prevent any functional changes that may break other backends that follow the strict name suffix on Linux. While this adds an extra attempt to the loading on Windows, I think it's less invasive than trying without suffix first (that would add an extra attempt to all other backends before finding their driver files on Linux)

va/win32/va_win32.c
sivileri: The full path should be `.\vaon12_drv_video.dll` intended as default for the [Nuget package](https://www.nuget.org/packages/Microsoft.Direct3D.VideoAccelerationCompatibilityPack/) that ships `vaon12_drv_video.dll` in the same directory as `va.dll`, `va_win32.dll`. 

As in Win `LoadLibrary` (aliased to `dlopen`) supports the file string to [have multiple semantics](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya#remarks) I only left the filename, which will first still try to open `.\vaon12_drv_video.dll` and then attempt again using the default Windows library search order.

- If no file name extension is specified in the lpFileName parameter, the default library extension .dll is appended. 
- When no path is specified, the function searches for loaded modules whose base name matches the base name of the module to be loaded. If the name matches, the load succeeds. Otherwise, the function searches for the file. The first directory searched is the directory containing the image file used to create the calling process...

sivileri:
> Major problem which I have with this PR is: what are the VAAPI drivers which are/soon will be available from the device storage? As of now this PR is basically a placeholder for such drivers. If we will merge in this change, but such drivers won't appear we will need to support a code path which no one is actually using. Which does not make sense.
> 
> I suggest that we need to get an evidence that some VAAPI drivers in the device storage already exist or soon will appear to merge in this change.

The main intention of this change is to fix the loadable registry drivers, but this change also allows more flexibility from where to load vaon12_drv_video.dll from (default vaon12_drv_video loading from va.dll directory without setting `LIBVA_DRIVER_NAME` nor `LIBVA_DRIVERS_PATH`, more paths searched in LoadLibrary, specify LIBVA_DRIVER_NAME as a path or name without suffix, etc). 

The most immediate effect would be allowing using the [Nuget](https://www.nuget.org/packages/Microsoft.Direct3D.VideoAccelerationCompatibilityPack/) without setting `LIBVA_DRIVER_NAME` nor `LIBVA_DRIVERS_PATH` or when you build vaon12 along with libva DLLs in any path (which now works with your fix to default to the bindir on Windows which is where both DLLs are).

Even without the existence of other loadable drivers now I still think the scenario should not remain broken until one of them ships as it may take some time to deploy a new libva version to fix that when time comes.

sivileri:
> @sivileri this seems to ignore my [earlier suggestions](https://github.com/intel/libva/pull/684#issuecomment-1426366896) - is that intentional?
> 
> Namely we seems to be having having divergent codeflow with a fallback/retry + random path/name construction at runtime. Instead we could have a static inline function, effectively constexp, based on the platform - WIN32/*NiX, etc.

I tried the suggested approach at first, but when I got to reimplementing the whole suffix, multiple name overrides, multiple path search overrides in va_win32.c seemed to be duplicating a significant amount of complex code and logic from va/va.c and possibly adding new bugs in that re-implementation.
I also had the `IsDriverNameAbsolute` at first, and it would have worked if I reimplemented all the path generation logic in va/win32/va_win32.c but to avoid that reimplementation, I thought just adding a fallback (which is also not breaking back-compat or other backends) and reuse the core naming/suffix/path generation logic with all the overrides and just falling back to a path otherwise.

evelikov:
If I'm understanding the initial proposal correctly, the goal is to have the full path + filename fetched from the registry. As such, the separate VA path/driver env. variable overrides are non-applicable, right? Hence there is no need to (re)implement either of those.

Alternatively, I you're saying that the env. overrides (as you saw with the test suite, getenv behaves strangely on Windows) must be honoured - how is that expected to work. More importantly does it need to work?

sivileri:
> If I'm understanding the initial proposal correctly, the goal is to have the full path + filename fetched from the registry. As such, the separate VA path/driver env. variable overrides are non-applicable, right? Hence there is no need to (re)implement either of those.
> 
> Alternatively, I you're saying that the env. overrides (as you saw with the test suite, getenv behaves strangely on Windows) must be honoured - how is that expected to work. 

1. `Win32->GetDriverNames()` returns the driver for a given adapter LUID that comes from the Windows registry and a default driver. Those strings must be passed as-is to `dlopen/LoadLibrary` to work in Windows as intended. 

2. We also want to keep the flexibility of the libva overrides with the same precedence over the `GetDriverNames` strings, but _also_ allowing the name override to work as a path/without the suffix.

> More importantly does it need to work?

Yes, actually this is the way it's working today with [Windows VaOn12 Nuget package](https://www.nuget.org/packages/Microsoft.Direct3D.VideoAccelerationCompatibilityPack/) by setting `LIBVA_DRIVER_NAME=vaon12` and `LIBVA_DRIVERS_PATH` to where `vaon12_drv_video.dll` is.

evelikov:
Right so both options must work - that sounds unfortunate. Is that the long term plan to have both, or we can switch to registry only once  Nuget has migrated?

How are things supposed to work in the following situations:
 - both the registry and `LIBVA_DRIVER_NAME` are set
 - both the registry and `LIBVA_DRIVERS_PATH` are set
 - the registry is unset, while only `LIBVA_DRIVER_NAME` or `LIBVA_DRIVERS_PATH` is set
 
Should there be warnings/errors emitted by libva at any point in the above combinations?

sivileri:
> Right so both options must work - that sounds unfortunate. Is the long term plan to have both, or we can switch to registry only once Nuget has migrated?

We would like keep both, to allow overrides in some dev/test environments and in other situations like apps wanting to override it for their own process.

> How are things supposed to work in the following situations:
> 
> * both the registry and `LIBVA_DRIVER_NAME` are set

With the same precedence as today, `LIBVA_DRIVER_NAME` takes precedence over any driver returned by `vaGetDriverNames` (including the registry driver).

> * both the registry and `LIBVA_DRIVERS_PATH` are set

`LIBVA_DRIVERS_PATH` takes precedence as today. What actually happens here with these changes is that libva attempts to concat `LIBVA_DRIVERS_PATH` to the registry (absolute) path (which doesn't work) and then it tries with the second option (default driver) given by `GetDriverNames` = `<LIBVA_DRIVERS_PATH>\vaon12_drv_video.dll`

> * the registry is unset, while only `LIBVA_DRIVER_NAME` or `LIBVA_DRIVERS_PATH` is set

`LIBVA_DRIVER_NAME` or `LIBVA_DRIVERS_PATH` take precedence as usual over `vaGetDriverNames`.

> Should there be warnings/errors emitted by libva at any point in the above combinations?

Every attempt is currently being logged by `va_infoMessage("Trying to open...")` in `va_openDriverFromPath` as before, now just has more combinations to fallback into if the previously existing ones fail.

evelikov:
Thanks, now it makes more sense. Will need to think a bit - currently it sounds like you want to [have your the cake and eat it](https://en.wikipedia.org/wiki/You_can%27t_have_your_cake_and_eat_it) which makes is rather hairy.

evelikov:
Disclaimer: coffee might not have kicked in yet... hope the following is at least partially coherent :sweat_smile: 

Building on the earlier suggestion:
 - the helper looks good, couple of nitpicks:
   - NEW: optionally returns early (as suggested by @dvrogozh)
   - NEW: optionally remove the [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions)  
 - introduce a static inline bool IsDriverNameAbsolute(void) - explain in the commit message how it'll be used
   - return false for now
 - create a "fullpath" variant of va_openDriver()
    - no fancy LIBVA_DRIVERS_PATH handling, etc
    - EDIT: use the above helper to select "at runtime" (at compile time in practise due to DCE) the correct code path, _when both variables are not set_ (example below)
 -  adjust the win32 backend to produce absolute path
    - use LIBVA_DRIVERS_PATH + VA_DRIVERS_PATH + DRIVER_EXTENSION, or use alternative as applicable to construct the default
   - no need to rename the driver_name -> driver_path - just add an inline comment with the explanation from your colleague
   - add an ifdef _WIN32 case in IsDriverNameAbsolute() that returns true

Hand-wavy example:
```
const bool has_overrides = getenv() && getenv()

for X in drivers;
    ...
    
    /* IsDriverNameAbsolute() is WIN32 specific quirk, where by default we
     * want the driver to provide the full path, as stored in the registry.
     * 
     * Although we still want to have the option to override via env. variables
     * $FOR_REASONS.
     *
     * The function is constexp and will be DCE discarded for non-Windows
     * platforms.
     */
    if (IsDriverNameAbsolute() && !has_override)
        va_no_construct_open()
    else
        normal_va_open()
        
    ...
```

Aside: it makes sense to push the search path (getuid/geteuid/getenv + fallback) handling a level up into `va_new_opendriver()` and pass the `search_path` as an argument to `va_openDriver()`. Ideally that will be a separate commit in this PR, but I don't have strong opinion.

This should handle all the cases you've outlined appropriately, Without all the weird stuff this PR does for the non-windows builds.

PR #719, opened 2023-06-22
Mem leaks

evelikov:
- va: don't leak driver names, when override is set
- va: add missing space in the env.var override info message

No rocket science here ;-)


dvrogozh (approved):
looks good

XinfengZhang (approved):

PR #720, opened 2023-06-25
[Don't Merge] Add repeat frame flag and HEVC PAK statistics for multi-passes encoding

ChenXiaomin0306:
1. repeat frame flag to denote whether this is one new frame or repeated frame of previous frame for encoding context update.
2. HEVC CU level stream out from PAK engine can be output to application.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -2316,7 +2316,11 @@ typedef enum {
     /** \brief Buffer type used for FEI input frame level parameters */
     VAEncMiscParameterTypeFEIFrameControl = 18,
     /** \brief encode extension buffer, ect. MPEG2 Sequence extenstion data */
-    VAEncMiscParameterTypeExtensionData = 19
+    VAEncMiscParameterTypeExtensionData = 19,
+    /** \brief Buffer type used for sending repeat frame flag to the encoder */
+    VAEncMiscParameterTypeRepeatFrame = 20,
+    /** \brief Buffer type used for HEVC Pak statistics */
+    VAEncMiscParameterTypeHevcPakStat = 21
 } VAEncMiscParameterType;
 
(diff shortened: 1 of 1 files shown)


va/va.h
```diff
@@ -2316,7 +2316,11 @@ typedef enum {
     /** \brief Buffer type used for FEI input frame level parameters */
     VAEncMiscParameterTypeFEIFrameControl = 18,
     /** \brief encode extension buffer, ect. MPEG2 Sequence extenstion data */
-    VAEncMiscParameterTypeExtensionData = 19
+    VAEncMiscParameterTypeExtensionData = 19,
+    /** \brief Buffer type used for sending repeat frame flag to the encoder */
+    VAEncMiscParameterTypeRepeatFrame = 20,
+    /** \brief Buffer type used for HEVC Pak statistics */
+    VAEncMiscParameterTypeHevcPakStat = 21
```
dvrogozh, line 2323: Why only hevc? what about other codes - avc, av1, etc?

va/va.h
```diff
@@ -2316,7 +2316,11 @@ typedef enum {
     /** \brief Buffer type used for FEI input frame level parameters */
     VAEncMiscParameterTypeFEIFrameControl = 18,
     /** \brief encode extension buffer, ect. MPEG2 Sequence extenstion data */
-    VAEncMiscParameterTypeExtensionData = 19
+    VAEncMiscParameterTypeExtensionData = 19,
+    /** \brief Buffer type used for sending repeat frame flag to the encoder */
+    VAEncMiscParameterTypeRepeatFrame = 20,
```
XinfengZhang, line 2321: could we reuse skip frame?

PR #728, opened 2023-07-09
va/Android.mk: Fixes for compilation

muhammad23012009:


va/Android.mk
```diff
@@ -25,8 +25,8 @@
 
 LOCAL_PATH:= $(call my-dir)
 
-LIBVA_DRIVERS_PATH_32 := /vendor/lib/dri
-LIBVA_DRIVERS_PATH_64 := /vendor/lib64/dri
+LIBVA_DRIVERS_PATH_32 := /system/lib/dri
```
xmao4-intel, line 28: libva should be put at vendor partition since system partition is for android AOSP and can be updated seperately with vendor partition.

va/Android.mk
```diff
@@ -51,14 +51,14 @@ LOCAL_CFLAGS_64 += \
 
 LOCAL_CFLAGS := \
 	$(IGNORED_WARNNING) \
+	-DSYSCONFDIR=\"/system/etc\" \
```
xmao4-intel, line 54: this can be put at /vendor/etc/

xmao4-intel (commented):
libva is a vendor specified module and should be in vendor partition.

PR #730, opened 2023-07-11
api: extend va_deprecate to modern c/c++ and MSVC

dvrogozh:
Since libva was ported to Windows we should extend va_deprecate to work for MSVC where possible. Otherwise there might be confisions on which API is or is not deprecated.

We also can use advantages of modern c/c++ standards providing deprecating attributes.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -86,13 +86,22 @@
 extern "C" {
 #endif
 
-#if defined(__GNUC__) && !defined(__COVERITY__)
+#if defined(__cplusplus) && __cplusplus >= 201402L
+#define va_deprecated [[deprecated]]
+#define va_deprecated_enum [[deprecated]]
+#elif __STDC__ && __STDC_VERSION__ >= 202311L
+#define va_deprecated [[deprecated]]
+#define va_deprecated_enum [[deprecated]]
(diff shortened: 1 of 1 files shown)


va/va.h
```diff
@@ -86,13 +86,22 @@
 extern "C" {
 #endif
 
-#if defined(__GNUC__) && !defined(__COVERITY__)
+#if defined(__cplusplus) && __cplusplus >= 201402L
+#define va_deprecated [[deprecated]]
+#define va_deprecated_enum [[deprecated]]
+#elif __STDC__ && __STDC_VERSION__ >= 202311L
+#define va_deprecated [[deprecated]]
+#define va_deprecated_enum [[deprecated]]
```
evelikov, line 94: Don't know how much we care, but this will break Coverity Scan - both within this project and other projects that use libva. See 392187428cee9f5062c1b989cbe83a0a5e17c861 for details.

dvrogozh: Added notation is a part of C/C++ standard which tools should now or later be capable to understand. Whether Coverity understands above notation as of today - that's a question. I wonder how we can check before the merge? @uartie : do we have a way?

Other than that it's easy to modify PR to expand to nothing for coverity (as it was before).

PR #731, opened 2023-07-17
minor fixes for driver name rework

XinfengZhang:


LhGu (approved):

va/drm/va_drm_utils.c
```diff
@@ -89,8 +89,8 @@ VA_DRM_GetDriverNames(VADriverContextP ctx, char **drivers, unsigned *num_driver
         if (strcmp(map[i].drm_driver, drm_driver) == 0) {
             const char * const *va_drivers = map[i].va_driver;
 
-            while (va_drivers[count]) {
-                if (count < MAX_NAMES && count < *num_drivers)
+            while (count < MAX_NAMES && va_drivers[count]) {
+                if (count < *num_drivers)
                     drivers[count] = strdup(va_drivers[count]);
```
evelikov, line 94: The original code was essentially mixing two conflicting premises:
 - there will be a sentinel - aka the `while (va_drivers[count])`
 - there will not be a sentinel - aka `count < MAX_NAMES`

It would be a lot easier to read if we use

```
for ( ; count < MAX_NAMES && count < *num_drivers; count++)
    drivers[count] = strdup(va_drivers[count]);
```

Can we use ^^ instead? Plus, please prefix the commit message with `drm:`

evelikov (changes requested):

va/drm/va_drm_utils.c
XinfengZhang: > 
> Can we use ^^ instead? Plus, please prefix the commit message with `drm:`

yes, will change the commit message, but
```
for ( ; count < MAX_NAMES && count < *num_drivers; count++)
    drivers[count] = strdup(va_drivers[count]);
```
is not enough , if map[i].va_drivers only have 1 element.  the pointer map[i].va_drivers[1] is nullptr. 
we always need check drivers[count] is not null.

evelikov: Note to self: don't write any code before having coffee.

You're right. The following should fix that right?

```
for ( ; count < MAX_NAMES && va_drivers[count] && count < *num_drivers; count++)
    drivers[count] = strdup(va_drivers[count]);
```

evelikov: The API was aimed to take an array + its size as input. Then each implementation is should populate the array, _up-to_ the original length and return the actual length. If I'm understanding correctly it sounds like we had bug in the original.

That said, just had a cup of coffee so it would be better if I have another look + send a patch documenting the API later today.

Thanks

XinfengZhang: updated. thanks

evelikov:
Looks like other backends need a similar fix. Let me fix that up... /me writes a patch.

PR #732, opened 2023-07-20
va: Fix -Wl,--version-script check with LLD 17

vimproved:
In the upcoming LLD 17 (and previously in LLD 16.0.0, but reverted in 16.0.1), --no-undefined-version is becoming the default behavior. This causes the configure check for --version-scripts support to fail due to the symbols in the version script not being defined, which will cause the version script to not be used, and the build will fail due to --no-undefined-version. This commit adds '-Wl,--undefined-version' to the args of the configure check.

### DIFF
--- a/conftest.syms
+++ b/conftest.syms
@@ -1,0 +1,6 @@
+VERSION_1 {
+    global:
+        main;
+    local:
+        *;
+};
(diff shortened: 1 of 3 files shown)


XinfengZhang:
@vimproved , could you help to address the action issue?

vimproved:
> @vimproved , could you help to address the action issue?

Yes, sorry. I've been meaning to look at this for a while now, since I think this PR is not the best approach.

vimproved:
Updated this PR with a different method of fixing that should be more correct.

PR #735, opened 2023-07-24
Added Q416 fourcc for 16-bit YUV 4:4:4 planar format

thesword53:
Needed to support YUV 4:4:4 10 and 12 bits HEVC decoding for https://github.com/elFarto/nvidia-vaapi-driver.
Nvidia proprietary driver uses planar YUV 4:4:4 for 16 bits surfaces according to [cuviddec.h](https://github.com/NVIDIA/DALI/blob/7268fe15c50db664b4df5226b66cb46e717f46be/dali/operators/reader/loader/video/nvdecode/cuviddec.h#L91).

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -4583,6 +4583,11 @@ VAStatus vaSyncBuffer(
  * Four bytes per pixel: X, Y, U, V.
  */
 #define VA_FOURCC_XYUV          0x56555958
+/** Q416: three-plane 16-bit YUV 4:4:4.
+ *
+ * The three planes contain Y, U and V respectively.
+ */
+#define VA_FOURCC_Q416          0x36313451
 
 /* byte order */
 #define VA_LSB_FIRST        1


XinfengZhang (approved):
have no concern about VA_FOURCC('Q', '4', '1' '6')
but there are usage to define a planar 444 16 bit?

thesword53:
> but there are usage to define a planar 444 16 bit?

[nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver) is a VAAPI implementation on top of NVDEC (Nvidia proprietary hardware decode API) and it requires planar YUV444 16 bits surface to decode HEVC YUV444 10 and 12 bits videos and there is no fourcc in va.h for `yuv444p16` but only one for packed [YUVA 444 16 bits](https://github.com/intel/libva/blob/807044bd0fcb944edf230052fa09acc4266790fe/va/va.h#L4487C33-L4487C43).

> have no concern about VA_FOURCC('Q', '4', '1' '6')

Q416 fourcc is inspired from DRM_FORMAT_Q410 [drm_fourcc.h](https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm_fourcc.h) (10 bits planar YUV 444) in linux kernel headers.

PR #737, opened 2023-07-25
Unlock mutex before return

MicroYY:
Always release mutex before return.

Signed-off-by: Jay Yang <jay.yang@intel.com>

### DIFF
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -6335,12 +6335,14 @@ void va_TraceExportSurfaceHandle(
     va_TraceMsg(trace_ctx, "\tflags     = 0x%08x\n", flags);
 
     if (memType != VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2) {
+        DPY2TRACE_VIRCTX_EXIT(pva_trace);
         return;
     }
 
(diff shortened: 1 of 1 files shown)


Dylan-debug (approved):

XinfengZhang (approved):

va/va_trace.c
```diff
@@ -6344,6 +6334,16 @@ void va_TraceExportSurfaceHandle(
         return;
     }
 
+    int i;
+
+    DPY2TRACE_VIRCTX(dpy);
+
+    TRACE_FUNCNAME(idx);
+
+    va_TraceMsg(trace_ctx, "\tsurfaceId = 0x%08x\n", surfaceId);
```
dvrogozh, line 6343: I liked that these 3 items would be printed regardless of memType check.... Consider that someone will need to extend this function to cover other mem types - he will need to basically revert your change.

Thus, please, revert back the change and handle mutex issue in a different way, i.e. either call `DPY2TRACE_VIRCTX_EXIT` before each return or use `goto exit` instead of return.

dvrogozh (changes requested):

va/va_trace.c
MicroYY: Thanks. Updated.

dvrogozh (approved):
looks good

PR #740, opened 2023-08-02
Clarify the usage of flag VA_PROC_PIPELINE_FAST.

FurongZhang:


va/va_vpp.h
```diff
@@ -510,11 +510,15 @@ typedef struct _VABlendState {
...
- * decoded from a secondary bitstream.
+ * This flag is a hint to the video processor to achieve faster processing speed
+ * (better performance)in the pipeine. Hardware vendor might have several approaches
+ * for one functionality, this flag is to indicate the one with best performance.
+ * Performance depends on the hardware capablity, plesae refer to the specific hardware
+ * specification from hardware vendors. Here, lists a few examples as the reference.
+ * 1) When processing several surfaces, it may be necessary to prioritize more certain
+ * pipelines with this flag than others. 2) This flag is only a hint to the video processor
+ * so that it can omit certain filters for better performance for example. 3) Typically, this
+ * flag could be used with video surfaces decoded from a secondary bitstream.
```
HeJunyan, line 521: I think we should give more description about the relationship between this flag and the VA_FILTER_SCALING_FAST and VA_FILTER_SCALING_HQ flag here.

va/va_vpp.h
```diff
@@ -510,11 +510,20 @@ typedef struct _VABlendState {
...
+ *
+ * Regarding the usage of this flag, there are some filter properties in vaapi such as
+ * VA_FILTER_SCALING_XXX, when VA_PROC_PIPELINE_FAST is combined with these properties,
+ * the backend implementation should honor VA_PROC_PIPELINE_FAST which means this flag
+ * could change the filter properties implicitly for better performance.
```
xhaihao, line 526: Does it mean VA_FILTER_SCALING_XXX won't take effects if `VA_PROC_PIPELINE_FAST` is set ? How does a user know `VA_PROC_PIPELINE_FAST` should be set or not ?

FurongZhang: if VA_PROC_PIPELINE_FAST is set, VA_FILTER_SCALING_XXX will not take effect. If the user wants to get better performance, VA_PROC_PIPELINE_FAST should be set.

PR #743, opened 2023-08-16
trace: fix minor issue about printf data type and value range

XinfengZhang:


va/va_trace.c
```diff
@@ -1243,10 +1243,10 @@ static void va_TraceSurfaceAttributes(
                     for (j = 0; j < tmp->num_objects; j++) {
                         va_TraceMsg(trace_ctx, "\t\t\tobjects[%d].fd=%d\n", j, tmp->objects[j].fd);
                         va_TraceMsg(trace_ctx, "\t\t\tobjects[%d].size=%d\n", j, tmp->objects[j].size);
-                        va_TraceMsg(trace_ctx, "\t\t\tobjects[%d].drm_format_modifier=%d\n", j, tmp->objects[j].drm_format_modifier);
+                        va_TraceMsg(trace_ctx, "\t\t\tobjects[%d].drm_format_modifier=%llx\n", j, tmp->objects[j].drm_format_modifier);
                     }
                     va_TraceMsg(trace_ctx, "\t\t  num_layers=%d\n", tmp->num_layers);
-                    for (j = 0; j < tmp->num_layers; j++) {
+                    for (j = 0; j < tmp->num_layers && tmp->num_layers <= 4; j++) {
```
dvrogozh, line 1249: 1. You should be doing same for `num_objects`, right?
2. I think better to derive from real array size and avoid hardcoding 4 here:
```
tmp->num_layers <= sizeof(tmp->layers)/sizeof(tmp->layers[0])
```

dvrogozh (changes requested):

va/va_trace.c
XinfengZhang: > 1. You should be doing same for `num_objects`, right?

yes, correct, will update

> 2. I think better to derive from real array size and avoid hardcoding 4 here:

"4" comes from [VADRMPRIMESurfaceDescriptor](https://github.com/intel/libva/blob/master/va/va_drmcommon.h#L158) , it is fix value.

dvrogozh (approved):
looks good

PR #756, opened 2023-09-20
refine encode segment id map buffer

XinfengZhang:
attention: these change will break backward compatibility, the pre-assumption is no one is using them besides intel media driver and intel oneVPL runtime, if there are other customer,  it should be risky 
so, please help to check whether there are somebody are using segment id map buffer for avc or vp9 encoding.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1491,6 +1491,19 @@ typedef union _VAConfigAttribValEncPerBlockControl {
 /** \brief Driver supports decode processing rate report  */
 #define VA_PROCESSING_RATE_DECODE                     0x00000002
 /**@}*/
+
+/** @name segment ID map block size */
+/**@{*/
+/** \brief each segmentID represent a 16x16 block */
+#define VA_SEGID_BLOCK_16X16                          0
+/** \brief each segmentID represent a 32x32 block */
+#define VA_SEGID_BLOCK_32X32                          1
+/** \brief each segmentID represent a 64x64 block */
+#define VA_SEGID_BLOCK_64X64                          2
+/** \brief each segmentID represent a 8x8 block */
+#define VA_SEGID_BLOCK_8X8                            3
+/**@}*/
+
 /**
  * if an attribute is not applicable for a given
  * profile/entrypoint pair, then set the value to the following
--- a/va/va.h
+++ b/va/va.h
@@ -1053,6 +1053,14 @@ typedef enum {
      * columns supported for encoding with tile support.
      */
     VAConfigAttribEncMaxTileCols        = 57,
+    /**
+     * \brief VP9 encoding attribute. Read-only.
+     *
+     * This attribute exposes a number of capabilities of the underlying
+     * VP9 implementation. The attribute value is partitioned into fields as defined in the
+     * VAConfigAttribValEncVP9 union.
+     */
+    VAConfigAttribEncVP9                = 58,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
--- a/va/va_enc_av1.h
+++ b/va/va_enc_av1.h
@@ -148,8 +148,10 @@ typedef union _VAConfigAttribValEncAV1Ext1 {
          */
         uint32_t interpolation_filter          : 5;
         /**
-         * \brief Min segmentId block size accepted.
+         * \brief segmentId block size accepted.
          * Application need to send seg_id_block_size in PPS equal or larger than this value.
+         * one bit represent one block size defined as VA_SEGID_BLOCKXXXX
+         * should be (1 << VA_SEGID_BLOCKXXX | 1 << VA_SEGID_BLOCKXXX ... )
          */
         uint32_t min_segid_block_size_accepted : 8;
         /**
@@ -667,10 +669,7 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
     } picture_flags;
 
     /** \brief Block size for each Segment ID in Segment Map.
-     *  0: 16x16 block size, default value;
-     *  1: 32x32 block size;
-     *  2: 64x64 block size;
-     *  3: 8x8 block size.
+     *  should be \c VA_SEGID_BLOCK_XXXX;
      */
     uint8_t     seg_id_block_size;
 
--- a/va/va_enc_vp9.h
+++ b/va/va_enc_vp9.h
@@ -595,8 +595,10 @@ typedef struct _VAEncMiscParameterTypeVP9PerSegmantParam {
  * \brief VP9 Block Segmentation ID Buffer
  *
  * The application provides a buffer of VAEncMacroblockMapBufferType containing
- * the initial segmentation id for each 8x8 block, one byte each, in raster scan order.
- * Rate control may reassign it.  For example, a 640x480 video, the buffer has 4800 entries.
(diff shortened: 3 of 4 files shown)


va/va_enc_av1.h
```diff
@@ -426,19 +428,19 @@ typedef struct _VAEncSegParamAV1 {
...
-     * in this map contains the segment id of a particular block.
-     * The element is indexed by raster scan order.
-     * The value of each entry should be in the range [0..7], inclusive.
-     */
-    uint8_t    *pSegmentMap;
-} VAEncSegMapBufferAV1;
+/**
+ * \brief AV1 Block Segmentation ID Buffer
+ *
+ * The application provides a buffer of VAEncMacroblockMapBufferType containing
```
rosetta-jpn, line 435: Do we need to set size and num_elements in vaCreateBuffer to width and height as well as in vp9, respectively?

va/va.h
```diff
@@ -1474,6 +1482,17 @@ typedef union _VAConfigAttribValEncPerBlockControl {
 /** \brief Driver supports decode processing rate report  */
 #define VA_PROCESSING_RATE_DECODE                     0x00000002
 /**@}*/
+
+/** @name segment ID map block size */
+/**@{*/
```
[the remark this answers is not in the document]
XinfengZhang: if we add them, to keep some basic compatible , the value should be 3, 4,  it looks a little strange, I suspect who are using them.

va/va_enc_av1.h
XinfengZhang: it is a kind of backward combability issue, I want to unify them to a linear buffer. so, the question is: is there other user beside intel media driver , is there any application is using the older interface?

leyu-yao: Would it be a better solution if we keep the original behavior when calling vaCreateBuffer with VAEncMacroblockMapBufferType, and we define a new BufferType (e.g. named VAEncUnifiedSegmentationMapBufferType). 
There are some benefits but libva will a little bit complicated.
We can keep all the backward compability and suggest/guide app-developers to switch to the new/unified interface.

XinfengZhang: so, we support both for VP9?

leyu-yao: Yes, I would prefer that. 
It would benefit the application maintenance effort. 
However, it will make libva more complicated......

va/va.h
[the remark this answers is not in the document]
rosetta-jpn: But segmentation is 8x8 in vp9 and 4x4 in av1. So a driver should be able to support up the size?

va/va_enc_av1.h
```diff
@@ -148,8 +148,10 @@ typedef union _VAConfigAttribValEncAV1Ext1 {
          */
         uint32_t interpolation_filter          : 5;
         /**
-         * \brief Min segmentId block size accepted.
+         * \brief segmentId block size accepted.
          * Application need to send seg_id_block_size in PPS equal or larger than this value.
+         * one bit represent one block size defined as VA_SEGID_BLOCKXXXX
+         * should be (1 << VA_SEGID_BLOCKXXX | 1 << VA_SEGID_BLOCKXXX ... )
          */
         uint32_t min_segid_block_size_accepted : 8;
```
rosetta-jpn, line 156: Can we remove min_ from the variable name?

va/va_enc_av1.h
```diff
@@ -670,7 +672,6 @@ typedef struct  _VAEncPictureParameterBufferAV1 {
      *  0: 16x16 block size, default value;
      *  1: 32x32 block size;
      *  2: 64x64 block size;
-     *  3: 8x8 block size.
```
rosetta-jpn, line 673: Why is this removed? 
For the consisency to vp9, should this be VA_SEGID_BLOCKXXX?

va/va.h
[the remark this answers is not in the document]
rosetta-jpn: Ping. How do you think about this comment?

rosetta-jpn (changes requested):

va/va.h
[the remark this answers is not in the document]
XinfengZhang: I see , but we could add it later, at least , there are no such support from HW vendors?

va/va_enc_av1.h
XinfengZhang: change the name will cause API compatible issue.

rosetta-jpn: Ack.

va/va_enc_av1.h
XinfengZhang: updated.

XinfengZhang:
@rosetta-jpn , updated. how about this version?

va/va_enc_vp9.h
```diff
@@ -587,13 +595,41 @@ typedef struct _VAEncMiscParameterTypeVP9PerSegmantParam {
  * \brief VP9 Block Segmentation ID Buffer
  *
  * The application provides a buffer of VAEncMacroblockMapBufferType containing
- * the initial segmentation id for each 8x8 block, one byte each, in raster scan order.
- * Rate control may reassign it.  For example, a 640x480 video, the buffer has 4800 entries.
+ * the initial segmentation id for each block, block size is specified by seg_id_block_size
+ * in VAEncPictureParameterBufferVP9  , one byte each, in raster scan order.
+ * Rate control may reassign it.  For example, a 640x480 video, set_id_block_size is
```
leyu-yao, line 600: There is a typo...  set_id_block_size

va/va_enc_vp9.h
```diff
@@ -587,13 +595,41 @@ typedef struct _VAEncMiscParameterTypeVP9PerSegmantParam {
...
+/** \brief Attribute value for VAConfigAttribEncVP9. */
+typedef union _VAConfigAttribValEncVP9 {
+    struct {
+        /**
+         * \brief segmentId block size accepted.
+         * This is the granularity of segmentation map.
+         * one bit represent one block size defined as VA_SEGID_BLOCKXXXX
+         * should be (1 << VA_SEGID_BLOCKXXX | 1 << VA_SEGID_BLOCKXXX ... )
+         */
+        uint32_t segid_block_size_accepted :    8;
```
leyu-yao, line 616: Av1 is using "seg_id_block_size", should we keep it consistent?

leyu-yao (approved):

va/va.h
```diff
@@ -1491,6 +1491,19 @@ typedef union _VAConfigAttribValEncPerBlockControl {
...
+/** \brief each segmentID represent a 16x16 block */
+#define VA_SEGID_BLOCK_16X16                          0
+/** \brief each segmentID represent a 32x32 block */
+#define VA_SEGID_BLOCK_32X32                          1
+/** \brief each segmentID represent a 64x64 block */
+#define VA_SEGID_BLOCK_64X64                          2
+/** \brief each segmentID represent a 8x8 block */
+#define VA_SEGID_BLOCK_8X8                            3
```
rosetta-jpn, line 1512: nit: Shall we add VA_SEGID_BLOCK_4x4? AV1 is 4x4 seg map.

rosetta-jpn (approved):

rosetta-jpn:
Sorry for so delayed review.
LGTM % nit. Feel free to submit the patches.

PR #757, opened 2023-09-22
va: Add VAConfigAttribEncMaxTileRowsCols

sivileri:
Currently there is no way of exposing the driver tile_rows and tile_cols limitations for encoding with tiles. For codecs like AV1 there is a cap specifying the max number of tiles, but without differentiating cols/rows.

Different hardware may have restrictions which may need to be taken into account. More specifically, D3D12 encode exposes the maximum rows/cols and adding VAConfigAttribEncMaxTileRowsCols allows for mapping that information to the VAOn12 driver.

@dvrogozh @XinfengZhang


va/va.h
```diff
@@ -1036,6 +1036,16 @@ typedef enum {
      * The value returned uses the VAConfigAttribValEncPerBlockControl type.
      */
     VAConfigAttribEncPerBlockControl    = 55,
+    /**
+     * \brief Maximum number of tile rows and cols. Read-only.
+     *
+     * This attribute determines the maximum number of tile
+     * rows and columns supported for encoding with tile support.
+     *
+     * The bottom 16 bits specify the maximum number of rows
```
dvrogozh, line 1045: Is there an actual need to complicate working with this attribute and/or this aligns to something else in VAAPI? I mean, maybe separate `VAConfigAttribEncMaxTileRows` and `VAConfigAttribEncMaxTileCols` will be easier for user?

sivileri: Thanks for taking a look @dvrogozh . This pattern is used on other caps like `VAConfigAttribEncMaxRefFrames`, and then vainfo prints both values by parsing the low/high bit regions. I'm open to splitting it into `VAConfigAttribEncMaxTileRows` and `VAConfigAttribEncMaxTileCols` is that's the consensus, but I think max tile/cols should really be a single cap reported by driver with two values.

```
    /**
     * \brief Maximum number of reference frames. Read-only.
     *
     * This attribute determines the maximum number of reference
     * frames supported for encoding.
     *
     * Note: for H.264 encoding, the value represents the maximum number
     * of reference frames for both the reference picture list 0 (bottom
     * 16 bits) and the reference picture list 1 (top 16 bits).
     */
    VAConfigAttribEncMaxRefFrames       = 13,
```

sivileri: @dvrogozh I pushed the change to split them into two separate queries. Taking deeper look I found other existing caps that are queried separately and I agree with you it may be easier for the user to just get the value without having to parse the low/high bits. 

ps. Also updated the vainfo PR: https://github.com/intel/libva-utils/pull/344

va/va.h
```diff
@@ -1036,6 +1036,20 @@ typedef enum {
      * The value returned uses the VAConfigAttribValEncPerBlockControl type.
      */
     VAConfigAttribEncPerBlockControl    = 55,
+    /**
+     * \brief Maximum number of tile rows. Read-only.
+     *
+     * This attribute determines the maximum number of tile
+     * rows supported for encoding with tile support.
+     */
+    VAConfigAttribEncMaxTileRows    = 56,
```
dvrogozh, line 1045: minor: fix alignment for `= 56`

va/va.h
```diff
@@ -1036,6 +1036,20 @@ typedef enum {
...
+    /**
+     * \brief Maximum number of tile cols. Read-only.
+     *
+     * This attribute determines the maximum number of tile
+     * columns supported for encoding with tile support.
+     */
+    VAConfigAttribEncMaxTileCols    = 57,
```
dvrogozh, line 1052: same

va/va.h
dvrogozh: > ps. Also updated the vainfo PR: https://github.com/intel/libva-utils/pull/344

Was about to ask. Thanks for raising up.

dvrogozh (approved):
looks good

XinfengZhang (commented):
one value is enough? 16 bit for each?

dvrogozh:
These attributes are not directly involved in bitstream packing. In such cases I think it's better to favor usability over compactness.

sivileri:
@XinfengZhang kindly pinging on this MR :)

XinfengZhang (approved):
LGTM,      VA_ATTRIB_NOT_SUPPORTED represent " no limitation" "comply with codec spec" right?

PR #763, opened 2023-10-25
va: export symbol vaMapBuffer2 for Windows

tong1wu:
This new added api should be exported otherwise it will cause link issues.

### DIFF
--- a/va/libva.def
+++ b/va/libva.def
@@ -32,6 +32,7 @@ EXPORTS
     vaGetDisplayAttributes
     vaGetImage
     vaMapBuffer
+    vaMapBuffer2
     vaMaxNumDisplayAttributes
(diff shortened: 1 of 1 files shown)


xhaihao (approved):
LGTM

va/libva.def
```diff
@@ -32,6 +32,7 @@ EXPORTS
     vaGetDisplayAttributes
     vaGetImage
     vaMapBuffer
+    vaMapBuffer2
```
dvrogozh, line 35: Ups. Yes.

dvrogozh (approved):

XinfengZhang (approved):

dvrogozh:
Non related to this PR, but something's wrong with freebsd ci... This is not on libva side though...

PR #767, opened 2023-10-31
Added 3DLUT interpolation method.

FurongZhang:


va/va_vpp.h
```diff
@@ -1426,8 +1436,12 @@ typedef struct _VAProcFilterParameterBuffer3DLUT {
...
     uint32_t            channel_mapping;
 
+    /** \brief interpolation_method defines the 3DLUT interpolation method, could be one of VA_3DLUT_INTERPOLATION_XXX*/
+    uint32_t            interpolation_method : 4;
+
     /** \brief reserved bytes for future use, must be zero */
-    uint32_t            va_reserved[VA_PADDING_HIGH];
+    uint32_t            va_reserved1 : 28;
+    uint32_t            va_reserved[VA_PADDING_HIGH - 1];
 } VAProcFilterParameterBuffer3DLUT;
```
LhGu, line 1446: For HDR10， If app needs to enable 3DLut TETRAHEDRAL interpolation, besides VAProcFilterHighDynamicRangeToneMapping filter type,  they also need to add VAProcFilter3DLUT because we only add this parameter in VAProcFilterParameterBuffer3DLUT. right?

LhGu (approved):

va/va_vpp.h
FurongZhang: App either use VAProcFilterHighDynamicRangeToneMapping with HDR10 metadata, or use VAProcFilterParameterBuffer3DLUT by using their own 3DLUT. Those 2 filters are usually not combined to use.

PR #714, opened 2023-06-05
Add new error code to indicate the reset happended in the driver.

pengxin99:
Use VA_STATUS_ERROR_RESET to indeicate media reset/engine reset have happended in the driver, user can use vaQuerySurfaceError() to get the detail info to do some customized work.

PR #774, opened 2023-11-27
Add description of the return status for vaSyncSurface()

pengxin99:
this will indicate APP behavior when meet VA_STATUS_ERROR_DECODING_ERROR from vaSyncSurface

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -4187,6 +4187,26 @@ Synchronization
  * This function blocks until all pending operations on the render target
  * have been completed.  Upon return it is safe to use the render target for a
  * different picture.
+ *
+ * Possible errors:
+ * - \ref VA_STATUS_ERROR_INVALID_CONTEXT: This error occurs when the function is provided with an invalid VADisplay(dpy).
+ *   It is typically due to the VADriverContext contained by VADisplay is not being properly initialized or already being freed.
+ *   To address this, ensure that the VADriverContext or VADisplay has been properly initialized and has not been freed.
+ * - \ref VA_STATUS_ERROR_INVALID_SURFACE: This error is triggered when the supplied render target is invalid,
+ *   possibly due to being freed or corrupted. To resolve this, ensure the usage of the correct render target
+ *   and verify its validity and integrity.
+ * - \ref VA_STATUS_ERROR_OPERATION_FAILED: When the attempt to retrieve the status report fails or related work
+ *   on the current surface fails, this error is generated. In such cases, the client can proceed with subsequent
+ *   frames by ignore the error, but it may involve incorrect result like corruption, MD5 mismatch or HW hang.
+ *   Initializing new context can avoid the potential issues.
+ * - \ref VA_STATUS_ERROR_HW_BUSY: This error indicates that synchronization is still in progress or GPU hang happens.
+ *   The client could call the function again with the same render_target within a period of time to wait hw execution completion.
+ *   If hw execution could not complete successfully, gpu hang and reset may occur, in this case,
+ *   application could try to recover media context by initializing new one.
+ * - \ref VA_STATUS_ERROR_DECODING_ERROR: This error is encountered when macroblock (MB) errors with
+ *   non-conformance input clips are detected during the decoding process. The application can proceed with
+ *   subsequent frames, but it is advised to be aware of potential corruption in the output.
+ *   Call vaQuerySurfaceError to get more details on mb error.
  */
 VAStatus vaSyncSurface(
     VADisplay dpy,


va/va.h
```diff
@@ -4187,6 +4187,12 @@ Synchronization
  * This function blocks until all pending operations on the render target
  * have been completed.  Upon return it is safe to use the render target for a
  * different picture.
+ *
+ * Possible errors:
+ * - \ref VA_STATUS_ERROR_DECODING_ERROR: This error is triggered when encountering
+ *   Macroblock (MB) errors with non-conformance input clips during the decoding process.
+ *   Application can proceed with the subsequent frames, it is advised that the application
+ *   be aware of potential corruption in the output.
```
Jexu, line 4195: check the call stack in media driver to list all possibe error return?
btw, it is for both decode and vp.

pengxin99: have added all possible error for decoding.

pengxin99:
Hi, @XinfengZhang @Jexu, have add more detail description, could you pls help to review this PR again?

Jexu (approved):

va/va.h
```diff
@@ -4187,6 +4187,25 @@ Synchronization
...
+ * Possible errors:
+ * - \ref VA_STATUS_ERROR_INVALID_CONTEXT: This error occurs when the function is provided with an invalid context.
+ *   It is typically due to the context not being properly initialized or already being freed.
+ *   To address this, ensure that the context is correctly initialized and has not been freed.
+ * - \ref VA_STATUS_ERROR_INVALID_SURFACE: This error is triggered when the supplied render target is invalid,
+ *   possibly due to being freed or corrupted. To resolve this, ensure the usage of the correct render target
+ *   and verify its validity and integrity.
+ * - \ref VA_STATUS_ERROR_OPERATION_FAILED: When the attempt to retrieve the status report fails or related work
+ *   on the current surface fails, this error is generated. In such cases, the client can proceed with subsequent
+ *   frames by ignore the error, but it may involve incorrect result like corruption.
```
XinfengZhang, line 4200: it means "operation failed" could be ignored by application

va/va.h
```diff
@@ -4187,6 +4187,25 @@ Synchronization
...
+ * - \ref VA_STATUS_ERROR_HW_BUSY: This error indicates that synchronization is still in progress or GPU hang happens.
+ *   The client could call the function again within a period of time to wait hw execution completion.
```
XinfengZhang, line 4202: same surfaceID again?

va/va.h
```diff
@@ -4187,6 +4187,25 @@ Synchronization
...
+ *   If hw execution could not complete successfully, gpu hang and reset may occur, in this case,
+ *   application could try to recover media context by initialing new one.
```
XinfengZhang, line 4204: must recreate context? could application just ignore it?

va/va.h
```diff
@@ -4187,6 +4187,25 @@ Synchronization
...
+ * - \ref VA_STATUS_ERROR_DECODING_ERROR: This error is encountered when macroblock (MB) errors with
+ *   non-conformance input clips are detected during the decoding process. The application can proceed with
+ *   subsequent frames, but it is advised to be aware of potential corruption in the output.
+ *   Call vaQuerySurfaceError could get more details of mb error information.
```
XinfengZhang, line 4208: difference comparing with  VA_STATUS_ERROR_OPERATION_FAILED

XinfengZhang:
I agree with the direction, provide hint to application , then application could handle correlated errors , ignore , or reset the context etc. .  
but I suggest to clarify the difference. currently:
VA_STATUS_ERROR_DECODING_ERROR may caused by a bistream corruption. related with syntax
VA_STATUS_ERROR_HW_BUSY is about the HW reset , but looks could not distinguish , media reset or gpu reset. if media reset , it could be ignored directly.  it also related with #714?

but operation failed is unclear.

va/va.h
pengxin99: VA_STATUS_ERROR_OPERATION_FAILED is caused by status report fail which cannot be used to determine whether the result is correct. So declare here that APP can ignore this error but may meet incorrect result.

And will add more detail like _"To init new context to avoid protentional incorrect result."_

va/va.h
pengxin99: Yes, will add it

va/va.h
pengxin99: I think the recreate context is must, VA_STATUS_ERROR_HW_BUSY indicate that HW reset happened, this need to recover media context. Media reset will not return this error.

pengxin99:
@XinfengZhang Thanks for the comments.   and I will refine this PR for below two point:

> VA_STATUS_ERROR_HW_BUSY is about the HW reset , but looks could not distinguish , media reset or gpu reset. if media reset , it could be ignored directly. it also related with https://github.com/intel/libva/pull/714?

The VA_STATUS_ERROR_HW_BUSY is only for HW reset, so cannot ignore this error, vaSyncSurface will return success if media reset enable. And I just find the description to be inaccurate in https://github.com/intel/libva/pull/714, will also refine int this PR.

>but operation failed is unclear.

can see the below comment: https://github.com/intel/libva/pull/774#discussion_r1449800161

va/va.h
```diff
@@ -4187,6 +4187,26 @@ Synchronization
...
+ * - \ref VA_STATUS_ERROR_INVALID_CONTEXT: This error occurs when the function is provided with an invalid context.
+ *   It is typically due to the context not being properly initialized or already being freed.
+ *   To address this, ensure that the context is correctly initialized and has not been freed.
+ * - \ref VA_STATUS_ERROR_INVALID_SURFACE: This error is triggered when the supplied render target is invalid,
+ *   possibly due to being freed or corrupted. To resolve this, ensure the usage of the correct render target
+ *   and verify its validity and integrity.
+ * - \ref VA_STATUS_ERROR_OPERATION_FAILED: When the attempt to retrieve the status report fails or related work
+ *   on the current surface fails, this error is generated. In such cases, the client can proceed with subsequent
+ *   frames by ignore the error, but it may involve incorrect result like corruption, MD5 mismatch or HW hang.
+ *   Initialize new context can avoid the potential issues.
```
dvrogozh, line 4201:
```suggestion
 *   Initializing new context can avoid the potential issues.
```

va/va.h
```diff
@@ -4187,6 +4187,26 @@ Synchronization
...
+ * - \ref VA_STATUS_ERROR_HW_BUSY: This error indicates that synchronization is still in progress or GPU hang happens.
+ *   The client could call the function again with the same render_target within a period of time to wait hw execution completion.
+ *   If hw execution could not complete successfully, gpu hang and reset may occur, in this case,
+ *   application could try to recover media context by initialing new one.
```
dvrogozh, line 4205:
```suggestion
 *   application could try to recover media context by initializing new one.
```

va/va.h
```diff
@@ -4187,6 +4187,26 @@ Synchronization
...
+ * - \ref VA_STATUS_ERROR_DECODING_ERROR: This error is encountered when macroblock (MB) errors with
+ *   non-conformance input clips are detected during the decoding process. The application can proceed with
+ *   subsequent frames, but it is advised to be aware of potential corruption in the output.
+ *   Call vaQuerySurfaceError could get more details of mb error information.
```
dvrogozh, line 4209:
```suggestion
 *   Call vaQuerySurfaceError to get more details on mb error.
```

va/va.h
```diff
@@ -4187,6 +4187,26 @@ Synchronization
  * This function blocks until all pending operations on the render target
  * have been completed.  Upon return it is safe to use the render target for a
  * different picture.
+ *
+ * Possible errors:
+ * - \ref VA_STATUS_ERROR_INVALID_CONTEXT: This error occurs when the function is provided with an invalid context.
```
dvrogozh, line 4192: > This error occurs when the function is provided with an invalid context.

Function prototype does not have context at all. So, this error is inadequate ( `VA_STATUS_ERROR_INVALID_DISPLAY` seems more reasonable) or at least incorrectly described. `VAStatus vaSyncSurface(VADisplay dpy, VASurfaceID render_target)`.

dvrogozh (changes requested):

dvrogozh:
Also, please, squash commits. For this change there should be just 1 commit in the PR.

va/va.h
pengxin99: thx, fixed

va/va.h
pengxin99: thx, fixed

va/va.h
pengxin99: thx, fixed

va/va.h
pengxin99: thanks for this advice, the media-driver will return `VA_STATUS_ERROR_INVALID_CONTEXT `instead of `VA_STATUS_ERROR_INVALID_DISPLAY `, so I only correct the comments here to make sure the API compatible with the old version.

pengxin99:
> Also, please, squash commits. For this change there should be just 1 commit in the PR.

Thanks for your review, and have refined this PR, could pls have a look again?

pengxin99:
Hi, @dvrogozh @XinfengZhang , Can you take another look at this PR if you have time?

dvrogozh (approved):
looks good

pengxin99:
Hi, @XinfengZhang , Could you pls take another look at this PR when you have time?

PR #779, opened 2023-12-11
[TEST] promote version test

lixin5x:


XinfengZhang (commented):
please remove the version promotion patch,  the version only bumped when release.

lixin5x:
> please remove the version promotion patch, the version only bumped when release.

This is for test build, will not checkin.

PR #783, opened 2023-12-21
va: export symbol vaGetLibFunc for Windows

### LINKED ISSUE #782: vaGetLibFunc is not exported in libva.def
vaGetLibFunc is not exported in libva.def, it causes that APP can't use VA extension function on Windows.

sivileri:
Add symbol for exporting existing VA vaGetLibFunc function to .def file

Fixes #782


sivileri:
cc @XinfengZhang

XinfengZhang (approved):

PR #788, opened 2024-01-07
va: fix handling when all wayland backends fail

tleydxdy:
when all backend failed vaGetDisplayWl will just return a context that was already destroyed. This leads to crashes if downstream tries to use the context.

I'm not familiar with the unit test infra but it may make sense to add a unit test here..


XinfengZhang (approved):

tleydxdy:
oops I didn't mean to update this pr, just picking that commit to try it out myself...

emersion (approved):

emersion:
Is there anything holding off this patch?

PR #789, opened 2024-01-08
wayland: fix segfault when all backends fail

emersion:
When all Wayland backends fail, vaGetDisplayWl() would still return a half-constructed display context, resulting in a sefault in va_DRM_GetDrmDriverName() later on when trying to access ctx->drm_state->fd.

Instead, cleanup and return NULL.


XinfengZhang:
duplicate with #788 ?

emersion:
Indeed.

emersion:
Superseded by #788

PR #790, opened 2024-01-08
wayland: add support for linux-dmabuf

emersion:
wl_drm is a legacy protocol, and wlroots is getting rid of it [1]. Use the newer and standard linux-dmabuf protocol if available to get the DRM device.

[1]: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4397

### DIFF
--- a/va/wayland/va_wayland_linux_dmabuf.c
+++ b/va/wayland/va_wayland_linux_dmabuf.c
@@ -1,0 +1,364 @@
+/*
+ * va_wayland_drm.c - Wayland/linux-dmabuf helpers
+ *
+ * Copyright (c) 2024 Simon Ser
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sub license, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL INTEL AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include "sysdeps.h"
+#include <unistd.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <dlfcn.h>
+#include <sys/stat.h>
+#include <xf86drm.h>
+#include "va_drmcommon.h"
+#include "drm/va_drm_utils.h"
+#include "va_wayland_linux_dmabuf.h"
+#include "va_wayland_private.h"
+#include "linux-dmabuf-v1-client-protocol.h"
+
+typedef struct va_wayland_linux_dmabuf_context {
+    struct va_wayland_context base;
+    bool                      has_linux_dmabuf;
+    bool                      default_feedback_done;
+} VADisplayContextWaylandLinuxDmabuf;
+
+static void
+feedback_handle_done(
+    void                                *data,
+    struct zwp_linux_dmabuf_feedback_v1 *feedback
+)
+{
+    VADisplayContextP const pDisplayContext = data;
+    struct va_wayland_linux_dmabuf_context *wl_linux_dmabuf_ctx = pDisplayContext->opaque;
+
+    wl_linux_dmabuf_ctx->default_feedback_done = true;
+
+    zwp_linux_dmabuf_feedback_v1_destroy(feedback);
+}
+
+static void
+feedback_handle_format_table(
+    void                                *data,
+    struct zwp_linux_dmabuf_feedback_v1 *feedback,
+    int                                  fd,
+    uint32_t                             size
+)
+{
+    close(fd);
+}
+
+/* XXX: replace with drmGetDeviceFromDevId() */
+static drmDevice *
+get_drm_device_from_dev_id(dev_t dev_id)
+{
+    uint32_t flags = 0;
+    int devices_len, i, node_type;
+    drmDevice *match = NULL, *dev;
+    struct stat statbuf;
+
+    devices_len = drmGetDevices2(flags, NULL, 0);
+    if (devices_len < 0) {
+        return NULL;
+    }
+    drmDevice **devices = calloc(devices_len, sizeof(*devices));
+    if (devices == NULL) {
+        return NULL;
+    }
+    devices_len = drmGetDevices2(flags, devices, devices_len);
+    if (devices_len < 0) {
+        free(devices);
+        return NULL;
+    }
+
+    for (i = 0; i < devices_len; i++) {
+        dev = devices[i];
+        for (node_type = 0; node_type < DRM_NODE_MAX; node_type++) {
+            if (!(dev->available_nodes & (1 << node_type)))
+                continue;
+
+            if (stat(dev->nodes[node_type], &statbuf) != 0) {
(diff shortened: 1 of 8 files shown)


emersion:
Ubuntu 20.04 has a very old libdrm. Should I make this new backend optional, or should I bump the Ubuntu version, or something else?

emersion:
Gentle ping, any chance to get this reviewed?

va/wayland/va_wayland_linux_dmabuf.c
```diff
@@ -0,0 +1,312 @@
...
+    drmDevice *dev;
+    const char *dev_path;
+    VADisplayContextP const pDisplayContext = data;
+    VADriverContextP const ctx = pDisplayContext->pDriverContext;
+    struct drm_state * const drm_state = ctx->drm_state;
+
+    assert(device_array->size == sizeof(dev_id));
+    memcpy(&dev_id, device_array->data, sizeof(dev_id));
+
+    if (drmGetDeviceFromDevId(dev_id, 0, &dev) < 0) {
```
XinfengZhang, line 88: only this function need bump libdrm version?
why not use drmGetDevice2 or drmGetDevices2

emersion: Yeah, if we want to keep the current libdrm version requirement, we can hand-roll our own `drmGetDeviceFromDevId()` implementation (iterating over all devices and checking their device ID). However there is an additional complication: this patch requires wayland-protocols 1.24, and Ubuntu 20.04 only has [wayland-protocols 1.20](https://packages.ubuntu.com/focal/wayland-protocols).

I see two ways to fix this:

- Make the new backend optional, and keep the dependencies in this PR.
- Hand-roll `drmGetDeviceFromDevId()` and also vendor a copy of the [linux-dmabuf XML file](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/linux-dmabuf/linux-dmabuf-v1.xml?ref_type=heads).

Which one do you prefer?

Conan-Kudo: Bumping to Ubuntu 22.04 is also a reasonable choice, especially since Ubuntu 20.04 will be EOL next year.

XinfengZhang: both looks workable , I prefer 2nd,  because:
first solution actually have 2 separate steps: 1. optional 2. default enabled.   we actually delayed the version bump.  
a potential risk for 2nd solution is: whether the [linux-dmabuf XML file](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/linux-dmabuf/linux-dmabuf-v1.xml?ref_type=heads).will be updated and be incompatible in future?

Conan-Kudo: It will not. That's kind of the point of it being in "stable".

emersion: I've updated the patch to hand-roll `drmGetDeviceFromDevId()` and vendor a copy of the linux-dmabuf XML.

emersion:
Gentle ping, would you have the time to have another look at this?

va/wayland/va_wayland_linux_dmabuf.c
```diff
@@ -0,0 +1,364 @@
...
+    const char         *interface,
+    uint32_t            version
+)
+{
+    VADisplayContextP const pDisplayContext = data;
+    struct va_wayland_linux_dmabuf_context *wl_linux_dmabuf_ctx = pDisplayContext->opaque;
+    struct zwp_linux_dmabuf_v1 *linux_dmabuf;
+    struct zwp_linux_dmabuf_feedback_v1 *feedback;
+
+    if (strcmp(interface, zwp_linux_dmabuf_v1_interface.name) == 0 &&
```
XinfengZhang, line 220: TBH, I am not familiar with this interface, but from code logic, 
looks , it could be combined together with va_wayland_drm? add interface check inside registry_handle_global
if it is linuxdmabuf ,xxxx ; if it is wl_drm xxxx?

from libva perspective: it should retrieve the device fd though these function, then get the driver name though device fd.  looks both wl_drm and linuxdma buffer could do it.  but wl_drm interface will be deprecated?

from driver implementation perspective. what's the difference to implement vaGetSurfaceBufferWl and vaGetImageBufferWl? previously, we use "wl_proxy_create" and "wl_proxy_marshal" interface to share the dma buffer fd from vaSurface to wl_buffer, just like https://github.com/intel/intel-vaapi-driver/blob/master/src/i965_output_wayland.c#L231-L263.  looks we need new implementation now?

emersion: > looks , it could be combined together with va_wayland_drm? add interface check inside registry_handle_global
if it is linuxdmabuf ,xxxx ; if it is wl_drm xxxx?

That's possible indeed, but not sure it's a good idea. It would entangle the two implementations and make it more complicated to understand the code and drop the `wl_drm` part in the future.

> from libva perspective: it should retrieve the device fd though these function, then get the driver name though device fd. looks both wl_drm and linuxdma buffer could do it. but wl_drm interface will be deprecated?

Yes, `wl_drm` was never supposed to be used by clients such as libva, it was supposed to be an internal Mesa protocol.

Some Wayland compositors like Sway have already dropped support for `wl_drm`.

> from driver implementation perspective. what's the difference to implement vaGetSurfaceBufferWl and vaGetImageBufferWl? previously, we use "wl_proxy_create" and "wl_proxy_marshal" interface to share the dma buffer fd from vaSurface to wl_buffer, just like https://github.com/intel/intel-vaapi-driver/blob/master/src/i965_output_wayland.c#L231-L263. looks we need new implementation now?

Hm, yeah. This would need a new implementation indeed.

mpv seems to work just fine without this new implementation though. Maybe because they grab the DMA-BUFs directly, instead of `vaGetImageBufferWl`/`vaGetSurfaceBufferWl`?

There is nothing driver-specific in `vaGetSurfaceBufferWl`/`vaGetImageBufferWl` by the way. The implementation should be identical for all VA-API drivers.

emersion: FWIW, it seems like Mesa doesn't implement `VADriverVTableWayland`.

emersion:
Gentle ping

emersion:
Gentle ping

emersion:
Any chance you'd have time to look at this again @XinfengZhang @xhaihao?

XinfengZhang:
> Any chance you'd have time to look at this again @XinfengZhang @xhaihao?

sorry for slow response, TBH, I am not familiar with this part, so, I plan to release 2.21.0, then pay some time to check the details.  then merge it and keep it in master branch , and it will be a part of next release.  suppose I could finish at early of next week. 
is it urgent for your usage?

emersion:
Thanks for the reply. It's not super urgent, but the next wlroots release planned in May will not have `wl_drm`, so would be nice to have the fix released by then!

emersion:
Hi again, please have another look at this PR when you get this chance.

va/wayland/va_wayland_linux_dmabuf.c
```diff
@@ -0,0 +1,364 @@
...
+    dev = get_drm_device_from_dev_id(dev_id);
+    if (!dev) {
+        va_wayland_error("failed to get DRM device from device ID");
+        return;
+    }
+
+    if (!(dev->available_nodes & (1 << DRM_NODE_RENDER)))
+        goto end;
+
+    dev_path = dev->nodes[DRM_NODE_RENDER];
```
XinfengZhang, line 148: always render node? will not touch drm master node

va/wayland/va_wayland_linux_dmabuf.c
```diff
@@ -0,0 +1,364 @@
...
+    struct zwp_linux_dmabuf_feedback_v1 *feedback;
+
+    if (strcmp(interface, zwp_linux_dmabuf_v1_interface.name) == 0 &&
+        version >= 4) {
+        wl_linux_dmabuf_ctx->has_linux_dmabuf = true;
+        linux_dmabuf =
+            wl_registry_bind(registry, name, &zwp_linux_dmabuf_v1_interface, 4);
+        feedback = zwp_linux_dmabuf_v1_get_default_feedback(linux_dmabuf);
+        zwp_linux_dmabuf_feedback_v1_add_listener(feedback, &feedback_listener, data);
+        zwp_linux_dmabuf_v1_destroy(linux_dmabuf);
```
XinfengZhang, line 227: destroy here? or in handle_global_remove

XinfengZhang (commented):
LGTM besides some questions.

va/wayland/va_wayland_linux_dmabuf.c
emersion: Yeah, DRM master nodes are for KMS only nowadays.

va/wayland/va_wayland_linux_dmabuf.c
emersion: `handle_global_remove` is irrelevant here, the linux-dmabuf global is never removed by the compositor.

We destroy the `zwp_linux_dmabuf_v1` as soon as we don't need it anymore (ie, as soon as we've created a feedback object).

XinfengZhang (approved):
LGTM

emersion:
Thank you!

PR #791, opened 2024-01-10
meson: create dist archives suitable for building with configure, too

eli-schwartz:
Run autogen.sh as a dist script while creating the tarball. Also update autogen.sh to detect when it is being run from `meson dist`, and use that as the srcdir.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -36,7 +36,7 @@ cd "$srcdir"
 
 mkdir -p m4
 
-AUTORECONF=`which autoreconf`
+AUTORECONF=`command -v autoreconf`
 if test -z $AUTORECONF; then
     echo "*** No autoreconf found ***"
     exit 1
--- a/autogen.sh
+++ b/autogen.sh
@@ -23,6 +23,13 @@
 
 PROJECT="libva"
 
+# for `meson dist`
+if test -z "$srcdir"; then
+    srcdir="$MESON_PROJECT_DIST_ROOT"
+    test -n "$srcdir" || srcdir="$MESON_DIST_ROOT"
+    test -n "$srcdir" && NOCONFIGURE=1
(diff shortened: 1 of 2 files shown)


autogen.sh
```diff
@@ -23,6 +23,13 @@
 
 PROJECT="libva"
 
+# for `meson dist`
+if test -z "$srcdir"; then
+    srcdir="$MESON_PROJECT_DIST_ROOT"
+    test -n "$srcdir" || srcdir="$MESON_DIST_ROOT"
+    NOCONFIGURE=1
```
dvrogozh, line 30: As caught by ci, this line breaks typically autotools build flow many people have:
```
$ ./autogen.sh
<...>
$ make
make: *** No targets specified and no makefile found.  Stop.
```

Can we test if we under meson upfront checking existence of some meson specific environment variable and avoid breaking autotools experience? Maybe:
```
if test -n "$MESON_PROJECT_DIST_ROOT"
<...>
fi
```

dvrogozh (changes requested):

autogen.sh
eli-schwartz: We can check if the previous srcdir= assignment succeeded, I guess. ;)

eli-schwartz: I will add that the autotools flow many people have is broken and wrong and utterly incorrect -- people should never ever do this under any circumstances.

This is of course a personal opinion and I didn't intend to break it.

dvrogozh: > We can check if the previous srcdir= assignment succeeded, I guess. ;)

Works for me.

dvrogozh (approved):
looks good. @XinfengZhang : any objections to let this merged?

XinfengZhang (approved):

PR #794, opened 2024-01-31
va: add VASurfaceAttribAlignmentSize

ruijingdong:
Due to different HW implementation, the
surface could require a different alignment size
rather than the default ones, here introduces
a new VASurfaceAttribute, alignment size, which
contains two variables log2_width_alignment
and log2_height_alignment, each has 4 bits,
and the alignment needs to be left shifted
2**size from the application side.

The alignment is in the powers of 2 and
range in [2**0, ... 2**15] = [1, 2, 4, 8, ... 32768]

And this alignment should be met when creating
context as an add-on requirement. If not
implemented, the existing/default alignment
logic will be used.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1699,6 +1699,15 @@ typedef enum {
      * when importing an existing buffer.
      */
     VASurfaceAttribDRMFormatModifiers,
+    /** \brief width and height log2 aligment in pixels (int, read-only)
+     *
+     * For special HW requirement used in some codecs, if
+     * VASurfaceAttribAlignmentSize is not implemented in the driver, then
+     * the surface_width and surface_height should keep the original logic
+     * without any modification, this is an add-on requirement to
+     * surface_width and surface_height.
+     */
+    VASurfaceAttribAlignmentSize,
     /** \brief Number of surface attributes. */
     VASurfaceAttribCount
 } VASurfaceAttribType;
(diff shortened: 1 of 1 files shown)


XinfengZhang:
I am wondering what's the usage of this surface alignments usage: 
currently, when you call vaCreateSurfaces, you will get a surface with default alignment--- this default alignment should already met the alignment request. right? 
in another words, when you query the alignment request, you get a value 16, then when you create surface (480x270), you create it as (480x272). is it the expected behavior. 
but current design, when you create a 480x270, it will give a 480x272 to you by default, what's the difference of these two behaviors?

ruijingdong:
> I am wondering what's the usage of this surface alignments usage: currently, when you call vaCreateSurfaces, you will get a surface with default alignment--- this default alignment should already met the alignment request. right? in another words, when you query the alignment request, you get a value 16, then when you create surface (480x270), you create it as (480x272). is it the expected behavior. but current design, when you create a 480x270, it will give a 480x272 to you by default, what's the difference of these two behaviors?

The original issue has been discussed in  
https://gitlab.freedesktop.org/mesa/mesa/-/issues/9196

This HEVC problem can be resolved by hacking below code in   
vaapi\_encode\_h265.c in FFmpeg.  
  
from

>   
> ctx->surface\_width  = FFALIGN(avctx->width, priv->min\_cb\_size);  
> ctx->surface\_height = FFALIGN(avctx->height, priv->min\_cb\_size);

to

> ctx->surface\_width  = FFALIGN(avctx->width, 64);  
> ctx->surface\_height = FFALIGN(avctx->height, 16);

That is due to the AMD HEVC encoder HW implementation requires   
the image\_input surface dimension to be aligned with 64x16. And the surface\_width   
and surface\_height alignment seems nowhere to get. Considering this could be a   
general issue of different HW implementation constrains, we propose to have such  
an interface so that this alignment information is more flexible to different  
applications and HW implementations.

Thanks,  
Ruijing

XinfengZhang (commented):
the change itself looks good to me, but do not touch the micro_version, thanks
it will be a part of 2.21.0, not 2.21.1

meson.build
```diff
@@ -20,7 +20,7 @@ project(
 # - reset minor version to zero when major version is incremented
 va_api_major_version = 1
 va_api_minor_version = 21
-va_api_micro_version = 0
+va_api_micro_version = 1
```
XinfengZhang, line 23: please dont change it, 
our current release process is:
master: x.xx.0.pre1 
then release as x.xx.0
then bump master to x."xx+1".0.pre

if there are hotfix in release branch , we will bump micro version

XinfengZhang (changes requested):

meson.build
ruijingdong: Thanks for the explanation, updated without the version change.

XinfengZhang (approved):

ruijingdong:
Thanks for reviewing, anything I need to do from my side?

PR #796, opened 2024-02-07
va: Add encoding tuning mode

nowrep:
This provides a hint to driver about the requested encode tuning in addition to quality level number. It can be used by hardware encoders that supports low latency encoding as a special mode instead of setting the lowest quality level.

The tuning modes matches Vulkan VkVideoEncodeTuningModeKHR, except lossless which I don't think makes sense here.

Motivation is AMD VCN encoder which has low latency mode that should be enabled for usecases like realtime game streaming, and it works with all presets (quality levels), not only with the fastest preset.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1050,6 +1050,15 @@ typedef enum {
      * columns supported for encoding with tile support.
      */
     VAConfigAttribEncMaxTileCols        = 57,
+    /**
+     * \brief Encoding tuning mode attribute. Read-only.
+     *
+     * This attribute conveys whether the driver supports different tuning modes
+     * for encoding. The attribute value is any combination of VA_ENC_TUNING_MODE_XXX.
+     *
+     * The tuning mode is passed to the driver using VAEncMiscParameterBufferQualityLevel.
+     */
+    VAConfigAttribEncTuningMode         = 58,
     /**@}*/
     VAConfigAttribTypeMax
 } VAConfigAttribType;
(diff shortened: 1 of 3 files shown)


DemiMarie:
> This provides a hint to driver about the requested encode tuning in addition to quality level number. It can be used by hardware encoders that supports low latency encoding as a special mode instead of setting the lowest quality level.
> 
> The tuning modes matches Vulkan VkVideoEncodeTuningModeKHR, except lossless which I don't think makes sense here.

So there are two specific cases I can think of:

1. The encoding is happening in real-time, so the hardware must guarantee that it can keep up with the input stream.
2. Lossless encoding is needed, perhaps because the file will later be decompressed and re-encoded offline to a quality level that would not be possible with real-time lossy encoding.

XinfengZhang:
when we are talking about low latency, there are several perspectives, such as:
mini gop structure , whether some frame need to be hold
time consumption for one frame.  and it is actually related with frequency logic. 

looks it is not fully related with encoding tools

va/va.h
```diff
@@ -2707,6 +2716,15 @@ typedef struct _VAEncMiscParameterBufferMultiPassFrameSize {
     unsigned long           va_reserved[VA_PADDING_LOW];
 } VAEncMiscParameterBufferMultiPassFrameSize;
 
+/** \brief Default tuning mode */
+#define VA_ENC_TUNING_MODE_DEFAULT                      0x00000000
+/** \brief Compromise latency of video encode operations to improve quality  */
+#define VA_ENC_TUNING_MODE_HIGH_QUALITY                 0x00000001
```
XinfengZhang, line 2722: duplicate with TU1?

nowrep: This would likely be equal to `quality = 1`, yes. It can be removed if you prefer.

nowrep:
> time consumption for one frame. and it is actually related with frequency logic.

Yes, latency here meaning the time it takes to encode one frame.

XinfengZhang:
> > time consumption for one frame. and it is actually related with frequency logic.
> 
> Yes, latency here meaning the time it takes to encode one frame.

XinfengZhang:
> > 
> > Yes, latency here meaning the time it takes to encode one frame.

it means that backend driver need communication with Pcode...

va/va.h
XinfengZhang: yes, from my perspective, I suppose it should be removed, if we need define this tuning mode, it means it should be orthogonal with quality level.

XinfengZhang:
there are 3 quality/perf related control in vulkan, "usage" "content" "tuning" , I am wondering how to map them with quality level and tuning ...
seems it is series new interface.

va/va.h
nowrep: Removed

nowrep:
Replaced with https://github.com/intel/libva/pull/841

PR #799, opened 2024-02-12
va: Re-add drm_state and fd checks to VA_DRM_GetDriverNames

sivileri:
Re-adding `drm_state` null and checks to `VA_DRM_GetDriverNames`, apps like gstreamer will crash on WSL without this check, which existed before in VA_DRM_GetDriverName before but was removed in commit 
[f09781156b24c571c187250f6d0f5eeff042528b](https://github.com/intel/libva/commit/f09781156b24c571c187250f6d0f5eeff042528b).

After re-adding check gstreamer works fine again and checked vainfo/ffmpeg also work fine.

@dvrogozh @XinfengZhang could you please help review/merge ?


sivileri:
@XinfengZhang could you please merge this PR ? (Don't have permissions myself), thanks! :)

PR #800, opened 2024-02-18
Update Prime3 definition

lixin5x:
0x80000000 usually used as unsupport bit, update to new value.

Signed-off-by: Li, Xin6 <xin6.li@intel.com>


XinfengZhang (approved):
this PR will break compatibility, but I suppose no one is using PRIME3 now, so LGTM . thanks

PR #810, opened 2024-05-11
va:add synchornization fence for HW execution

XinfengZhang:
do not merge!
add fence in and fence out for vaEndPicture which is using to submit command buffer fence out will be signaled after HW execution, fence in is the dependencies

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -4175,6 +4175,24 @@ VAStatus vaEndPicture(
 );
 
 /**
+ * Same behavior as vaEndPicture except a sync fd list for synchronizations.
+ * if sync_num = 0 or sync_fds == NULL, the behavior should be same as vaEndPicture.
+ * if sync_num >= 1, the sync_fds[0] should be the fence out, this fd(or the fence
+ * behind this fd)  will be overridden by this call, and will be signaled to indicate
+ * current frame finishing. sync_num = 1 means only fence out is needed.
+ * sync_fds[1] to sync_fds[1 ~ sync_num-1] is fence in, current HW execution
+ * will be blocked until all these fences are signaled.
+ * these fence fds is file descriptor of dma_fence.
+ */
+
+VAStatus vaEndPicture2(
+    VADisplay dpy,
+    VAContextID context,
+    int32_t *sync_fds,
+    int32_t sync_num
+);
+
+/**
  * Make the end of rendering for a pictures in contexts passed with submission.
  * The server should start processing all pending operations for contexts.
  * All contexts passed should be associated through vaMFAddContext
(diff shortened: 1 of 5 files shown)


ceyusa:
is this Linux only?

XinfengZhang:
> is this Linux only?

good question, it is for a linux PoC , will not merge until we could verify it. 
I suppose we could refer vulkan vkSemaphore or vkFence in future to support all OS in future. 
such as: add a interface to createsyncobj and destroysyncobj, the syncobj could be used to encapsulate different OS sync object.

va/va.h
```diff
@@ -4174,6 +4174,24 @@ VAStatus vaEndPicture(
     VAContextID context
 );
 
+/**
+ * Same behavior with vaEndPicture except a sync fd list for synchronizations.
+ * if sync_num = 0 or sync_fds == NULL, the behavior should be same with vaEndPicture.
```
yellowdoge, line 4179: Should be: "Same behaviour as" and ... "the behaviour should be the same as vaEndPicture()"

va/va.h
```diff
@@ -4174,6 +4174,24 @@ VAStatus vaEndPicture(
...
+ * if sync_num >= 1, the sync_fds[0] should be the fence out, this fd(or the fence
+ * behind this fd)  will be overridden by this call, and will be signaled to indicate
+ * current frame finishing.
+ * sync_fds[1] to sync_fds[1 ~ sync_num-1] is fence in, current HW execution
+ * will be blocked until all these fences are signaled.
+ * these fence fds is file descriptor of dma_fence.
```
yellowdoge, line 4185: Is it possible to call vaEndPicture2() with a fence-out but no fence-in(s)? If so it'd be good to specify it in this function comment (and maybe other expected example use patterns);

va/va.h
```diff
@@ -4174,6 +4174,24 @@ VAStatus vaEndPicture(
...
+ */
+
+VAStatus vaEndPicture2(
+    VADisplay dpy,
+    VAContextID context,
+    int32_t *sync_fds,
```
yellowdoge, line 4191: Is there any advantage in bundling fences out and in vs e.g.

```
   int32_t *sync_fd_out,     // Out. Can be nullptr.
   int32_t* sync_fds_in,     // In. Can be nullptr.
   int32_t sync_fds_in_num,  // In. Can be 0.
```

Or something like that?

XinfengZhang: both ok to me, but now , it is just a PoC, and the implementation is ready here:
https://github.com/intel/media-driver/pull/1803

in the future:
1. libva should support both linux and windows, so, we will have some other interfaces to hide the dma fence, unify the syn definition in future.
2. decouple the fence in and out to too parameters as your mentioned.

PR #811, opened 2024-05-13
va: remove older 0.32 and 0.33 support

XinfengZhang:
libva version 2.x should be compatible with 2.x series , not 1.x series and 0.x series


XinfengZhang:
could be used to fix the issue mentioned in #758

PR #812, opened 2024-05-15
va: remove older 0.32 and 0.33 support

XinfengZhang:
libva version 2.x should be compatible with 2.x series , not 1.x series and 0.x series
copy #811, but reserve va_compat.h , because maybe some existing usage refer this file


xhaihao (approved):
LGTM

PR #817, opened 2024-05-23
Add flag to swap references.

FurongZhang:
There are different understanding of forward references(could be past frames or future frames) and backward references (could be past frames or future frames).

This flag allows the user to configure the reference as past or future frames.
     - Value 0: by default, forward_references mean future frames, backward_references mean past frames
     - Value 1: swap, forward_references mean past frames, backward_references mean future frames

### DIFF
--- a/va/va_vpp.h
+++ b/va/va_vpp.h
@@ -989,11 +989,11 @@ typedef struct _VAProcPipelineParameterBuffer {
     VABufferID         *filters;
     /** \brief Actual number of filters. */
     uint32_t           num_filters;
-    /** \brief Array of forward reference frames (past frames). */
+    /** \brief Array of forward reference frames (future frames by default). */
     VASurfaceID        *forward_references;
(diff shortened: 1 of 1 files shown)


va/va_vpp.h
```diff
@@ -989,11 +989,11 @@ typedef struct _VAProcPipelineParameterBuffer {
     VABufferID         *filters;
     /** \brief Actual number of filters. */
     uint32_t           num_filters;
-    /** \brief Array of forward reference frames (past frames). */
+    /** \brief Array of forward reference frames (future frames by default). */
```
XinfengZhang, line 992: non-backward change

XinfengZhang (changes requested):

ceyusa:
would be possible to the driver to ignore the swap request?

FurongZhang:
> would be possible to the driver to ignore the swap request?

If the application does not need swap, the application can ignore the swap.

PR #820, opened 2024-06-04
va: Add VA_SURFACE_ATTRIB_USAGE_HINT_SCANOUT

vcheah:
This can be used to tell the driver set scanout surface. XE DMA DRM render case required this.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1813,6 +1813,10 @@ typedef struct _VASurfaceAttribExternalBuffers {
 /** \brief Surface used for export to third-party APIs, e.g. via
  *  vaExportSurfaceHandle(). */
 #define VA_SURFACE_ATTRIB_USAGE_HINT_EXPORT     0x00000020
+/** \brief Surface read by decoder or video post-processing.
+ * use to set DRM_XE_GEM_CREATE_FLAG_SCANOUT which is required
+ * for XE DMA DRM render case */
+#define VA_SURFACE_ATTRIB_USAGE_HINT_SCANOUT   0x00000040
 
 /**@}*/
 


XinfengZhang:
1. please add signoff
2. when the scanout is set, the surface is displayable ? how about "if not"? ,

vcheah:
> 1. please add signoff
> 2. when the scanout is set, the surface is displayable ? how about "if not"? ,

The scanout surface only required for XE DRM render case. I have tested this on MTL on XE. https://github.com/vcheah/libraries.media.vpl.tools/tree/scanout

If scanout not set, drmModeAddFB2WithModifiers() will fail (-EINVA). The kernel message error points to https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/xe/display/intel_fb_bo.c#L23

XinfengZhang:
> If scanout not set, drmModeAddFB2WithModifiers() will fail (-EINVAL). The kernel message error points to https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/xe/display/intel_fb_bo.c#L23

the problem is: if not set ,will cause issue, then this PR break the compatibility .

vcheah:
> > If scanout not set, drmModeAddFB2WithModifiers() will fail (-EINVAL). The kernel message error points to https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/xe/display/intel_fb_bo.c#L23
> 
> the problem is: if not set ,will cause issue, then this PR break the compatibility .

Do you have any other good recommendation in mind?  The problem is when the created surfaces and had undergo VM_BIND operation, The list_empty(&bo->ttm.base.gpuva.list) will never be empty, hence, it will always encounters the failing case.

vcheah:
> > If scanout not set, drmModeAddFB2WithModifiers() will fail (-EINVAL). The kernel message error points to https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/xe/display/intel_fb_bo.c#L23
> 
> the problem is: if not set ,will cause issue, then this PR break the compatibility. 

The flags is actually only required for DRM render case in order for render to work. 

For XE native weston10 case, I have checked (MTL/RPL), sometime ago, having the scanout surface set or NOT, the render output is correct. However, weston10 + scanout surface, it seems to suffer some performance degradation based on some early data that I've collected sometime ago.

vcheah:
Closing the MR.  Track this:  https://github.com/intel/media-driver/pull/1822

PR #821, opened 2024-06-06
meson:remove autogen.sh from the meson script

### LINKED ISSUE #804: Does not build: Program './autogen.sh' not found or not executable
When I try building the latest release 2.21.0, I get an error while configuring:

```
meson setup . debian/obj-amd64 -Dbuildtype=release -Dprefix=/usr -Dlibdir=/usr/lib64
The Meson build system
Version: 1.4.0
Source dir: /home/unfrobbed-dev/pkgs/libva/libva-2.21.0
Build dir: /home/unfrobbed-dev/pkgs/libva/libva-2.21.0/debian/obj-amd64
Build type: native build
Project name: libva
Project version: 2.21.0
C compiler for the host machine: cc (gcc 13.2.0 "cc (GCC) 13.2.0")
C linker for the host machine: cc ld.bfd 2.42
Host machine cpu family: x86_64
Host machine cpu: x86_64
Library dl found: YES
Found pkg-config: YES (/bin/pkg-config) 0.29.2
Run-time dependency libdrm found: YES 2.4.120
Run-time dependency x11 found: YES 1.8.8
Run-time dependency xext found: YES 1.3.6
Run-time dependency xfixes found: YES 6.0.1
Run-time dependency x11-xcb found: YES 1.8.8
Run-time dependency xcb found: YES 1.16.1
Run-time dependency xcb-dri3 found: YES 1.16.1
Run-time dependency gl found: YES 1.2
Run-time dependency wayland-client found: YES 1.22.0
Build-time dependency wayland-scanner found: YES 1.22.0
Program /usr/bin/wayland-scanner found: YES (/usr/bin/wayland-scanner)
Checking for function "secure_getenv" : YES 
Configuring va_version.h using configuration
Checking if "-Wl,--version-script" : links: YES 
Program doxygen found: YES (/bin/doxygen)

meson.build:161:6: ERROR: Program './autogen.sh' not found or not executable

A full log can be found at /home/unfrobbed-dev/pkgs/libva/libva-2.21.0/debian/obj-amd64/meson-logs/meson-log.txt
```

XinfengZhang:
suppose no need to use customized package options. Fixes #804

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -158,4 +158,3 @@ if get_option('enable_docs') and doxygen.found()
   subdir('doc')
 endif
 
-meson.add_dist_script('./autogen.sh')


XinfengZhang:
ping @eli-schwartz , please help to take a look.

xhaihao:
LGTM.

xhaihao (approved):

eli-schwartz:
The purpose of that line was so that meson could be used to create dist tarballs (that contain a ./configure script). Removing this means that you cannot do `meson dist` and upload the resulting tarball as a universal release asset.

That may or may not be desirable. An alternative would be to include it via EXTRA_DIST.

XinfengZhang:
> The purpose of that line was so that meson could be used to create dist tarballs (that contain a ./configure script). Removing this means that you cannot do `meson dist` and upload the resulting tarball as a universal release asset.
> 
thanks, so your initial purpose is adding the missing files into the tarball generated by meson dist, and these file are not using meson , but auto config.  so it is not a universal release asset if there are no such files in the tarball. 
but from another perspective. people should not expect to call autogen.sh if there are only meson.


> That may or may not be desirable. An alternative would be to include it via EXTRA_DIST.

will you contribute a patch?   suppose new patch will add missing files, but will not ask auto config support when run meson...

XinfengZhang:
> > will you contribute a patch? suppose new patch will add missing files, but will not ask auto config support when run meson...
> 
> Sure.
> 
> Note that meson only wants auto config support when running `meson dist`, it is simply that the helper script run during dist should always exist.

thanks,

quick summary:
 the origin issue #804 is caused by two problem:
1. meson need autogen.sh when run "meson setup builddir", we are expecting it only be needed only when "meson dist"
2. the autogen.sh is not existing in the tarball, because there are no such file in the tarbar generated by  "make dist-bzip2"

PR #832, opened 2024-08-12
va: limit to VAProfileAV1Profile2_420

davidwuAMD:
Some hardwares do not support a full set of AV1 profile2, in this case use a subset instead. At the moment, VAProfileAV1Profile2_420 is for supporting 12-bit 4:2:0.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -542,7 +542,7 @@ typedef enum {
     VAProfileH264High10                 = 36,
     VAProfileVVCMain10                  = 37,
     VAProfileVVCMultilayerMain10        = 38,
-    VAProfileAV1Profile2                = 39
+    VAProfileAV1Profile2_420            = 39
 } VAProfile;
 
 /**
--- a/va/va_str.c
+++ b/va/va_str.c
@@ -64,7 +64,7 @@ const char *vaProfileStr(VAProfile profile)
         TOSTR(VAProfileHEVCSccMain444);
         TOSTR(VAProfileAV1Profile0);
         TOSTR(VAProfileAV1Profile1);
-        TOSTR(VAProfileAV1Profile2);
+        TOSTR(VAProfileAV1Profile2_420);
         TOSTR(VAProfileHEVCSccMain444_10);
         TOSTR(VAProfileProtected);
         TOSTR(VAProfileVVCMain10);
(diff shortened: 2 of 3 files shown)


davidwuAMD:
this is to address the issue in https://github.com/intel/libva/pull/828

davidwuAMD:
@XinfengZhang, any concern about this patch? please let me know I can take a look.

davidwuAMD:
wondering if this change can be merged? Am I missing something?

XinfengZhang:
I am ok to merge it , but still not understand why profile + vaSurfaceAttrib could not report the restriction, such as 12bit 422only could use profile2+P010 format.  of course, it will be more complex than current solution.

XinfengZhang (approved):

davidwuAMD:
OK - I will see if I can find something with profile + vaSurfaceAttrib. I think there could be a reporting issue (vainfo) as it cannot show 420only if we did not define a profile for it. (but I need to look into more details to be sure). 
When will be the next libva release? I'd like to see if we need to apply this merge or not before that.

XinfengZhang:
suppose next release will happened by end of this year.
I am ok to merge current change, because it is simpler than profile+vaSurfaceAttrib. application just need to query the profile list.  of course, from other hand, if backend driver want to support both. they have to report both profile. 

I am thinking whether there are similar problem with other codec,   and whether we need to use "profile + surfaceAttrib" to  keep behavior consistent.

just like AV1 profile0, how to clarify driver support monochrome or not?  ho to differentiate driver support 444 8bit or 444 10bit for AV1 profile1? and suppose there are also some problem with VP9. 

so, both solution is ok to me, feel free to ping me to merge current PR.

davidwuAMD:
Hi @XinfengZhang - either VAProfileAV1Profile2_420 or VAProfileAV1Profile2 should be OK for mesa or libva. The real issue is in ffmpeg which uses the best match profile/format to decode video. Once ffmpeg enables full set profile2 support then ffmpeg will use p012 to decode a YUV444 video clip even though it is not supported in mesa (only 12bit YUV420 is supported).

davidwuAMD:
let me drop this PR as there is a solution in ffmpeg to check the RT format so it won't use an unexpected format to decode

PR #835, opened 2024-09-25
Add Android.bp for libva 2.22

JeevakaPrabu:
Changes include:
- Replace Android.mk with Android.bp for libva 2.22
- Add pre-generated va/va_version.h and remove it from .gitignore


XinfengZhang (commented):
va_version.h is generated under meson and make, why change to pre-generated under android?

JeevakaPrabu:
> va_version.h is generated under meson and make, why change to pre-generated under android?

Tried to generate the va/va_version.h from Android.bp but met with some road blocks. So for the time being added the pre-generated va_version.h

XinfengZhang:
I still hope we could generate va/va_version.h from android.bp,  then the behavior will be aligned between linux/win/android. @rosetta-jpn any comments

XinfengZhang:
@xuguangxin

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+license {
+    name: "external_intel_libva_license",
+    visibility: [":__subpackages__"],
+    license_text: [
+        "COPYING",
+    ],
+}
+
+sh_binary_host {
+    name: "libva_version_gen_script",
```
andrescj-chromium, line 34: nit: `libva_gen_version_script`

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+    src: "build/gen_version.sh",
+}
+
+genrule {
+    name: "libva_version_gen",
```
andrescj-chromium, line 39: nit: `libva_gen_version`

theandi666 (approved):
LGTM

Android.bp
JeevakaPrabu: > nit: `libva_gen_version_script`

Change done

Android.bp
JeevakaPrabu: > nit: `libva_gen_version`

Change done

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+        "COPYING",
+    ],
+}
+
+sh_binary_host {
+    name: "libva_gen_version_script",
+    src: "build/gen_version.sh",
+}
+
+genrule {
```
andrescj-chromium, line 38: REVISIT.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+    out: ["va/va_version.h"],
+    cmd: "$(location libva_gen_version_script) --configure_ac $(in) > $(out)",
+}
+
+cc_library_headers {
+    name: "libva_headers",
+
+    export_include_dirs: [
+        ".",
+        "va",
```
andrescj-chromium, line 56: Is this actually needed?

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+    name: "libva_gen_version",
+    srcs: [
+        "configure.ac",
+        "va/va_version.h.in",
+    ],
+    tools: [
+        "libva_gen_version_script",
+    ],
+    out: ["va/va_version.h"],
+    cmd: "$(location libva_gen_version_script) --configure_ac $(in) > $(out)",
```
andrescj-chromium, line 48: Up to you, but consider changing this `cmd` to be as follows in order to avoid having to make the changes to `gen_version.sh` (that way, you don't have to maintain a special path from `--configure_ac`):

```
    cmd: "$(location libva_gen_version_script) " +
        "$$(dirname $(location configure.ac)) " +
        "$(location va/va_version.h.in) > $(out)",
```

Android.bp
andrescj-chromium, line 56: Are `"va"` and `"va/drm"` actually needed? Users of libva tend to include headers using the `va/` prefix, such as `"va/va.h"` instead of `"va.h"`.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+
+    shared_libs: [
+        "libdl",
+        "libdrm",
+        "libcutils",
+        "liblog",
+    ],
+
+    local_include_dirs: [
+        ".",
```
andrescj-chromium, line 87: The `"."` include doesn't seem necessary. I tested locally and libva still builds without it.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+
+    generated_headers: [
+        "libva_gen_version",
+    ],
+    export_generated_headers: [
+        "libva_gen_version",
+    ],
+
+    header_libs: [
+        "libutils_headers",
```
andrescj-chromium, line 99: Is `"libutils_headers"` needed? If it's here because of [this](https://github.com/intel/libva/blob/01153523f25ac703e7ce2afc0023788a90f6c1bb/va/va.c#L51-L53), maybe it's more appropriate to use `"liblog_headers"`

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+    srcs: [
+        "va/va.c",
+        "va/va_trace.c",
+        "va/va_str.c",
+        "va/drm/va_drm.c",
+        "va/drm/va_drm_auth.c",
+        "va/drm/va_drm_utils.c",
+    ],
+
+    cflags: [
```
andrescj-chromium, line 111: Consider adding `-Wall` here.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+        "-Werror",
+        "-Wno-error",
```
andrescj-chromium, line 113: Wouldn't this `"-Wno-error"` cancel out the effect of `"-Werror"`?

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+        "-DSYSCONFDIR=\"/vendor/etc\"",
+        "-DLOG_TAG=\"libva\"",
+    ],
+
+    arch: {
+        x86_64: {
+            cflags: ["-DVA_DRIVERS_PATH=\"/vendor/lib64\""],
```
andrescj-chromium, line 120: This should probably be as follows to avoid changing behavior from Android.mk:

```
"-DVA_DRIVERS_PATH=\"/vendor/lib64/dri\""
```

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+            enabled: true,
+        },
+    },
+
+    vendor: true,
+    enabled: false,
+}
```
andrescj-chromium, line 127: Do you also want to add `proprietary: true` to match the [Android.mk](https://github.com/intel/libva/blob/01153523f25ac703e7ce2afc0023788a90f6c1bb/va/Android.mk#L61C1-L61C25)? In fact, to match Android.mk, maybe every declaration in this Android.bp (except for `package` and `license`) should have `proprietary: true`.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+
+cc_library_shared {
+    name: "libva-android",
+
+    static_libs: [
```
andrescj-chromium, line 132: It looks like you can remove this entire `static_libs`.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+        "libdrm",
+        "liblog",
+    ],
+
+    local_include_dirs: [
+        "va",
+        "va/drm",
+    ],
+
+    header_libs: [
```
andrescj-chromium, line 147: It looks like you can remove this `header_libs`.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+
+    srcs: [
+        "va/android/va_android.cpp",
+        "va/drm/va_drm_utils.c",
+    ],
+
+    cflags: [
+        "-Wall",
+        "-Werror",
+        "-Wno-error",
```
andrescj-chromium, line 160: This seems like it would cancel out the effect to `-Werror`.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+    ],
```
andrescj-chromium, line 161: You may also want to add `"-DLOG_TAG=\"libva-android\""` to match the [Android.mk](https://github.com/intel/libva/blob/01153523f25ac703e7ce2afc0023788a90f6c1bb/va/Android.mk#L90).

andrescj-chromium (changes requested):
Looks essentially good to me. Thanks Jeevaka!

Android.bp
JeevakaPrabu: Driver path is /vendor/lib64/. No need to change to /vendor/lib64/dri

Android.bp
JeevakaPrabu: As there are few components using without va/ prefix, for now will leave this untouched. Once the other components(onevpl) move to va/ usage, will provide another PR.

Android.bp
andrescj-chromium: Ack. Did you also mean to use `vendor: true` for the `sh_binary_host` and the `genrule`? I guess those things are host-side, so probably not applicable, but just making sure.

andrescj-chromium (approved):
Approved from my end % changes.

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+        "va",
+        "va/drm",
+    ],
+
+    srcs: [
+        "va/android/va_android.cpp",
+        "va/drm/va_drm_utils.c",
+    ],
+
+    cflags: [
```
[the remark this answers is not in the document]
JeevakaPrabu: Done

Android.bp
```diff
@@ -0,0 +1,170 @@
...
+    srcs: [
+        "va/va.c",
+        "va/va_trace.c",
+        "va/va_str.c",
+        "va/drm/va_drm.c",
+        "va/drm/va_drm_auth.c",
+        "va/drm/va_drm_utils.c",
+    ],
+
+    cflags: [
```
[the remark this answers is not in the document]
JeevakaPrabu: Recommended change done

Android.bp
XinfengZhang: linux  driver location is /lib/x86_64-linux-gnu/dri/
so , it is different, right?

XinfengZhang (approved):

PR #836, opened 2024-09-25
va: Add VA_PICTURE_H264_NON_EXISTING flag

nowrep:
This flag indicates that the picture is marked
as “non-existing” (spec section 8.2.5.2)

Mesa impl: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/31368

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -3584,6 +3584,11 @@ typedef struct _VAPictureH264 {
 #define VA_PICTURE_H264_BOTTOM_FIELD        0x00000004
 #define VA_PICTURE_H264_SHORT_TERM_REFERENCE    0x00000008
 #define VA_PICTURE_H264_LONG_TERM_REFERENCE 0x00000010
+/** \brief Non existing picture
+ * This flag indicates that the picture is marked
+ * as “non-existing” (spec section 8.2.5.2)
+ */
+#define VA_PICTURE_H264_NON_EXISTING        0x00000020
 
 /** H.264 Picture Parameter Buffer */
 /*


va/va.h
```diff
@@ -3584,6 +3584,11 @@ typedef struct _VAPictureH264 {
 #define VA_PICTURE_H264_BOTTOM_FIELD        0x00000004
 #define VA_PICTURE_H264_SHORT_TERM_REFERENCE    0x00000008
 #define VA_PICTURE_H264_LONG_TERM_REFERENCE 0x00000010
+/** \brief Non existing picture
+ * This flag indicates that the picture is marked
+ * as “non-existing” (spec section 8.2.5.2)
+ */
+#define VA_PICTURE_H264_NON_EXISTING        0x00000020
```
XinfengZhang, line 3591: dose it mean , you still provide the reference picture ID , but mark it as non-existing?
from my understanding, if there are no reference picture ID, it will be non-existing

nowrep: Yes, picture ID is still provided but marked as non-existing. This is what apps currently do, they provide picture ID but the driver has no way knowing it should be non-existing.

It should have the same meaning as Vulkan `is_non_existing`: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkVideoDecodeH264DpbSlotInfoKHR.html#_description

XinfengZhang: thanks. 
two questions:
1. whether it also should be applied to other codec, such as VA_PICTURE_HEVC_XXX
2. if application already know it is non_existing, why still provide valid picture_id?

nowrep: 1. There is a similar flag for VVC `VA_PICTURE_VVC_UNAVAILABLE_REFERENCE`.  Not sure about other codecs (HEVC), currently AMD only supports this for H264 so I wouldn't be able to verify it works as expected.
2. Applications (ffmpeg/gstreamer) currently provide valid picture_id in this case, so I think this behavior should not change to not break compatibility with drivers not supporting or ignoring this new flag. So the valid picture_id should be provided and additionally the `VA_PICTURE_H264_NON_EXISTING` flag be set.

nowrep:
Looks like Intel also supports this flag for H264, implemented in Mesa https://gitlab.freedesktop.org/mesa/mesa/-/blob/992ada2f8fb9c5dbafe7f119f4a680df92b144aa/src/intel/vulkan/genX_cmd_video.c?page=2#L1033 but seems not used in intel-media-driver. 

So it should be useful also for Intel.

PR #843, opened 2024-12-12
va_dec_av1: Add max_frame_width/height_minus_1

nowrep:
AMD VCN needs this to correctly decode av1-1-b8-22-svc-L2T1 and av1-1-b8-22-svc-L2T2.


XinfengZhang (commented):
not understand , this structure is frame level, each frame already has current width and height.  from my understanding , max_frame_width_minus_1 max_frame_height_minus_1. should in seq level to determine the maximum resolution. why we need it for one frame ?

PR #847, opened 2025-03-10
va: Add encoding QP map parameters

ruijingdong:
QP map is a way of controlling block  
level QP(QI). Here it introduces two  
modes, delta and absolute. Delta mode  
will add/minus extra QP(QI) values  
based on CQP/RC's QP result, while  
absolute mode is directly using the QP  
(QI) values from QP map whenever possible;  
in absolute mode, since RC's adjustment  
on QP values not working, it is only  
valid in CQP mode, and RC adjustment  
in fact is processed externally.

Adding this is to support some encoding  
HWs, and provides a way of QP adjustment  
in addition to HW internal RC modes.

This defines VAEncMiscParameterQPMap  
structure and its corresponding attribute  
VAConfigAttribValEncQPMap, also defines  
the format of QP map element and buffer  
specification.

In contrast with FEI, QP map is an  
independent parameter set, and can be  
applied to the needed profiles, only  
with extra processing of the proposed  
misc parameters.

### DIFF
--- a/va/va.h
+++ b/va/va.h
@@ -1062,6 +1062,17 @@ typedef enum {
      * VAConfigAttribValEncVP9 union.
      */
     VAConfigAttribEncVP9                = 58,
+    /**
+     * \brief QP map encoding attribute. Read-only.
+     *
+     * This attribute describes whether encoder supports QP map (delta QP or absolute QP),
+     * and the specification of the supported QP map buffer, the block size each element
+     * covered as well as the number of bytes used by each element, the assumption
+     * here is each MB (h264), CTB (h265), or SB (AV1) is sharing the same QP/QI, and the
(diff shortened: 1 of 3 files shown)


ruijingdong:
Hi @XinfengZhang, 

I proposed a structure of QP map adjustment based on current AMD VCN implementation to VAAPI interface, this just provides an additional way to rate control, and to let external driver or application be able to handle rate control partially or totally. It is sharing the same logic as ROI in AMD driver stack, just QP map is more flexible, and it operates on the whole frame level.

Also, I have made some draft code changes below for testing.

Libva-utility test draft code
https://github.com/ruijingdong/libva-utils/commits/qp_map_test
 
Mesa draft code
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33778/commits

Let me know your comments.
Thanks,
Ruijing

PR #849, opened 2025-04-08
Fix subproject dependency version

xclaesse:
It must match the pkg-config version which uses va_api_version.

### DIFF
--- a/va/meson.build
+++ b/va/meson.build
@@ -86,7 +86,8 @@ libva = shared_library(
 libva_dep = declare_dependency(
   link_with : libva,
   include_directories : configinc,
-  dependencies : [ dl_dep ])
+  dependencies : [ dl_dep ],
+  version : va_api_version)
 
 if WITH_DRM
   libva_drm_sources = [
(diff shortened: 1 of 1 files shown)


xclaesse:
For context see https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/8793#note_2857006.

ceyusa:
Let me explain why this change is required:

On one side, libva handles two versions: one for the library and one for the API. The current release, for example, the library is 2.22, while the API is 1.22.

When reporting the library version through pkg-config, libva reports back API version.

It's strange, but it has been as is since ever.  No issues.

Nevertheless, meson has the [subproject](https://mesonbuild.com/Subprojects.html) feature as a way to "bundle third party libraries in your source tree". And GStreamer uses it for certain dependencies, as it does for libva.

The problem lies here: by default meson reports back, when a third party library is built as a subproject, the **library version** (2.22), when in this case we must report back the **API version** (1.22) in order to keep the same checks when using the library via pkg-config.

And that is what this patch fixes.

So far, this issue has passed unaware because most of the people uses system installed libraries, where the pkg-config is used. Just recently was found out because of H.266 decoder.

xclaesse:
CC @XinfengZhang

PR #850, opened 2025-04-09
Add VAProfileH264High422

vcheah:
Add VAProfileH264High422 to support 10bit 422 AVC Decoding


XinfengZhang (changes requested):
this change break ABI, please ensure the value keep unchange, new added one should use new value.

vcheah:
> this change break ABI, please ensure the value keep unchange, new added one should use new value.
I have updated accordingly.

XinfengZhang (approved):

PR #853, opened 2025-04-29
Add HEVC RefPicSetLtCurr and NumPocLtCurr picture parameters

nowrep:
Corresponds to the variables of same name as defined in section 8.3.2 of the HEVC spec.

The RefPicSetStCurrBefore and RefPicSetStCurrAfter can be derived by sorting them, however there is no defined order for the RefPicSetLtCurr and thus it must be provided explicitly.

This is needed for AMD driver and also to match the DXVA and Vulkan Video interfaces.

### DIFF
--- a/va/va_dec_hevc.h
+++ b/va/va_dec_hevc.h
@@ -185,8 +185,15 @@ typedef struct  _VAPictureParameterBufferHEVC {
      */
     uint32_t                st_rps_bits;
 
+    /* Corresponds to the variables of same name as defined in section 8.3.2
+     * of the HEVC spec. Index into ReferenceFrames[]. */
+    uint8_t                 RefPicSetLtCurr[8];
+    uint8_t                 NumPocLtCurr;
+
+    uint8_t                 reserved[3];
+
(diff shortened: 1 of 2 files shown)


Jexu (commented):
RefPicSetStCurrBefore, RefPicSetStCurrAfter and RefPicSetLtCurr could be all derived by sorting ReferenceFrames. And you can refer to it in media driver.

https://github.com/intel/media-driver/blob/master/media_softlet/linux/common/codec/ddi/dec/ddi_decode_hevc_specific.cpp#L402

nowrep:
That's not enough, because this way the lists will be in the order as they appear in DPB. But the order here does matter, so this is wrong.

The correct order of RefPicSetStCurrBefore and RefPicSetStCurrAfter can be derived by sorting them by POC (desc/asc). 

But this is not possible for RefPicSetLtCurr as the order is as it appear in the slice header (and the spec doesn't require them to be sorted as is the case with short term refs). Since there is no way to derive this, the RefPicSetLtCurr must be provided by application.

PR #865, opened 2025-09-24
trace:add vpp output surface dump support

XinfengZhang:
use LIBVA_TRACE_SURFACE=vppout to dump vpp output surface


xhaihao (approved):
LGTM

PR #872, opened 2026-02-06
Add ProtectedSession Related Log in Trace

lixin5x:
Dump structure data of protected session events.

### DIFF
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -7543,6 +7543,8 @@ void va_TraceProtectedSessionExecute(
     VABufferType type;
     unsigned int size;
     unsigned int num_elements;
+    unsigned int i;
+    unsigned char *p;
     bool buf_valid = false;
     VAProtectedSessionExecuteBuffer *execute_buf = NULL;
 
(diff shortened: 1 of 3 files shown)


va/va_trace.c
```diff
@@ -4612,6 +4612,85 @@ static void va_TraceVAEncPackedHeaderParameterBufferType(
...
+            va_TraceMsg(trace_ctx, "\t\t\tsegment_length = %u\n",
+                        p->segment_info[i].segment_length);
+            va_TraceMsg(trace_ctx, "\t\t\tpartial_aes_block_size = %u\n",
+                        p->segment_info[i].partial_aes_block_size);
+            va_TraceMsg(trace_ctx, "\t\t\tinit_byte_length = %u\n",
+                        p->segment_info[i].init_byte_length);
+
+            /* Print first 16 bytes of IV/CTR */
+            va_TraceMsg(trace_ctx, "\t\t\taes_cbc_iv_or_ctr = ");
+            for (unsigned int j = 0; j < 16 && j < 64; j++) {
```
lindongw: duplicate condition: j < 16 && j < 64
more in following code

lixin5x: 64 is the buffer size, i think it is meaningful to keep this code here to prevent overflow once someone want to change the dump size 16 by hand, maybe more clear if I use sizeof()

XinfengZhang (approved):

PR #882, opened 2026-05-18
va/va_h264_enc: Add constrained flag for AVC encoder

satchall-amd:
In android, few testcases in CtsMediaV2TestCases module related to AVC encoder are failed, due to mismatch in profile. By adding enc_constraint_set_flags for AVC encoder, these constraint flags are set based on profile.


satchall-amd:
This patch is used in the frontends/va of the mesa3d repository. A patch is submitted in the mesa3d which uses this enc_constraint_set_flags. 

Here the patch details: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/41652

satchall-amd:
Closing this pull request as this change is required only with old version of mesa3d, With the latest changes this is not required.

PR #892, opened 2026-07-31
va/trace: add microsecond resolution to trace file names

XinfengZhang:
The trace log file name is composed as
"<prefix>.<HHMMSS>.thd-<thread_id>", which only has second resolution and no per-display component. Each VADisplay has its own log_files_manager, so when an application creates multiple VADisplays on the same thread within the same second (e.g. one for decode and one for encode), they all generate an identical file name and open it with fopen(..., "w") (O_TRUNC). The later opens truncate the earlier ones, so some traces are silently lost (commonly the encode trace).

Whether the traces collide is purely a matter of timing: if the displays straddle a second boundary the names differ and all traces survive, otherwise one clobbers the others. This made trace capture non-deterministic.

Include microseconds in the timestamp so displays initialized within the same second get distinct file names, and update the buffer length guard accordingly.

### DIFF
--- a/va/va_trace.c
+++ b/va/va_trace.c
@@ -551,18 +551,25 @@ static void FILE_NAME_SUFFIX(
     if (suffix_str)
         size = strlen(suffix_str);
 
-    if (left < (size + 8 + 10))
+    if (left < (size + 15 + 10))
         return;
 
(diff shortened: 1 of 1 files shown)


xhaihao (approved):
LGTM

